PR #8, opened 2025-09-14
feat: Introduce Experimental `tvm_ffi.dataclasses.c_class`

junrushao:
Depends on #32.

This PR introduces an experimental `@c_class` decorator that enables Python dataclass-like syntax for exposing C++ objects through TVM FFI. The decorator automatically handles field reflection, inheritance, and constructor generation for FFI-backed classes.

## Example

On C++ side, register types using `tvm::ffi::reflection::ObjectDef<>`:

```C++
TVM_FFI_STATIC_INIT_BLOCK() {
    namespace refl = tvm::ffi::reflection;
    refl::ObjectDef<MyClass>()
        .def_static("__init__", [](int64_t v_i64, int32_t v_i32, double v_f64, float v_f32) -> Any {
            return ObjectRef(ffi::make_object<MyClass>(v_i64, v_i32, v_f64, v_f32));
        })
        .def_rw("v_i64", &MyClass::v_i64)
        .def_rw("v_i32", &MyClass::v_i32)
        .def_rw("v_f64", &MyClass::v_f64)
        .def_rw("v_f32", &MyClass::v_f32);
}
```

Mirror the same structure in Python using dataclass-style annotations:

```python
from tvm_ffi.dataclasses import c_class, field

@c_class("example.MyClass")
class MyClass:
    v_i64: int
    v_i32: int
    v_f64: float = field(default=0.0)
    v_f32: float = field(default_factory=lambda: 1.0)

obj = MyClass(v_i64=4, v_i32=8)
obj.v_f64 = 3.14  # transparently forwards to the underlying C++ object
```

## Future work

Supporting as many dataclass features as possible, including: `repr`, `init`, `order`, etc.

PR #31, opened 2025-09-19
feat: Expose `__ffi_init__` as a de-facto shortcut to calling `_C___init__`

junrushao:
Per discussion w/ @tqchen.

In dataclass-syntax classes as introduced in #8, we likely always use the reflected `_C___init__` method to initialize an instance. This PR gives a shortcut to this convention, where

```pytthon
self.__init_handle_by_constructor__(_MyClass._C___init__, ...)
```

is equivalent to

```
self.__ffi_init__(v_i64 + 1, v_i32 + 2)
```


tqchen:
minor nit, on c function naming, how about `__c_init__` for the pattern, it aligns with other dlpack style dunder naming(see https://github.com/dmlc/dlpack/issues/175, also shorter to type in def_static

junrushao:
Merged this PR into #8

PR #35, opened 2025-09-21
 feat: Enable mypy type checking

junrushao:
With this PR, we are able to enable mypy type checking for the following directories:
- `python/` and `tests/`
- `examples/inline_module`
- `examples/packaging`
- `examples/quick_start`


examples/quick_start/run_example.py
```diff
@@ -16,15 +16,9 @@
 # under the License.
 """Quick start script to run tvm-ffi examples from prebuilt libraries."""
 
-import tvm_ffi
-
-try:
-    import torch
-except ImportError:
-    torch = None
```
junrushao, line 24: We may simply assume `torch` exists under `examples`

python/tvm_ffi/_ffi_api.py
```diff
@@ -16,6 +16,37 @@
 # under the License.
 """FFI API."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
 from . import registry
 
+if TYPE_CHECKING:
```
junrushao: Alternatively, we may:
- Option A) Suppress type errors on all `_ffi_api.*` access
- Option B) Auto generate stubs in another `*.pyi` file

python/tvm_ffi/access_path.py
```diff
@@ -39,11 +39,30 @@ class AccessKind(IntEnum):
...
+    if TYPE_CHECKING:
+        kind: AccessKind
+        key: Any
+
 
 @register_object("ffi.reflection.AccessPath")
 class AccessPath(core.Object):
     """Access path container."""
 
+    if TYPE_CHECKING:
```
junrushao: ditto [here](https://github.com/apache/tvm-ffi/pull/35#discussion_r2366472871). We may want either a separate stub file via stub generation, or suppress type errors

python/tvm_ffi/cpp/load_inline.py
```diff
@@ -358,17 +358,17 @@ def _decorate_with_tvm_ffi(source: str, functions: Mapping[str, str]) -> str:
...
-def load_inline(
+def load_inline(  # noqa: PLR0912, PLR0915
     name: str,
     *,
-    cpp_sources: str | None = None,
-    cuda_sources: str | None = None,
-    functions: Sequence[str] | None = None,
+    cpp_sources: Sequence[str] | str | None = None,
+    cuda_sources: Sequence[str] | str | None = None,
+    functions: Mapping[str, str] | Sequence[str] | str | None = None,
```
junrushao, line 368: seems that mypy catches a type error here

python/tvm_ffi/testing.py
```diff
@@ -16,23 +16,37 @@
...
 @register_object("testing.TestObjectBase")
 class TestObjectBase(Object):
     """Test object base class."""
 
+    if TYPE_CHECKING:
+        v_i64: int
+        v_f64: float
+        v_str: str
+
+        def add_i64(self, x: int) -> int: ...  # noqa: D102
```
junrushao: ditto [this](https://github.com/apache/tvm-ffi/pull/35#discussion_r2366472871)

python/tvm_ffi/testing.py
```diff
@@ -90,5 +108,5 @@ class _TestCxxClassDerived(_TestCxxClassBase):
 
 @c_class("testing.TestCxxClassDerivedDerived")
 class _TestCxxClassDerivedDerived(_TestCxxClassDerived):
-    v_str: str = field(default_factory=lambda: "default")
-    v_bool: bool
+    v_str: str = field(default_factory=lambda: "default")  # type: ignore[assignment]
```
junrushao, line 106: TODO: need to figure out why `field(...)` triggers a mypy complain, despite `field_specifiers` are used properly

python/tvm_ffi/testing.py
tqchen: does type annocation have to be conditioned on TYPE_CHECKING? i feel we should be fine without such condition?

python/tvm_ffi/access_path.py
tqchen: Does this need to depend on TYPE_CHECKING?

python/tvm_ffi/testing.py
junrushao: It’s fine without such condition

junrushao:
Looks like the only controversy [[1](https://github.com/apache/tvm-ffi/pull/35#discussion_r2366473744)][[2](https://github.com/apache/tvm-ffi/pull/35#discussion_r2366475183)] is whether or not we want generated methods to be put under `TYPE_CHECKING` or not.

I don't really have strong opinion so @tqchen you may make the call. Arguments on both sides are valid:
- Reason why they should reside under `TYPE_CHECKING`: Those are not real methods with concrete implementation.
- Reason why they should not reside under `TYPE_CHECKING`: the extra indent looks not super necessary

tqchen:
if correctness is not an issue, i feel not put under `TYPE_CHECKING` is better, it is usually ok to put type annotation this way IMO

python/tvm_ffi/access_path.py
```diff
@@ -39,11 +39,28 @@ class AccessKind(IntEnum):
...
+    def _root() -> "AccessPath": ...  # type: ignore[empty-body]
+    def _path_equal(self, other: "AccessPath") -> bool: ...  # type: ignore[empty-body]
+    def _is_prefix_of(self, other: "AccessPath") -> bool: ...  # type: ignore[empty-body]
+    def _attr(self, attr_key: str) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _attr_missing(self, attr_key: str) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _array_item(self, index: int) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _array_item_missing(self, index: int) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _map_item(self, key: Any) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _map_item_missing(self, key: Any) -> "AccessPath": ...  # type: ignore[empty-body]
+    def _to_steps(self) -> list["AccessStep"]: ...  # type: ignore[empty-body]
```
junrushao: @tqchen I'm able to remove all `if TYPE_CHECKING` guards for stubs. The caveat is that for every stub method, we will need to suppress mypy's warning, e.g.

```
python/tvm_ffi/testing.py:38: error: Missing return statement  [empty-body]
```

junrushao: A reasonable balance to keep is:
- reflected fields are annotated without `TYPE_CHECKING`
- reflected methods are NOT explicitly written out, and warnings for them are suppressed individually, e.g.

```
return self._map_item(key)  # type: ignore[attr-defined]
```

python/tvm_ffi/testing.py
```diff
@@ -16,23 +16,35 @@
...
+
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from . import _ffi_api
 from .core import Object
 from .dataclasses import c_class, field
 from .registry import register_object
 
+if TYPE_CHECKING:
+    from tvm_ffi import Array, Map
```
tqchen: Given array map is used below (and notunder type chekcing), should it now not use type checking? or use String?

python/tvm_ffi/_ffi_api.py
tqchen: get rid of this TYPE_CHECKING or move to a pyi file

python/tvm_ffi/testing.py
junrushao: the original intention is to avoid potential cyclic dependencies, but indeed there's no cyclic dependency in this case. will remove

python/tvm_ffi/_ffi_api.py
junrushao: will move to a pyi file

PR #38, opened 2025-09-22
doc: Enable Sphinx Build Checks in CI

junrushao:
This PR adds an additional stage in parallel to stage `lint`, which builds Python docs via uv + sphinx on every commit and catches any warning from the building process via:

```
sphinx-build -W --keep-going -b html docs docs/_build/html
```

A few issues:
- doc building right now depends on wheel building, which may add some unnecessary CI time;
- tests for C++ doc building is not enabled yet. it seems plenty of warnings there.

image

### DIFF
--- a/docs/_stubs/cpp_index.rst
+++ b/docs/_stubs/cpp_index.rst
@@ -1,0 +1,24 @@
+..  Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+..    http://www.apache.org/licenses/LICENSE-2.0
+
+..  Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+==============
+Full API Index
+==============
+
+This page contains stub C++ API indexs to avoid Sphinx warning in CI testing.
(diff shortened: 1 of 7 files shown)


yzh119 (approved):

docs/reference/cpp/generated/index.rst
```diff
@@ -0,0 +1,22 @@
+..  Licensed to the Apache Software Foundation (ASF) under one
```
yzh119, line 1: Should we ignore `generated/` directory?

tqchen: agree this is something we should ignore

junrushao: It’s intentional - let me explain it a little bit: this page serves as a stub page to suppress warnings when `BUILD_CPP_DOCS` is not set. Otherwise CI will fail with an error saying this particular cpp related page is not found

tqchen: issue is that if the file get checked into ci and it get changed, we can easily commit such change(during git commit). Maybe another workaround is to first touch this file in CI check if not exist

junrushao: Sounds like a good idea 👍

junrushao: Updated! The most feasible workaround seems to keep this file in a separate path `docs/_stubs/cpp_index.rst` and `mv` to the right place in CI.

docs/_stubs/cpp_index.rst
```diff
@@ -0,0 +1,22 @@
...
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+==============
+Full API Index
+==============
+
+This page contains the full API index for the C++ API.
```
tqchen: maybe state this page contains stub API index for API building to avoid sphinx warning

junrushao: Updated. Please take another look!

tqchen (approved):

PR #49, opened 2025-09-23
feat: Auto-create Python classes when missing

junrushao:


python/tvm_ffi/cython/object.pxi
```diff
@@ -264,28 +266,42 @@ cdef inline object make_ret_opaque_object(TVMFFIAny result):
...
-    warnings.warn(f"Returning type `{type_key}` which is not registered via register_object, fallback to Object")
-    obj = _CLASS_OBJECT.__new__(_CLASS_OBJECT)
-    (<Object>obj).chandle = result.v_obj
-    return obj
+    else:
+        # object is not found in registered entry
+        # in this case we need to report an warning
+
+        from .utils import prototype_py
+
```
tqchen: i feel we do not need exec to create a new type, instead we can use `type` function?

junrushao: Updated with an approach that doesn't rely on `exec`. Let me know if it makes sense

python/tvm_ffi/cython/object.pxi
```diff
@@ -261,29 +263,53 @@ cdef inline object make_ret_opaque_object(TVMFFIAny result):
...
+            obj = Object.__new__(Object)
+            (<Object>obj).chandle = result.v_obj
+            return cls.__from_tvm_ffi_object__(cls, obj)
+        else:
+            pass
+    else:
+        # Slow path: object is not found in registered entry
+        # In this case we need to report a warning, create a dummy stub class
+        # and register the class for future usage.
+        cls = make_fallback_cls_for_type_index(type_index)
```
tqchen, line 324: did we register for future? we should update TYPE_INDEX_TO_CLS so that next time it is not so slow

Maybe the function already did that, in which case the comment should explicitly say so, maybe rename to `make_and_register_fallback_cls_for_type_index`

python/tvm_ffi/cython/object.pxi
```diff
@@ -261,29 +263,53 @@ cdef inline object make_ret_opaque_object(TVMFFIAny result):
...
+    cdef object parent_type_info = type_info.parent_type_info
+    cdef object cls
+    cdef dict ann
+
+    assert type_info.type_cls is None
+    assert parent_type_info is not None
+
+    warnings.warn(
+        f"Returning type `{type_key}` which is not registered via `@register_object`, "
+        f"falling back to auto-generated prototype:\n{type_info.prototype_py()}",
```
tqchen: likely we want to remove warning onwards since we might explicitly relyon this behavior

python/tvm_ffi/cython/object.pxi
```diff
@@ -261,29 +263,53 @@ cdef inline object make_ret_opaque_object(TVMFFIAny result):
...
+    )
+
+    from tvm_ffi.dataclasses import c_class
```
tqchen: this is a reverse dep, ideally we want to bring up related c_class logics to type_info.pxi so we don't have such reverse dep

python/tvm_ffi/cython/type_info.pxi
```diff
@@ -109,6 +110,50 @@ class TypeInfo:
...
+            return
+        parent_type_index = self.type_ancestors[-1]
+        parent_type_key = _type_index_to_key(parent_type_index)
+        self.parent_type_info = _lookup_type_info_from_type_key(parent_type_key)  # ensure parent is registered
+
+    def prototype_py(self) -> str:
+        """Generate Python source code for a prototype class of this type.
+
+        The generated class will have the correct inheritance structure and
+        field definitions, but will not have any method definitions.
```
tqchen: not sure if the string building is needed here now specifically for this given we have type based construction, likely not?

python/tvm_ffi/cython/object.pxi
junrushao: Oh yeah by calling `make_fallback_cls_for_type_index`, it is already registering the new `cls` into `TYPE_INDEX_TO_CLS ` as you suggested. See assert on line 291

python/tvm_ffi/cython/object.pxi
junrushao: I don't have conclusive thoughts if no warning is good behavior. On one hand, the warning does emit a prototype code for downstream users to handily copy paste; On the other hand, not sure if it means stricter requirement to have explicit class definitions

python/tvm_ffi/cython/object.pxi
junrushao: Also the unittest tests there's strictly a single warning

tqchen: right, in this case documentation and naming clarity would help a bit

python/tvm_ffi/cython/object.pxi
tqchen: indeed it can be debatable. Originally I thought we need a warning.

However, now that we have a richer exposures of the class, it effectively behaves like pybind, and there could be effective usecases for such cxx only class in JITing scenarios, as a result the note. Alternatively, if we can have a context that marks this is intentional, which could also be fine. But i think no warning should also be ok

python/tvm_ffi/cython/type_info.pxi
junrushao: It's not executed and is only used to provide hints to users who want to write `@c_class`. Should be pretty copy-paste-ready quality as it looks like:

```python
from typing import Any
from tvm_ffi import dataclasses as D
from tvm_ffi import core

PARENT_TYPE = core._lookup_type_info_from_type_key('ffi.Object').type_cls
assert PARENT_TYPE is not None

@D.c_class('testing.TestUnregisteredObject')
class TestUnregisteredObject(PARENT_TYPE):
  a: Any
  @staticmethod
  def __ffi_init__(*args): ...
```

python/tvm_ffi/cython/object.pxi
junrushao: I see - good point! How about having warnings only if a certain env variable is turned on?

python/tvm_ffi/cython/object.pxi
junrushao: Agreeing reverse dependency is in general anti-pattern, but in this particular case, I'd love to argue that reverse dependency isn't necessarily bad in this particular case for a few reasons:
- Consistency. A stub class registered here is guaranteed to be consistent with an explicitly written `@c_class`;
- Code size. Relevant impls under `dataclasses/` is relatively a larger chunk of code, which I would consider a bit oversized to be moved into Cython core

Let me know your thoughts!

tqchen: I am not sure if the intention would be always the returning a dataclass in this case. My original thinking is that we could make it an Object with related fields and function being exported, unless user explicitly wanted a dataclass, in which case perhaps explicit `@c_class` decorator is a better idea.

python/tvm_ffi/cython/type_info.pxi
tqchen: Ah i see, i see the tention, i think ideally for the case where returning object is unregistered, maybe we do not need to make it a dataclass..

python/tvm_ffi/cython/object.pxi
tqchen: warning under a flag seems to be ok

junrushao:
Updated the PR per offline discussion with @tqchen. To summarize the changes:
- `prototype_py()` is no longer used in class creation;
- No warnings are thrown when encountering unregistered classes to be consistent with pybind behavior;
- `@c_class` is not used in this case, instead, we refactored existing `register_object(...)` behavior into `TypeMethod.as_callable(...)` and `TypeField.as_property(...)`, which is shared by both `register_object(...)` and `make_fallback_cls_for_type_index(...)`;
- Clarified in the doc that the slow path `make_fallback_cls_for_type_index` is called only once for every missing type.

CC: @tqchen

python/tvm_ffi/cython/type_info.pxi
```diff
@@ -109,6 +125,57 @@ class TypeInfo:
...
+        cdef int parent_type_index
+        cdef str parent_type_key
+        if not self.type_ancestors:
+            return
+        parent_type_index = self.type_ancestors[-1]
+        parent_type_key = _type_index_to_key(parent_type_index)
+        self.parent_type_info = _lookup_type_info_from_type_key(parent_type_key)  # ensure parent is registered
+
+    def prototype_py(self) -> str:
+        """Generate Python source code for a prototype class of this type.
```
tqchen: seems now we no longer need prototype_py() as part of warning, we can leave it out for now in favor of future stub gen

junrushao: sounds good

python/tvm_ffi/cython/type_info.pxi
```diff
@@ -101,6 +102,21 @@ class TypeMethod:
...
+        cdef object func = self.func
+        if self.is_static:
+            func = staticmethod(func)
+        else:
+            func = _member_method_wrapper(func)
+        func.__module__ = cls.__module__
+        func.__name__ = name
+        func.__qualname__ = f"{cls.__qualname__}.{name}"
+        func.__doc__ = doc
+        return func
```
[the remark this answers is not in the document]
junrushao: I tested locally but can't reproduce the claimed issue. @gemini Could you double check?

junrushao:
@gemini review

python/tvm_ffi/cython/type_info.pxi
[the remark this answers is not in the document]
junrushao: /gemini I tested locally but can't reproduce the claimed issue. Could you double check?

junrushao:
**UPDATE**: Likely fixed by #55.

-----------------

The sporadic error message [[link](https://github.com/apache/tvm-ffi/actions/runs/17992193977/job/51184459149?pr=49#step:10:120)] looks alarming to me:

```
Exception caught during TVMFFIGetTypeInfo:
Traceback (most recent call last):
  File "include/tvm/ffi/type_traits.h", line 92, in tvm::ffi::TypeTraitsBase::GetMismatchTypeInfo[abi:cxx11](TVMFFIAny const*)
  File "include/tvm/ffi/object.h", line 121, in tvm::ffi::TypeIndexToTypeKey[abi:cxx11](int)
  File "src/ffi/object.cc", line 493, in TVMFFIGetTypeInfo
  File "src/ffi/object.cc", line 193, in tvm::ffi::TypeTable::Entry* tvm::ffi::TypeTable::GetTypeEntry(int32_t)
InternalError: Check failed: (entry != nullptr) is false: Cannot find type info for type_index=74
```

Particularly, <del>there's no type whose type index is 74 (details below):</del> Turns out it's `OpaquePyObject`.

<details>

```
Register type_index=64, type_key=ffi.Object, parent_index=-1 num_child_slots=0
Register type_index=0, type_key=None, parent_index=-1 num_child_slots=0
Register type_index=1, type_key=int, parent_index=-1 num_child_slots=0
Register type_index=3, type_key=float, parent_index=-1 num_child_slots=0
Register type_index=2, type_key=bool, parent_index=-1 num_child_slots=0
Register type_index=8, type_key=const char*, parent_index=-1 num_child_slots=0
Register type_index=4, type_key=void*, parent_index=-1 num_child_slots=0
Register type_index=5, type_key=DataType, parent_index=-1 num_child_slots=0
Register type_index=6, type_key=Device, parent_index=-1 num_child_slots=0
Register type_index=9, type_key=TVMFFIByteArray*, parent_index=-1 num_child_slots=0
Register type_index=10, type_key=ObjectRValueRef, parent_index=-1 num_child_slots=0
Register type_index=11, type_key=ffi.SmallStr, parent_index=-1 num_child_slots=0
Register type_index=12, type_key=ffi.SmallBytes, parent_index=-1 num_child_slots=0
Register type_index=67, type_key=ffi.Error, parent_index=64 num_child_slots=0
Register type_index=66, type_key=ffi.Bytes, parent_index=64 num_child_slots=0
Register type_index=65, type_key=ffi.String, parent_index=64 num_child_slots=0
Register type_index=72, type_key=ffi.Map, parent_index=64 num_child_slots=0
Register type_index=68, type_key=ffi.Function, parent_index=64 num_child_slots=0
Register type_index=71, type_key=ffi.Array, parent_index=64 num_child_slots=0
Register type_index=69, type_key=ffi.Shape, parent_index=64 num_child_slots=0
Register type_index=70, type_key=ffi.Tensor, parent_index=64 num_child_slots=0
Register type_index=128, type_key=ffi.reflection.AccessStep, parent_index=64 num_child_slots=0
Register type_index=129, type_key=ffi.reflection.AccessPath, parent_index=64 num_child_slots=0
Register type_index=73, type_key=ffi.Module, parent_index=64 num_child_slots=0
Register type_index=130, type_key=testing.TestIntPair, parent_index=64 num_child_slots=0
Register type_index=131, type_key=testing.TestObjectBase, parent_index=64 num_child_slots=1
Register type_index=132, type_key=testing.TestObjectDerived, parent_index=131 num_child_slots=0
Register type_index=133, type_key=testing.TestCxxClassBase, parent_index=64 num_child_slots=0
Register type_index=134, type_key=testing.TestCxxClassDerived, parent_index=133 num_child_slots=0
Register type_index=135, type_key=testing.TestCxxClassDerivedDerived, parent_index=134 num_child_slots=0
Register type_index=136, type_key=testing.TestCxxInitSubset, parent_index=64 num_child_slots=0
Register type_index=137, type_key=testing.TestUnregisteredBaseObject, parent_index=64 num_child_slots=0
Register type_index=138, type_key=testing.TestUnregisteredObject, parent_index=137 num_child_slots=0
```

</details>

I was unable to reproduce this error locally...digging

junrushao:
@tqchen may I get a re-review?

python/tvm_ffi/cython/string.pxi
```diff
@@ -77,3 +77,5 @@ class Bytes(bytes, PyNativeObject):
 
 
 _register_object_by_index(kTVMFFIBytes, Bytes)
+_register_object_by_index(kTVMFFIObject, Object)
+_register_object_by_index(kTVMFFIDevice, Device)
```
tqchen: device is not Object, need to remove

python/tvm_ffi/cython/type_info.pxi
```diff
@@ -109,6 +125,23 @@ class TypeInfo:
...
     parent_type_info: Optional[TypeInfo]
+
+    def __post_init__(self):
+        cdef int parent_type_index
+        cdef str parent_type_key
+        if not self.type_ancestors:
+            return
+        parent_type_index = self.type_ancestors[-1]
+        parent_type_key = _type_index_to_key(parent_type_index)
+        self.parent_type_info = _lookup_type_info_from_type_key(parent_type_key)  # ensure parent is registered
```
tqchen: given behavior, likely need to rename to _lookup_or_register_type_info_from_type_key

tqchen (changes requested):

junrushao:
@tqchen Review comments addressed 🫡 Let me know if there's anything further I could iterate on!

PR #57, opened 2025-09-25
[FFI][ABI] TVMFFIObject ABI to further align with torch intrusive_ptr

tqchen:
This PR updates the TVMFFIObject ABI to use strong_ref_count/weak_ref_count
into u32 and which can be packed into the first u64 field in the object header.

This change will open up further optimizations that combines weak/strong into
a single u64 atomic that skips the need of extra atomic reading of weak value
during deletion. The technique was recently in torch

We also reordered the counter to be first since most rc ops are frequent and
it would be more cache friendly. We do give up the same type_index first location
with TVMFFIAny, which is OK as they optimize for different goals.


tqchen:
referencing pytorch prs https://github.com/pytorch/pytorch/pull/163394. Thanks to @swolchok for valuable suggestion

tqchen:
https://github.com/apache/tvm-ffi/pull/58 contains a followup to actually use combined_ref_count

PR #63, opened 2025-09-27
[CYTHON] Fix potential flaky issue in dispatcher

tqchen:
The C++ dispatcher dispatches the argument passing by TYPE(obj) pointer which
is non-owning. This means that there is the following edge case:
- type A is registered through dispatcher
- type A gets garbage collected (because it is a local type)
- type B is created and uses the same memory address as type A

Then when we pass in type B, it will mistakenly use the dispatch function for type A

To prevent this, we keep alive the types that are registered through dispatcher
by appending them to _DISPATCH_TYPE_KEEP_ALIVE

NOTE that the total number of types that are registered through dispatcher is expected
to be limited in practice so we can afford to keep them alive

### DIFF
--- a/python/tvm_ffi/cython/function.pxi
+++ b/python/tvm_ffi/cython/function.pxi
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 import ctypes
+import threading
 import os
 from numbers import Real, Integral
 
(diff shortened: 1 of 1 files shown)


python/tvm_ffi/cython/function.pxi
```diff
@@ -501,6 +501,8 @@ cdef int TVMFFIPyArgSetterDTypeFromNumpy_(
     out.v_dtype = NUMPY_DTYPE_TO_DTYPE[py_obj]
     return 0
 
+cdef _DISPATCH_TYPE_KEEP_ALIVE = set()
```
junrushao, line 506: I was worried if there're concurrent calls from multiple threads - in this case will this corrupt this global variable?

junrushao: i see - the factory is only called once. Probably fine

junrushao (approved):

python/tvm_ffi/cython/function.pxi
tqchen: great call out

tqchen:
pushed an update with lock for nogil version safety

PR #78, opened 2025-10-01
feat: Address warnings from clang-tidy 

junrushao:
This PR introduces `clang-tidy` and applies a large number of fixes it suggested.

The changes include:
- style modernizations, e.g., using `_v` and `_t` suffixes, `using` instead of `typedef`, range-based for loops
- perf improvements, e.g., using `std::move`
- and correctness fixes, e.g., adding `noexcept`, explicit casts, fixing dangling pointers

Note that `clang-tidy` is not added to CI in this PR, and to manually trigger trigger clang-tidy in this codebase, we will have to run:

```
uv run --with "clang-tidy==21.1.1"          \
  python tests/lint/clang_tidy_precommit.py \
    --build-dir=build-pre-commit            \
    --jobs=16                               \
    ./src/ ./include ./tests
```

### DIFF
--- a/cmake/Utils/AddLibbacktrace.cmake
+++ b/cmake/Utils/AddLibbacktrace.cmake
@@ -62,8 +62,9 @@ function (_libbacktrace_compile)
   add_library(libbacktrace STATIC IMPORTED)
   add_dependencies(libbacktrace project_libbacktrace)
   set_target_properties(
-    libbacktrace PROPERTIES IMPORTED_LOCATION ${libbacktrace_prefix}/lib/libbacktrace.a
-                            INTERFACE_INCLUDE_DIRECTORIES ${libbacktrace_prefix}/include
+    libbacktrace
+    PROPERTIES IMPORTED_LOCATION ${libbacktrace_prefix}/lib/libbacktrace.a
+               INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_CURRENT_LIST_DIR}/../../3rdparty/libbacktrace/
   )
 endfunction ()
 
--- a/include/tvm/ffi/any.h
+++ b/include/tvm/ffi/any.h
@@ -83,12 +83,12 @@ class AnyView {
   /*! \brief Copy assignment operator */
   AnyView& operator=(const AnyView&) = default;
   /*! \brief Move constructor */
-  AnyView(AnyView&& other) : data_(other.data_) {
+  AnyView(AnyView&& other) noexcept : data_(other.data_) {
     other.data_.type_index = TypeIndex::kTVMFFINone;
     other.data_.zero_padding = 0;
     other.data_.v_int64 = 0;
   }
-  TVM_FFI_INLINE AnyView& operator=(AnyView&& other) {
+  TVM_FFI_INLINE AnyView& operator=(AnyView&& other) noexcept {
     // copy-and-swap idiom
     AnyView(std::move(other)).swap(*this);  // NOLINT(*)
     return *this;
@@ -286,7 +286,7 @@ class Any {
    * \brief Move constructor from another Any
    * \param other The other Any
    */
-  Any(Any&& other) : data_(other.data_) {
+  Any(Any&& other) noexcept : data_(other.data_) {
     other.data_.type_index = TypeIndex::kTVMFFINone;
     other.data_.zero_padding = 0;
     other.data_.v_int64 = 0;
@@ -304,7 +304,7 @@ class Any {
    * \brief Move assign from another Any
    * \param other The other Any
    */
-  TVM_FFI_INLINE Any& operator=(Any&& other) {
+  TVM_FFI_INLINE Any& operator=(Any&& other) noexcept {
     // copy-and-swap idiom
     Any(std::move(other)).swap(*this);  // NOLINT(*)
     return *this;
@@ -326,7 +326,9 @@ class Any {
     return *this;
   }
   /*! \brief Any can be converted to AnyView in zero cost. */
-  operator AnyView() const { return AnyView::CopyFromTVMFFIAny(data_); }
+  operator AnyView() const {  // NOLINT(google-explicit-constructor)
+    return AnyView::CopyFromTVMFFIAny(data_);
+  }
   /*!
    * \brief Constructor from a general type
    * \tparam T The value type of the other
@@ -540,12 +542,12 @@ struct AnyUnsafe : public ObjectUnsafe {
     return result;
   }
 
-  TVM_FFI_INLINE static Any MoveTVMFFIAnyToAny(TVMFFIAny&& data) {
+  TVM_FFI_INLINE static Any MoveTVMFFIAnyToAny(TVMFFIAny* data) {
     Any any;
-    any.data_ = data;
-    data.type_index = TypeIndex::kTVMFFINone;
-    data.zero_padding = 0;
-    data.v_int64 = 0;
+    any.data_ = *data;
+    data->type_index = TypeIndex::kTVMFFINone;
+    data->zero_padding = 0;
+    data->v_int64 = 0;
     return any;
   }
 
--- a/include/tvm/ffi/base_details.h
+++ b/include/tvm/ffi/base_details.h
@@ -121,13 +121,13 @@
  * \brief Define the default copy/move constructor and assign operator
  * \param TypeName The class typename.
  */
-#define TVM_FFI_DEFINE_DEFAULT_COPY_MOVE_AND_ASSIGN(TypeName) \
-  TypeName(const TypeName& other) = default;                  \
-  TypeName(TypeName&& other) = default;                       \
-  TypeName& operator=(const TypeName& other) = default;       \
-  TypeName& operator=(TypeName&& other) = default;
+#define TVM_FFI_DEFINE_DEFAULT_COPY_MOVE_AND_ASSIGN(TypeName)                                    \
+  TypeName(const TypeName& other) = default;            /* NOLINT(bugprone-macro-parentheses) */ \
+  TypeName(TypeName&& other) noexcept = default;        /* NOLINT(bugprone-macro-parentheses) */ \
+  TypeName& operator=(const TypeName& other) = default; /* NOLINT(bugprone-macro-parentheses) */ \
+  TypeName& operator=(TypeName&& other) noexcept = default; /* NOLINT(bugprone-macro-parentheses)*/
 
-/**
+/*!
  * \brief marks the begining of a C call that logs exception
  */
 #define TVM_FFI_LOG_EXCEPTION_CALL_BEGIN() \
@@ -151,9 +151,9 @@
  * This macro is used to clear the padding parts for hash and equality check
  * in 32bit platform.
  */
-#define TVM_FFI_CLEAR_PTR_PADDING_IN_FFI_ANY(result)                    \
-  if constexpr (sizeof((result)->v_obj) != sizeof((result)->v_int64)) { \
-    (result)->v_int64 = 0;                                              \
+#define TVM_FFI_CLEAR_PTR_PADDING_IN_FFI_ANY(result) \
+  if constexpr (sizeof(void*) != sizeof(int64_t)) {  \
+    (result)->v_int64 = 0;                           \
   }
 
 namespace tvm {
@@ -182,7 +182,7 @@ void for_each(const F& f, Args&&... args) {  // NOLINT(*)
  * \param value The right operand.
  * \return the combined result.
  */
-template <typename T, std::enable_if_t<std::is_convertible<T, uint64_t>::value, bool> = true>
+template <typename T, std::enable_if_t<std::is_convertible_v<T, uint64_t>, bool> = true>
 TVM_FFI_INLINE uint64_t StableHashCombine(uint64_t key, const T& value) {
   // XXX: do not use std::hash in this function. This hash must be stable
   // across different platforms and std::hash is implementation dependent.
@@ -211,12 +211,14 @@ TVM_FFI_INLINE uint64_t StableHashBytes(const void* data_ptr, size_t size) {
     // if alignment requirement is met, directly use load
     if (reinterpret_cast<uintptr_t>(it) % 8 == 0) {
       for (; it + 8 <= end; it += 8) {
+        // NOLINTNEXTLINE(clang-analyzer-security.ArrayBound)
         u.b = *reinterpret_cast<const uint64_t*>(it);
         result = (result * kMultiplier + u.b) % kMod;
       }
     } else {
       // unaligned version
       for (; it + 8 <= end; it += 8) {
+        // NOLINTNEXTLINE(clang-analyzer-security.ArrayBound)
         u.a[0] = it[0];
         u.a[1] = it[1];
         u.a[2] = it[2];
@@ -262,8 +264,6 @@ TVM_FFI_INLINE uint64_t StableHashBytes(const void* data_ptr, size_t size) {
     }
     if (it + 1 <= end) {
       a[0] = it[0];
-      it += 1;
-      a += 1;
     }
     if constexpr (!TVM_FFI_IO_NO_ENDIAN_SWAP) {
       std::swap(u.a[0], u.a[7]);
--- a/include/tvm/ffi/c_api.h
+++ b/include/tvm/ffi/c_api.h
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+// NOLINTBEGIN(modernize-use-using,bugprone-reserved-identifier,modernize-deprecated-headers)
 /*
  * \file tvm/ffi/c_api.h
  * \brief This file defines the C convention of the FFI convention
@@ -1183,3 +1183,4 @@ inline DLDevice TVMFFIDLDeviceFromIntPair(int32_t device_type, int32_t device_id
 }
 #endif  // __cplusplus
 #endif  // TVM_FFI_C_API_H_
+// NOLINTEND(modernize-use-using,bugprone-reserved-identifier,modernize-deprecated-headers)
--- a/include/tvm/ffi/container/map.h
+++ b/include/tvm/ffi/container/map.h
@@ -68,7 +68,7 @@ class MapObj : public Object {
   /*! \brief Iterator class */
   class iterator;
 
-  static_assert(std::is_standard_layout<KVType>::value, "KVType is not standard layout");
+  static_assert(std::is_standard_layout_v<KVType>, "KVType is not standard layout");
   static_assert(sizeof(KVType) == 32, "sizeof(KVType) incorrect");
 
   /// \cond Doxygen_Suppress
@@ -421,8 +421,8 @@ class SmallMapObj : public MapObj,
       ++map_node->size_;
       return;
     }
-    uint64_t next_size = std::max(map_node->NumSlots() * 2, uint64_t(kInitSize));
-    next_size = std::min(next_size, uint64_t(kMaxSize));
+    uint64_t next_size = std::max(map_node->NumSlots() * 2, kInitSize);
+    next_size = std::min(next_size, kMaxSize);
     TVM_FFI_ICHECK_GT(next_size, map_node->NumSlots());
     ObjectPtr<Object> new_map = CreateFromRange(next_size, map_node->begin(), map_node->end());
     InsertMaybeReHash(std::move(kv), &new_map);
@@ -520,9 +520,9 @@ class DenseMapObj : public MapObj {
   /*! \brief Maximum load factor of the hash map */
   static constexpr double kMaxLoadFactor = 0.99;
   /*! \brief Binary representation of the metadata of an empty slot */
-  static constexpr uint8_t kEmptySlot = uint8_t(0b11111111);
+  static constexpr uint8_t kEmptySlot = static_cast<uint8_t>(0b11111111);
   /*! \brief Binary representation of the metadata of a protected slot */
-  static constexpr uint8_t kProtectedSlot = uint8_t(0b11111110);
+  static constexpr uint8_t kProtectedSlot = static_cast<uint8_t>(0b11111110);
   /*! \brief Number of probing choices available */
   static constexpr int kNumJumpDists = 126;
   /*! \brief Index indicator to indicate an invalid index */
@@ -536,14 +536,14 @@ class DenseMapObj : public MapObj {
     uint64_t next = kInvalidIndex;
 
     explicit ItemType(KVType&& data) : data(std::move(data)) {}
-    explicit ItemType(key_type key, mapped_type value) : data(key, value) {}
+    explicit ItemType(key_type key, mapped_type value) : data(std::move(key), std::move(value)) {}
   };
   /*! \brief POD type of a block of memory */
   struct Block {
     uint8_t bytes[kBlockCap + kBlockCap * sizeof(ItemType)];
   };
   static_assert(sizeof(Block) == kBlockCap * (sizeof(ItemType) + 1), "sizeof(Block) incorrect");
-  static_assert(std::is_standard_layout<Block>::value, "Block is not standard layout");
+  static_assert(std::is_standard_layout_v<Block>, "Block is not standard layout");
 
   /*!
    * \brief Deleter for the Block
@@ -858,8 +858,8 @@ class DenseMapObj : public MapObj {
       ItemType* data_ptr = reinterpret_cast<ItemType*>(GetBlock(bi)->bytes + kBlockCap);
       for (int j = 0; j < kBlockCap; ++j, ++meta_ptr, ++data_ptr) {
         uint8_t& meta = *meta_ptr;
-        if (meta != uint8_t(kProtectedSlot) && meta != uint8_t(kEmptySlot)) {
-          meta = uint8_t(kEmptySlot);
+        if (meta != kProtectedSlot && meta != kEmptySlot) {
+          meta = kEmptySlot;
           data_ptr->ItemType::~ItemType();
         }
       }
@@ -903,7 +903,7 @@ class DenseMapObj : public MapObj {
     p->iter_list_head_ = kInvalidIndex;
     p->iter_list_tail_ = kInvalidIndex;
     for (uint64_t i = 0; i < n_blocks; ++i, ++block) {
-      std::fill(block->bytes, block->bytes + kBlockCap, uint8_t(kEmptySlot));
+      std::fill(block->bytes, block->bytes + kBlockCap, kEmptySlot);
     }
     return p;
   }
@@ -934,7 +934,7 @@ class DenseMapObj : public MapObj {
            ++j, ++meta_ptr_from, ++data_ptr_from, ++meta_ptr_to, ++data_ptr_to) {
         uint8_t& meta = *meta_ptr_to = *meta_ptr_from;
         TVM_FFI_ICHECK(meta != kProtectedSlot);
-        if (meta != uint8_t(kEmptySlot)) {
+        if (meta != kEmptySlot) {
           new (data_ptr_to) ItemType(*data_ptr_from);
         }
       }
@@ -984,7 +984,9 @@ class DenseMapObj : public MapObj {
    * \brief Check whether the hash table is full
    * \return A boolean indicating whether hash table is full
    */
-  bool IsFull() const { return size_ + 1 > NumSlots() * kMaxLoadFactor; }
+  bool IsFull() const {  // NOLINTNEXTLINE(bugprone-narrowing-conversions)
+    return (size_ + 1) > static_cast<uint64_t>(NumSlots()) * kMaxLoadFactor;
+  }
   /*!
    * \brief Increment the pointer
    * \param index The pointer to be incremented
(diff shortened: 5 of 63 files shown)


include/tvm/ffi/extra/module.h
```diff
@@ -113,7 +115,7 @@ class TVM_FFI_EXTRA_CXX_API ModuleObj : public Object {
    * \param format Format of the source code, can be empty by default.
    * \return Possible source code when available, or empty string if not available.
    */
-  virtual String InspectSource(const String& format = "") const { return String(); }
+  virtual String InspectSource(const String& format) const { return String(); }
```
junrushao, line 118: Virtual method w/ default arguments is highly recommended against

include/tvm/ffi/extra/module.h
```diff
@@ -201,7 +203,7 @@ class Module : public ObjectRef {
   /*!
    * \brief Property of ffi::Module
    */
-  enum ModulePropertyMask : int {
+  enum ModulePropertyMask : std::uint8_t {
```
junrushao: Recommended by [`performance-enum-size`](https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html) but i guess it likely doesn't matter

include/tvm/ffi/extra/module.h
```diff
@@ -233,7 +235,9 @@ class Module : public ObjectRef {
    * \brief Constructor from ObjectPtr<ModuleObj>.
    * \param ptr The object pointer.
    */
-  explicit Module(ObjectPtr<ModuleObj> ptr) : ObjectRef(ptr) { TVM_FFI_ICHECK(ptr != nullptr); }
+  explicit Module(const ObjectPtr<ModuleObj>& ptr) : ObjectRef(ptr) {
```
junrushao, line 238: Copy elision won't take place in the original impl

include/tvm/ffi/reflection/access_path.h
```diff
@@ -40,7 +41,7 @@ namespace reflection {
 /*!
  * \brief The kind of the access pattern.
  */
-enum class AccessKind : int32_t {
+enum class AccessKind : std::uint8_t {
```
junrushao: Ditto 

https://github.com/apache/tvm-ffi/pull/78/files#r2404208762

include/tvm/ffi/any.h
```diff
@@ -83,12 +83,12 @@ class AnyView {
   /*! \brief Copy assignment operator */
   AnyView& operator=(const AnyView&) = default;
   /*! \brief Move constructor */
-  AnyView(AnyView&& other) : data_(other.data_) {
+  AnyView(AnyView&& other) noexcept : data_(other.data_) {
```
junrushao, line 86: Recommended by [`performance-noexcept-move-constructor`](https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html)

include/tvm/ffi/container/variant.h
```diff
@@ -256,7 +257,7 @@ struct TypeTraits<Variant<V...>> : public TypeTraitsBase {
   }
 
   TVM_FFI_INLINE static Variant<V...> MoveFromAnyAfterCheck(TVMFFIAny* src) {
-    return Variant<V...>(details::AnyUnsafe::MoveTVMFFIAnyToAny(std::move(*src)));
+    return Variant<V...>(details::AnyUnsafe::MoveTVMFFIAnyToAny(src));
```
junrushao, line 260: `std::move` of the variable 'src' of the trivially-copyable type 'TVMFFIAny' has no effect

include/tvm/ffi/base_details.h
```diff
@@ -151,9 +151,9 @@
  * This macro is used to clear the padding parts for hash and equality check
  * in 32bit platform.
  */
-#define TVM_FFI_CLEAR_PTR_PADDING_IN_FFI_ANY(result)                    \
-  if constexpr (sizeof((result)->v_obj) != sizeof((result)->v_int64)) { \
-    (result)->v_int64 = 0;                                              \
+#define TVM_FFI_CLEAR_PTR_PADDING_IN_FFI_ANY(result) \
+  if constexpr (sizeof(void*) != sizeof(int64_t)) {  \
```
junrushao, line 155: Recommended by [`bugprone-sizeof-expression`](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/sizeof-expression.html)

include/tvm/ffi/c_api.h
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+// NOLINTBEGIN(performance-enum-size,modernize-use-using,bugprone-reserved-identifier,modernize-deprecated-headers)
```
junrushao: Skip diagnostics in C headers

include/tvm/ffi/dtype.h
```diff
@@ -39,7 +40,7 @@ namespace ffi {
  *
  * This class is always consistent with the DLPack.
  */
-enum DLExtDataTypeCode { kDLExtCustomBegin = 129 };
+enum DLExtDataTypeCode : std::uint8_t { kDLExtCustomBegin = 129 };
```
junrushao: Ditto https://github.com/apache/tvm-ffi/pull/78/files#r2404208762

src/ffi/extra/json_parser.cc
```diff
@@ -541,7 +542,8 @@ class JSONParser {
   }
 
  private:
-  explicit JSONParser(String json_str) : ctx_(json_str.data(), json_str.data() + json_str.size()) {}
+  explicit JSONParser(String json_str)
+      : storage_(std::move(json_str)), ctx_(storage_.data(), storage_.data() + storage_.size()) {}
```
junrushao, line 546: Added a field `storage_: String` to handle potential lifetime issue, but I suppose in this specific case it could be false positive

src/ffi/extra/buffer_stream.h
```diff
@@ -67,7 +69,7 @@ class BufferInStream {
   bool Read(T* data) {
     bool ret = Read(static_cast<void*>(data), sizeof(T)) == sizeof(T);  // NOLINT(*)
     if (!TVM_FFI_IO_NO_ENDIAN_SWAP) {
-      ByteSwap(&data, sizeof(T), 1);
+      ByteSwap(static_cast<void*>(&data), sizeof(T), 1);
```
junrushao: Not sure if this `&` is desirable

src/ffi/backtrace.cc
```diff
@@ -136,7 +136,7 @@ const TVMFFIByteArray* TVMFFIBacktrace(const char* filename, int lineno, const c
   // libbacktrace eats memory if run on multiple threads at the same time, so we guard against it
   if (tvm::ffi::_bt_state != nullptr) {
     static std::mutex m;
-    std::lock_guard<std::mutex> lock(m);
+    std::scoped_lock<std::mutex> lock(m);
```
junrushao, line 139: Recommended by [`modernize-use-scoped-lock`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-scoped-lock.html). Always replace `std::lock_guard` with `std::scoped_lock`

include/tvm/ffi/extra/module.h
tqchen: for ABI stability, we should keep it as int

include/tvm/ffi/reflection/access_path.h
tqchen: i think this is not something we should do, mainly for ABI stability and chance of future growth

include/tvm/ffi/extra/json.h
```diff
@@ -76,7 +76,7 @@ TVM_FFI_EXTRA_CXX_API json::Value Parse(const String& json_str, String* error_ms
  * \return The output JSON string.
  */
 TVM_FFI_EXTRA_CXX_API String Stringify(const json::Value& value,
-                                       Optional<int> indent = std::nullopt);
+                                       const Optional<int>& indent = std::nullopt);
```
tqchen: change this would change ABI, and pas sby value is OK in this setting

src/ffi/extra/buffer_stream.h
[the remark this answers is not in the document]
junrushao: @tqchen Could you please comment on this? The change per se follows the original logics but it's uncertain if the original logic is correct.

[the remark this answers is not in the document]
tqchen: you are right, it is is a bug, we didn't find it out because big endian is uncommon, makes sense to send a separate patch

src/ffi/extra/json_parser.cc
```diff
@@ -630,7 +632,7 @@ class JSONParser {
       return true;
     }
     // non-empty object
-    while ((next_char = ctx_.Peek()) != -1) {
+    while (ctx_.Peek() != -1) {
```
tqchen: there is next char query that immediately follows this, remove next_char assignment may leads to an issue here

include/tvm/ffi/container/map.h
```diff
@@ -984,7 +984,9 @@ class DenseMapObj : public MapObj {
    * \brief Check whether the hash table is full
    * \return A boolean indicating whether hash table is full
    */
-  bool IsFull() const { return size_ + 1 > NumSlots() * kMaxLoadFactor; }
+  bool IsFull() const {
+    return static_cast<double>(size_ + 1) > static_cast<double>(NumSlots()) * kMaxLoadFactor;
```
tqchen: the comparison shuold not happn in double...

include/tvm/ffi/function.h
```diff
@@ -711,11 +713,11 @@ class TypedFunction<R(Args...)> {
    * \param args The arguments
    * \returns The return value.
    */
-  TVM_FFI_INLINE R operator()(Args... args) const {
+  TVM_FFI_INLINE R operator()(const Args&... args) const {
```
tqchen: I am not sure const reference here is better than std::forward, likely keeping original approach is better

include/tvm/ffi/object.h
```diff
@@ -202,10 +202,7 @@ class Object {
   TVMFFIObject header_;
 
  public:
-  Object() {
-    header_.combined_ref_count = 0;
-    header_.deleter = nullptr;
-  }
+  Object() : header_{} {}
```
tqchen: readability, direct blanket intiialization is not as readable, even perhaps less efficient than assign

src/ffi/extra/json_writer.cc
```diff
@@ -34,14 +34,15 @@
...
+#include <utility>
 
 namespace tvm {
 namespace ffi {
 namespace json {
 
 class JSONWriter {
  public:
-  static String Stringify(const json::Value& value, Optional<int> indent) {
+  static String Stringify(const json::Value& value, const Optional<int>& indent) {
```
tqchen: optional of pod type can be passed by value

src/ffi/extra/json_writer.cc
```diff
@@ -187,7 +188,7 @@ class JSONWriter {
     if (indent_ != 0) {
       total_indent_ += indent_;
     }
-    for (size_t i = 0; i < value.size(); ++i) {
+    for (int64_t i = 0, n = static_cast<int64_t>(value.size()); i < n; ++i) {
```
tqchen: likely original code is better

src/ffi/extra/library_module_dynamic_lib.cc
```diff
@@ -45,7 +45,7 @@ namespace ffi {
 class DSOLibrary final : public Library {
  public:
   explicit DSOLibrary(const String& name) { Load(name); }
-  ~DSOLibrary() {
+  ~DSOLibrary() override {
```
tqchen: virtual destructor do not need override keyword?

src/ffi/extra/library_module_system_lib.cc
```diff
@@ -40,7 +41,7 @@ class SystemLibSymbolRegistry {
     auto it = symbol_table_.find(name);
     if (it != symbol_table_.end() && ptr != (*it).second) {
       std::cerr << "Warning:SystemLib symbol " << name << " get overriden to a different address "
-                << ptr << "->" << (*it).second << std::endl;
+                << ptr << "->" << (*it).second << '\n';
```
tqchen: std::endl is better here

src/ffi/extra/testing.cc
```diff
@@ -159,7 +160,7 @@ class TestUnregisteredObject : public TestUnregisteredBaseObject {
                               TestUnregisteredBaseObject);
 };
 
-TVM_FFI_NO_INLINE void TestRaiseError(String kind, String msg) {
+TVM_FFI_NO_INLINE void TestRaiseError(const String& kind, const String& msg) {
```
tqchen: for the ffi registered functions, given we already ask for a fresh value that can be moved, pass by value is better, and allows small string be passed by stack

src/ffi/backtrace.cc
```diff
@@ -153,7 +153,7 @@ void TVMFFISegFaultHandler(int sig) {
   // crashing.
   const TVMFFIByteArray* backtrace = TVMFFIBacktrace(nullptr, 0, nullptr, 1);
   std::cerr << "!!!!!!! Segfault encountered !!!!!!!\n"
-            << std::string(backtrace->data, backtrace->size) << std::endl;
+            << std::string(backtrace->data, backtrace->size) << '\n';
```
tqchen: need `std::endl` to trigger a flush and show in stream

tqchen: need std::endl for flush

src/ffi/extra/structural_equal.cc
```diff
@@ -255,7 +255,7 @@ class StructEqualHandler {
     }
   }
 
-  bool CompareMap(Map<Any, Any> lhs, Map<Any, Any> rhs) {
+  bool CompareMap(const Map<Any, Any>& lhs, const Map<Any, Any>& rhs) {
```
tqchen: under this paticular context, non-const works ok and not worse than const refer, mainly because caller always move into the content here

include/tvm/ffi/function.h
```diff
@@ -711,11 +713,11 @@ class TypedFunction<R(Args...)> {
...
     if constexpr (std::is_same_v<R, void>) {
-      packed_(std::forward<Args>(args)...);
```
tqchen, line 716: likely std::forward is better here via perfect forwarding

src/ffi/extra/testing.cc
```diff
@@ -235,7 +236,7 @@ TVM_FFI_STATIC_INIT_BLOCK() {
                }
                std::this_thread::sleep_for(std::chrono::seconds(1));
              }
-             std::cout << "Function finished without catching signal" << std::endl;
+             std::cout << "Function finished without catching signal" << '\n';
```
tqchen: std::endl is needed for flush

junrushao:
There's a potential leak clang-tidy detected when allocating an object in `include/tvm/ffi/memory.h`:

<details>

```
/Users/jshao/Projects/tvm-ffi/tests/cpp/extra/test_structural_equal_hash.cc:154:1: error: Potential leak of memory pointed to by field 'data_' [clang-analyzer-unix.Malloc,-warnings-as-errors]
  154 | }
      | ^
/Users/jshao/Projects/tvm-ffi/tests/cpp/extra/test_structural_equal_hash.cc:145:55: note: Calling 'AccessStep::Attr'
  145 |                                                       refl::AccessStep::Attr("body"),
      |                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/reflection/access_path.h:122:12: note: Calling constructor for 'AccessStep'
  122 |     return AccessStep(AccessKind::kAttr, std::move(field_name));
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/reflection/access_path.h:114:19: note: Calling 'make_object<tvm::ffi::reflection::AccessStepObj, tvm::ffi::reflection::AccessKind &, tvm::ffi::Any>'
  114 |       : ObjectRef(make_object<AccessStepObj>(kind, std::move(key))) {}
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:265:10: note: Calling 'ObjAllocatorBase::make_object'
  265 |   return details::SimpleObjAllocator().make_object<T>(std::forward<Args>(args)...);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:116:13: note: Calling 'Handler::New'
  116 |       ptr = Handler::New(static_cast<Derived*>(this), std::forward<Args>(args)...);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:179:20: note: Calling 'AlignedAlloc<8UL>'
  179 |       void* data = AlignedAlloc<alignof(T)>(sizeof(T));
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:65:3: note: Taking true branch
   65 |   if constexpr (align <= alignof(std::max_align_t)) {
      |   ^
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:67:21: note: Memory is allocated
   67 |     if (void* ptr = std::malloc(size)) {
      |                     ^~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:67:15: note: Assuming 'ptr' is non-null
   67 |     if (void* ptr = std::malloc(size)) {
      |               ^~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:67:5: note: Taking true branch
   67 |     if (void* ptr = std::malloc(size)) {
      |     ^
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:179:20: note: Returned allocated memory
  179 |       void* data = AlignedAlloc<alignof(T)>(sizeof(T));
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:116:13: note: Returned allocated memory
  116 |       ptr = Handler::New(static_cast<Derived*>(this), std::forward<Args>(args)...);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/memory.h:265:10: note: Returned allocated memory
  265 |   return details::SimpleObjAllocator().make_object<T>(std::forward<Args>(args)...);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/reflection/access_path.h:114:19: note: Returned allocated memory
  114 |       : ObjectRef(make_object<AccessStepObj>(kind, std::move(key))) {}
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/reflection/access_path.h:122:12: note: Returning from constructor for 'AccessStep'
  122 |     return AccessStep(AccessKind::kAttr, std::move(field_name));
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/tests/cpp/extra/test_structural_equal_hash.cc:145:55: note: Returned allocated memory
  145 |                                                       refl::AccessStep::Attr("body"),
      |                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jshao/Projects/tvm-ffi/tests/cpp/extra/test_structural_equal_hash.cc:154:1: note: Potential leak of memory pointed to by field 'data_'
  154 | }
      | ^
```

</details>

Example code that triggers this warning:

```C++
std::vector<refl::AccessStep> steps = {refl::AccessStep::Attr("body")};
```

tqchen:
on the leak warning: seems it complains about new that was not freed, in this case, the memory is managed by the ref counter and will be freed by deleter, and likely not detected by the clang-tidy analysis given that free is dynamic in future

junrushao:
Another simplified version of this leak report:

The snippet below reports leak:

```C++
Optional<TInt> x;
TInt v = x.value_or(TInt(11));
```

However, lifting `TInt(11)` out gets rid of this leak:

```C++
Optional<TInt> x;
TInt y = TInt(11);
TInt v = x.value_or(y);
```

junrushao:
Further simplification:

```C++
TInt MyMove1(bool self, TInt&& x) 
  if (self) {
    return TInt(12);
  } else {
    return TInt(std::forward<TInt>(x));
  }
}

TInt MyMove2(bool self, TInt&& x) {
  return self ? TInt(12) : TInt(std::forward<TInt>(x));
}

void Test() {
  TInt v1 = MyMove1(true, TInt(11)); // ok
  TInt v2 = MyMove2(true, TInt(11)); // failed
}
```

include/tvm/ffi/function.h
junrushao: it does suggest us to use const reference, but i suppose with `std::move` there won't be significant overhead.

```
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/function.h:716:39: error: the parameter 'args' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
  716 |   TVM_FFI_INLINE R operator()(Args... args) const {
      |                                       ^
      |                               const  &
```

src/ffi/extra/library_module_dynamic_lib.cc
junrushao: That's not necessary but it's recommended by [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html). For some reason, the standard practice is to always mark override/final to prevent typos such as qualifier change

include/tvm/ffi/function.h
tqchen: std::forward effectively allows forward both const ref or move, depending on input

include/tvm/ffi/object.h
junrushao: Reverted back to using assign. I don't believe it's less efficient or less readable for POD types, but don't want to be fixated about this particular implementation detail.

cmake/Utils/AddLibbacktrace.cmake
```diff
@@ -62,8 +62,9 @@ function (_libbacktrace_compile)
   add_library(libbacktrace STATIC IMPORTED)
   add_dependencies(libbacktrace project_libbacktrace)
   set_target_properties(
-    libbacktrace PROPERTIES IMPORTED_LOCATION ${libbacktrace_prefix}/lib/libbacktrace.a
-                            INTERFACE_INCLUDE_DIRECTORIES ${libbacktrace_prefix}/include
+    libbacktrace
+    PROPERTIES IMPORTED_LOCATION ${libbacktrace_prefix}/lib/libbacktrace.a
+               INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_CURRENT_LIST_DIR}/../../3rdparty/libbacktrace/
```
junrushao, line 67: This change is needed so that clang-tidy is able to find `backtrace.h` statically with `compile_commands.json` without having to build the target

include/tvm/ffi/function.h
```diff
@@ -315,21 +315,21 @@ class Function : public ObjectRef {
    * \param packed_call The packed function signature
    */
   template <typename TCallable>
-  static Function FromPacked(TCallable packed_call) {
+  static Function FromPacked(TCallable&& packed_call) {
```
junrushao: Note: signature change. It's `std::forward`-ed in later use-sites, so I suppose this change should be fine

include/tvm/ffi/optional.h
```diff
@@ -129,13 +129,20 @@ class Optional<T, std::enable_if_t<!use_ptr_based_optional_v<T> && !std::is_same
    * \return the xvalue reference to the stored value.
    * \note only use this function after checking has_value()
    */
-  TVM_FFI_INLINE T&& operator*() && noexcept { return *std::move(data_); }
```
junrushao, line 132: I removed `noexcept` qualifier here give it's totally possible to crash

include/tvm/ffi/function.h
```diff
@@ -763,7 +767,8 @@ struct TypeTraits<TypedFunction<FType>> : public TypeTraitsBase {
   }
 
   TVM_FFI_INLINE static void MoveToAny(TypedFunction<FType> src, TVMFFIAny* result) {
-    TypeTraits<Function>::MoveToAny(std::move(src.packed()), result);
+    // Move from rvalue to trigger TypedFunction rvalue packed() overload
+    TypeTraits<Function>::MoveToAny(std::move(src).packed(), result);
```
junrushao, line 770: Seems plausible change but not entirely sure

src/ffi/function.cc
```diff
@@ -64,14 +66,15 @@ class GlobalFunctionTable {
       metadata_data = String(method_info->metadata.data, method_info->metadata.size);
       func_data = AnyView::CopyFromTVMFFIAny(method_info->method).cast<ffi::Function>();
       this->SyncMethodInfo(method_info->flags);
-      // no need to update method pointer as it would remain the same as func and we retained
```
junrushao, line 67: turns out I initialized `this->method` inside `SyncMethodInfo` just to be safe

include/tvm/ffi/optional.h
```diff
@@ -240,13 +242,13 @@ class Optional<T, std::enable_if_t<std::is_same_v<T, String> || std::is_same_v<T
...
    */
-  TVM_FFI_INLINE T&& operator*() && noexcept { return std::move(data_); }
+  TVM_FFI_INLINE T&& operator*() && { return std::move(data_); }
   /*!
    * \brief Direct access to the value.
    * \return the const reference to the stored value.
    * \note only use this function  after checking has_value()
    */
-  TVM_FFI_INLINE const T& operator*() const& noexcept { return data_; }
+  TVM_FFI_INLINE const T& operator*() const& { return data_; }
```
junrushao: Removed `noexcept` ditto: https://github.com/apache/tvm-ffi/pull/78/files#r2404750184

include/tvm/ffi/function.h
tqchen: this is dangerous, when packed func is a const reference&, std::forward will forward it as a const reference capture, this can cause issue when packed_call goes out of bound. For this particular use case, explcit pass by value, then capture by value(compiler may later optimize it away, or explicitly move capture) the closure is necessary.

include/tvm/ffi/function.h
tqchen: this is correct

include/tvm/ffi/optional.h
tqchen: this delegaes to` constexpr T&& operator*() && noexcept;`, according standard signature it is noexcept

tqchen: https://en.cppreference.com/w/cpp/utility/optional/operator%252A.html

include/tvm/ffi/container/tensor.h
```diff
@@ -335,8 +335,8 @@ class Tensor : public ObjectRef {
    * \param device The device of the Tensor.
    * \return The created Tensor.
    */
-  static Tensor FromDLPackAlloc(DLPackTensorAllocator allocator, ffi::Shape shape, DLDataType dtype,
-                                DLDevice device) {
+  static Tensor FromDLPackAlloc(DLPackTensorAllocator allocator, const ffi::Shape& shape,
```
tqchen: this should change to ffi::ShapeView shape

include/tvm/ffi/function.h
junrushao: This is indeed dangerous to forward expired pointers. Reverted

include/tvm/ffi/optional.h
junrushao: Calling `std::terminate()` is fine to me then

junrushao:
Addressed review comments. Would you love to review again? @tqchen

include/tvm/ffi/string.h
```diff
@@ -924,11 +931,11 @@ struct TypeTraits<std::string>
     return std::string(src->data, src->size);
   }
 
-  TVM_FFI_INLINE static std::string ConvertFallbackValue(Bytes src) {
+  TVM_FFI_INLINE static std::string ConvertFallbackValue(const Bytes& src) {
```
tqchen: convert fallback value is better handled pass by value, because the value is already being generated, and pass by value may enable possible followup opportunities of move

src/ffi/extra/structural_equal.cc
```diff
@@ -317,30 +319,34 @@ class StructEqualHandler {
...
     return false;
   }
 
+  // NOLINTNEXTLINE(performance-unnecessary-value-param)
   bool CompareShape(Shape lhs, Shape rhs) {
     if (lhs.size() != rhs.size()) {
       return false;
     }
-    for (size_t i = 0; i < lhs.size(); ++i) {
+    for (int64_t i = 0, n = static_cast<int64_t>(lhs.size()); i < n; ++i) {
```
tqchen: size_t is fine? this one is a bit strange here

include/tvm/ffi/string.h
junrushao: Looks like it's a complex case. 
- `Byte` can't be moved into `std::string` without an extra copy, which means analyzing locally on `ConvertFallbackValue` doesn't provide sufficient evidence to prefer `Bytes` to `const Bytes&`;
- However, when being invoked, `ConvertFallbackValue`'s arguments are always moved into. That said, `Bytes` is indeed better than `const Bytes&`.

src/ffi/extra/structural_equal.cc
junrushao: It's not a necessary change. Reverting it back

PR #85, opened 2025-10-03
[COMPILE] Fixed error compile when using compiler gcc 8.x

locnd182644:
Screenshot 2025-10-03 153819

Resolved: Add safe_underlying_type_t: check enum before using std::underlying_type


tqchen:
Thanks @locnd182644 , reading a bit more, this is indeed something that can be helpful, i made some direct edits and i think we can bring it in

PR #86, opened 2025-10-03
[ABI] Allow passing tensor and integer to pointer parameter

yaoyaoding:
This PR enables passing tensor or integer to a void pointer parameter.


yaoyaoding:
After discussing with tq, we decide to keep the original convention of parameter passing for void*.

PR #87, opened 2025-10-03
[CYTHON] Fix memory leak when tensor alloc returns to python

tqchen:
This PR fixes a memory leak when tensor alloc inside the function and return to python. We need to delete the chandle to prevent memory leak as the ownership count is transfered to DLManagedTensor.

### DIFF
--- a/tests/python/test_load_inline.py
+++ b/tests/python/test_load_inline.py
@@ -318,3 +318,34 @@ def test_load_inline_both() -> None:
     y_cuda = torch.empty_like(x_cuda)
     mod.add_one_cuda(x_cuda, y_cuda)
     torch.testing.assert_close(x_cuda + 1, y_cuda)
+
+
+@pytest.mark.skipif(
+    torch is None or not torch.cuda.is_available(), reason="Requires torch and CUDA"
+)
+def test_cuda_memory_alloc_noleak() -> None:
+    assert torch is not None
+    mod = tvm_ffi.cpp.load_inline(
+        name="hello",
+        cuda_sources=r"""
+            #include <tvm/ffi/function.h>
+            #include <tvm/ffi/container/tensor.h>
+
+            namespace ffi = tvm::ffi;
+
+            ffi::Tensor return_tensor(tvm::ffi::TensorView x) {
+                ffi::Tensor y = ffi::Tensor::FromDLPackAlloc(
+                    TVMFFIEnvGetTensorAllocator(), x.shape(), x.dtype(), x->device);
+                return y;
+            }
+        """,
+        functions=["return_tensor"],
+    )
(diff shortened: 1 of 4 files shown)


tqchen:
```python
import tvm_ffi
import tvm_ffi.cpp
import torch

def test_cuda_memory_allocated():
    mod = tvm_ffi.cpp.load_inline(
        name="hello",
        cuda_sources=r"""
            #include <tvm/ffi/function.h>

            namespace ffi = tvm::ffi;

            ffi::Tensor add_one_cuda(tvm::ffi::TensorView x) {
                ffi::Tensor y = ffi::Tensor::FromDLPackAlloc(
                    TVMFFIEnvGetTensorAllocator(), x.shape(), x.dtype(), x->device);
                return y;
            }
        """,
        functions=["add_one_cuda"],
    )
    x = torch.arange(1024 * 1024, dtype=torch.float32, device="cuda")
    repeat = 100
    for i in range(repeat):
        y = torch.empty_like(x)
        y = mod.add_one_cuda(x)
        allocated = torch.cuda.memory_allocated()
        print(f"Allocated memory: {allocated / (1024**2):.2f} MB")


test_cuda_memory_allocated()
```

Validated through the script above

cyx-6 (approved):

MasterJH5574:
Is it possible to add the validation script as a skipped test in the codebase?

tqchen:
@MasterJH5574 good point, added a regression test

tests/python/test_load_inline.py
```diff
@@ -318,3 +318,34 @@ def test_load_inline_both() -> None:
...
+        functions=["return_tensor"],
+    )
+    x = torch.arange(1024 * 1024, dtype=torch.float32, device="cuda")
+    current_allocated = torch.cuda.memory_allocated()
+    repeat = 8
+    for i in range(repeat):
+        mod.return_tensor(x)
+        diff = torch.cuda.memory_allocated() - current_allocated
+        # memory should not grow as we loop over
+        assert diff <= 1024**2 * 8
```
aleozlx, line 351: hi i have a naive question: why is the threshold 8MiB?

if return_tensor(x) is created and triggers decref immediately as no var is holding the reference to the returned obj, my naive understanding is that diff = 0, then if so i interpret 8MiB here as some safety margin, but not related to the x size nor loop iteration count, correct?

thanks

tqchen: yes, it is a safety margin, the allocated size is supposely to be 4MB. we are testing that memory do not grow

tqchen: You are right that in theory it should be zero, that can depend on the order of deletion in the loop(and our query)

PR #94, opened 2025-10-08
feat: Introduce flexible `TypeSchema.repr(ty_map)`

junrushao:
This PR adds a new API `TypeSchema.repr(ty_map: ...)` which allows downstream applications to override string representation of a certain type, e.g. override `list` to `Sequence`, `dict` to `Mapping`. This can be useful in schema generation.


python/tvm_ffi/cython/type_info.pxi
```diff
@@ -136,6 +126,29 @@ class TypeSchema:
     def from_json_str(s) -> "TypeSchema":
         return TypeSchema.from_json_obj(json.loads(s))
 
+    def repr(self, ty_map = None) -> str:
```
tqchen, line 129: need to document the function

junrushao: Updated with comprehensive documentation

junrushao: See [python/tvm_ffi/core.pyi‎](https://github.com/apache/tvm-ffi/pull/94/files#diff-8bcddcd9d52d633e1dde14080cf197a8a5ee9a66b7d69235ccd939f94eb10216)

PR #96, opened 2025-10-08
[DLPack] C Functions for DLPack Speed Exchange and Stream Handling

Kathryn-cat:
## Summary of Changes

This PR introduces a unified `DLPackExchangeAPI` struct as described in proposal [175](https://github.com/dmlc/dlpack/issues/175). This new convention replaces the previous mechanism of separate function pointers, and aligns with the latest DLPack standard as shown in PR [174](https://github.com/dmlc/dlpack/pull/174).

Within the new `DLPackExchangeAPI` struct, it also includes a `current_work_stream` function pointer that allows more robust and integrated querying of the current device stream (e.g., CUDA stream) during DLPack tensor exchanges. All the conversion from/to DLPack has been updated to `_no_sync`, meaning you should use `current_work_stream` to explicitly handle stream synchronization. It also includes a non-owning DLTensor conversion to avoid unnecessary reference counting.

Following this change, the Python FFI for PyTorch has been updated to expose the new `DLPackExchangeAPI` struct via `__c_dlpack_exchange_api__` on torch.Tensor.

The `3rdparty/dlpack` has been updated to incorporate the latest commit.

## Benchmark Results

The benchmark results show that the overhead of API by passing in arguments, doing conversion at cpp extension level, and return matches the expected speedup as described in the proposal [175](https://github.com/dmlc/dlpack/issues/175).

Screenshot 2025-10-11 at 1 33 20 PM

## Followup in later PR

This PR does not contain the unit tests for the cpp conversion. There will be a followup PR to add the tests.

### DIFF
--- a/python/tvm_ffi/cython/tensor.pxi
+++ b/python/tvm_ffi/cython/tensor.pxi
@@ -320,10 +320,16 @@ cdef const DLPackExchangeAPI* _dltensor_test_wrapper_get_exchange_api() noexcept
 
     # Initialize function pointers
     _dltensor_test_wrapper_static_api.managed_tensor_allocator = NULL
-    _dltensor_test_wrapper_static_api.managed_tensor_from_py_object_no_sync = <DLPackManagedTensorFromPyObjectNoSync>_dltensor_test_wrapper_from_pyobject
-    _dltensor_test_wrapper_static_api.managed_tensor_to_py_object_no_sync = <DLPackManagedTensorToPyObjectNoSync>_dltensor_test_wrapper_to_pyobject
+    _dltensor_test_wrapper_static_api.managed_tensor_from_py_object_no_sync = (
+        <DLPackManagedTensorFromPyObjectNoSync>_dltensor_test_wrapper_from_pyobject
+    )
+    _dltensor_test_wrapper_static_api.managed_tensor_to_py_object_no_sync = (
+        <DLPackManagedTensorToPyObjectNoSync>_dltensor_test_wrapper_to_pyobject
+    )
     _dltensor_test_wrapper_static_api.dltensor_from_py_object_no_sync = NULL
-    _dltensor_test_wrapper_static_api.current_work_stream = <DLPackCurrentWorkStream>_dltensor_test_wrapper_current_work_stream
+    _dltensor_test_wrapper_static_api.current_work_stream = (
+        <DLPackCurrentWorkStream>_dltensor_test_wrapper_current_work_stream
+    )
 
     return &_dltensor_test_wrapper_static_api
 
@@ -369,7 +375,9 @@ cdef inline object make_ret_dltensor(TVMFFIAny result):
     return tensor
 
 
-cdef inline object make_tensor_from_chandle(TVMFFIObjectHandle chandle, const DLPackExchangeAPI* c_ctx_dlpack_api = NULL):
+cdef inline object make_tensor_from_chandle(
+    TVMFFIObjectHandle chandle, const DLPackExchangeAPI* c_ctx_dlpack_api = NULL
+):
     # TODO: Implement
     cdef Tensor tensor
     cdef void* py_obj
--- a/python/tvm_ffi/cython/tvm_ffi_python_helpers.h
+++ b/python/tvm_ffi/cython/tvm_ffi_python_helpers.h
@@ -270,7 +270,8 @@ class TVMFFIPyCallManager {
    */
   TVM_FFI_INLINE int FuncCall(TVMFFIPyArgSetterFactory setter_factory, void* func_handle,
                               PyObject* py_arg_tuple, TVMFFIAny* result, int* c_api_ret_code,
-                              bool release_gil, const DLPackExchangeAPI** optional_out_ctx_dlpack_api) {
+                              bool release_gil,
+                              const DLPackExchangeAPI** optional_out_ctx_dlpack_api) {
     int64_t num_args = PyTuple_Size(py_arg_tuple);
     if (num_args == -1) return -1;
     try {


python/tvm_ffi/cython/function.pxi
```diff
@@ -142,35 +142,42 @@ cdef int TVMFFIPyArgSetterObject_(
...
+    cdef void* current_stream = NULL
+    cdef const DLPackExchangeAPI* api = this.c_dlpack_exchange_api
 
-    if this.c_dlpack_to_pyobject != NULL:
-        ctx.c_dlpack_to_pyobject = this.c_dlpack_to_pyobject
-    if this.c_dlpack_tensor_allocator != NULL:
-        ctx.c_dlpack_tensor_allocator = this.c_dlpack_tensor_allocator
+    # Set allocator and ToPyObject converter in context if available
+    if api.managed_tensor_allocator != NULL:
+        ctx.c_dlpack_tensor_allocator = <DLPackTensorAllocator>api.managed_tensor_allocator
```
tqchen: instead, change to ctx.dlpack_exchange_api = 

change ctx field

python/tvm_ffi/cython/tensor.pxi
```diff
@@ -275,33 +275,72 @@ _set_class_tensor(Tensor)
...
+# Module-level static DLPackExchangeAPI for DLTensorTestWrapper
+cdef DLPackExchangeAPI _dltensor_test_wrapper_static_api
+cdef bint _dltensor_test_wrapper_api_initialized = False
+
+cdef const DLPackExchangeAPI* _dltensor_test_wrapper_get_exchange_api() noexcept:
+    """Get the static DLPackExchangeAPI instance for DLTensorTestWrapper."""
+    global _dltensor_test_wrapper_static_api, _dltensor_test_wrapper_api_initialized
+
+    if not _dltensor_test_wrapper_api_initialized:
+        # Initialize header using macros from dlpack.h
```
tqchen: given we call it once, no need to do the intialization check, just do the assignment

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -513,16 +508,66 @@ def load_torch_c_dlpack_extension() -> Any:
...
+    // Use handle (non-owning) to avoid unnecessary refcount operations
+    py::handle handle(static_cast<PyObject*>(py_obj));
+    at::Tensor tensor = handle.cast<at::Tensor>();
+
+    // Fill in the pre-allocated DLTensor struct with direct pointers
+    // This is a non-owning conversion - the original PyObject owns the data
+    // and is kept alive by the caller for the duration of this call
+    out->data = tensor.data_ptr();
+    out->device = at::torchDeviceToDLDeviceForDLPackv1(tensor.device());
+    out->ndim = static_cast<int32_t>(tensor.dim());
```
tqchen: move the implementation to aten:: namespace, mainly because likely we need such split in upstream.

at::toDLPackNonOwning(const at::Tensor&, DLTensor*)

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -513,16 +508,66 @@ def load_torch_c_dlpack_extension() -> Any:
   }
 }
 
-int64_t TorchDLPackFromPyObjectPtr() {
-  return reinterpret_cast<int64_t>(TorchDLPackFromPyObject);
+int TorchDLTensorFromPyObject(void* py_obj, DLTensor* out) {
```
tqchen: naming, make it consistent with the new term TorchDLPackManagedTensorFromPyObjectNoSync

python/tvm_ffi/cython/tensor.pxi
Kathryn-cat: I think `DLTensorTestWrapper` called a couple times in `benchmark_dlpack.py`, so the initialization check would make sure that the global `_dltensor_test_wrapper_static_api` is only initialized once

python/tvm_ffi/_optional_torch_c_dlpack.py
Kathryn-cat: updated

python/tvm_ffi/_optional_torch_c_dlpack.py
Kathryn-cat: updated

python/tvm_ffi/cython/tensor.pxi
tqchen: the static initialization of the memebr is only called once

tqchen: specifically

```python
class DLTensorTestWrapper:
   # this is static member init and only called once
   __dlpack_exchange__api = xx
```

Kathryn-cat: updated

python/tvm_ffi/cython/function.pxi
Kathryn-cat: updated

Kathryn-cat:
@tqchen comments are addressed. I think this PR is in good shape now, let's separate the unit tests into another PR

python/tvm_ffi/cython/tvm_ffi_python_helpers.h
```diff
@@ -323,12 +315,15 @@ class TVMFFIPyCallManager {
           return -1;
         }
       }
-      if (prev_tensor_allocator != ctx.c_dlpack_tensor_allocator) {
+      if (ctx.c_dlpack_exchange_api != nullptr &&
```
tqchen, line 319: change optional_out_dlpack_importer to optional_out_ctx_dlpack_api, and pass back dlpack API, update make_ret to take optional_out_ctx_dlpack_api

Kathryn-cat: updated

PR #106, opened 2025-10-13
doc: Auto-extract type annotations from function definition

junrushao:
A new plugin `sphinx-autodoc-typehints` is introduced that helps automatically extract type annotations for function definitions.

It makes it possible, according to DRY (don't repeat yourself) rule, to avoid defining type annotations the 2nd time inside docstring, which is very much prone to inconsistency.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx.ext.ifconfig",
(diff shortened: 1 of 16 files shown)


junrushao:
@tqchen Off the topic, I noticed many python classes/methods that are not exposed in Python reference doc, e.g. `tvm_ffi.dtype`. Is it intentional?

tqchen:
@junrushao for public types we should expose, likely they are oversight

PR #124, opened 2025-10-15
[CMake] Enable automatic cmake config discovery

Ubospica:
Relocate CMake config files from `cmake/` to `share/cmake/tvm_ffi/` to match
CMake's standard search pattern `<prefix>/<name>*/share/cmake/<name>*/`.

This enables `find_package(tvm_ffi CONFIG REQUIRED)` to work automatically
without manually setting `tvm_ffi_ROOT`.

When `find_package(Python)` is called, CMake adds `sys.prefix` to search paths,
allowing it to discover tvm-ffi's config files in the standard location.

This also aligns with pybind's behavior. See also https://github.com/pybind/pybind11/blob/a2c59711b20707d64b7ad61d488e002611369780/.github/CONTRIBUTING.md#L325 and https://github.com/pybind/pybind11/blob/a2c59711b20707d64b7ad61d488e002611369780/pybind11/commands.py#L22.

Changes:
- CMakeLists.txt: Install config to `share/cmake/tvm_ffi/`
- libinfo.py: Update `find_cmake_path()` to check standard path first
- examples: Simplify CMakeLists.txt to use direct `find_package()`

This PR also avoids rebuilding in editable installation mode, and adds a cd command in the readme of examples to ensure user is in the correct directory.

### DIFF
--- a/examples/quick_start/CMakeLists.txt
+++ b/examples/quick_start/CMakeLists.txt
@@ -18,7 +18,14 @@
 cmake_minimum_required(VERSION 3.18)
 project(tvm_ffi_example)
 
-# Try to find tvm-ffi automatically (like pybind11)
+# Discover the Python interpreter so we can query tvm-ffi for its CMake package path.
+find_package(
+  Python
+  COMPONENTS Interpreter
+  REQUIRED
+)
+
+# Find tvm-ffi automatically
 find_package(tvm_ffi CONFIG REQUIRED)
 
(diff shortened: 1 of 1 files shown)


junrushao:
Nice work!! I was about to ask :)

tqchen:
note that this unfortunately does not work for cmake that is outside of scikit build part, so we still need to keep the cmake path for max compact

PR #125, opened 2025-10-15
[CI] Add run_example job to test examples across all platforms

Ubospica:
- Run examples/quick_start, examples/packaging, and examples/inline_module
- Test on Ubuntu (x86_64/ARM), Windows, macOS with different Python versions
- Verify error handling in packaging example


tqchen:
to reduce cost of tests, likely we should not include examples in UT, we can perhaps include some in nightly

PR #126, opened 2025-10-15
[CMake] Add config.cmake

Ubospica:
This PR adds a config.cmake file to conveniently configure the cmake options with file.


tqchen:
Thanks @Ubospica for this particular case, we are minimizing the set of configs so will not include config here

PR #145, opened 2025-10-16
Enable py38 build

oraluben:
Add support for cpython 3.8.

### DIFF
--- a/python/tvm_ffi/base.py
+++ b/python/tvm_ffi/base.py
@@ -29,8 +29,8 @@ logger = logging.getLogger(__name__)
 # ----------------------------
 # Python3 version.
 # ----------------------------
-if not (sys.version_info[0] >= 3 and sys.version_info[1] >= 9):
-    PY3STATEMENT = "The minimal Python requirement is Python 3.9"
+if not (sys.version_info[0] >= 3 and sys.version_info[1] >= 8):
+    PY3STATEMENT = "The minimal Python requirement is Python 3.8"
     raise Exception(PY3STATEMENT)
 
 # ----------------------------


tqchen:
on the torch side, perhaps you need to use macro to disable some of the optional path, the dlpack have fallback, but would be good to have the optional fallback to compile, to enable faster exchange

pyproject.toml
```diff
@@ -183,6 +183,7 @@ select = [
   "RUF", # ruff, https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf
   "NPY", # numpy, https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy
   "F",   # pyflakes, https://docs.astral.sh/ruff/rules/#pyflakes-f
+  "FA",
```
junrushao: let's add a link so it's easier for other maintainers to look up

```suggestion
  "FA",  # flake8-future-annotations, https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa
```

tqchen:
see example here https://github.com/apache/tvm-ffi/blob/main/python/tvm_ffi/_optional_torch_c_dlpack.py#L123

junrushao:
Overall looks awesome to me! Please fix the lint via `pre-commit` hooks:

```bash
uv pip install pre-commit
pre-commit run --all-files
```

Also, torch warnings are likely fixed if you'd love to try guarding with torch version macros, e.g. 

https://github.com/oraluben/tvm-ffi/blob/a942c7e323e60c50f8ecc82a7c5a460bfa91fb25/python/tvm_ffi/_optional_torch_c_dlpack.py#L125-L134

We're super happy to take it in regardless of the torch warning issue

oraluben:
This should be ready for review.

Actually our project still uses `torch.library.custom_op`, which is since torch 2.4 so I didn't test torch<=2.3 😂

junrushao:
Thanks for helping with cp38 support! 🎉

PR #150, opened 2025-10-16
chore(build): Skip Python 3.8 on arm64 macOS

junrushao:
ARM64 macOS doesn't have native support for Python 3.8. Therefore, it is pointless to release Python 3.8 wheels in this case.


tqchen (approved):

oraluben:
Python 3.8 provides arm support since [3.8.10](https://www.python.org/downloads/release/python-3810/), and both uv and brew provides python 3.8 for mac aarch64.

Both uv and brew also provide python 3.8 aarch64 package on Macos.

PR #159, opened 2025-10-17
MSVC compiler fix

SystemPanic:
This pull request adds a forward declaration of Tensor class to prevent errors when compiling with MSVC.

### DIFF
--- a/include/tvm/ffi/container/tensor.h
+++ b/include/tvm/ffi/container/tensor.h
@@ -38,6 +38,8 @@
 namespace tvm {
 namespace ffi {
 
+class Tensor;
+
 /*!
  * \brief Check if the device uses direct address, where address of data indicate alignment.
  * \param device The input device.


junrushao:
Could you help elaborate a little bit? I don't think we are using `cute::Tensor` anywhere in this repo

SystemPanic:
@junrushao You are right, cute::Tensor comes from the project where tvm_ffi was included, FlashInfer, not from tvm_ffi standalone compilation itself. 

MSVC just was looking for a definition of Tensor.

PR description updated

Anyway, compilation error is still valid, forward declaration is required to avoid this type of issues with MSVC, as is being used at https://github.com/apache/tvm-ffi/blob/39d675d5f9674345c040aae99b3b3a67b3f270a2/include/tvm/ffi/container/tensor.h#L172 before the class declaration at https://github.com/apache/tvm-ffi/blob/39d675d5f9674345c040aae99b3b3a67b3f270a2/include/tvm/ffi/container/tensor.h#L244

tqchen:
I think it is not strictly necessarily, but helpful when there are conflicting types named tensor and being used in the root namespace

tqchen (approved):

PR #169, opened 2025-10-18
[Docs] Add kernel library guide

cyx-6:
This PR adds the kernel library guide for composing or migrating a kernel library

### DIFF
--- a/docs/guides/kernel_library_guide.rst
+++ b/docs/guides/kernel_library_guide.rst
@@ -43,7 +43,7 @@ It is **recommended** to use :cpp:class:`~tvm::ffi::TensorView` when possible, t
 Tensor Attributes
 -----------------
 
-For the sake of convenience, :cpp:class:`~tvm::ffi::TensorView` and :cpp:class:`~tvm::ffi::Tensor` align the following attributes retrieval mehtods to :cpp:class:`at::Tensor` interface, to obtain tensor basic attributes and storage pointer:
+For convenience, :cpp:class:`~tvm::ffi::TensorView` and :cpp:class:`~tvm::ffi::Tensor` align the following attributes retrieval mehtods to :cpp:class:`at::Tensor` interface, to obtain tensor basic attributes and storage pointer:
 ``dim``, ``dtype``, ``sizes``, ``size``, ``strides``, ``stride``, ``numel``, ``data_ptr``, ``device``, ``is_contiguous``
 
 Please refer to the documentation of both tensor classes for their details. Here  highlight some non-primitive attributes:


docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Both ``ffi::Tensor`` and ``ffi::TensorView`` are designed to represent tensors in TVM FFI eco-system. The main difference is whether it is an owning tensor pointer.
+
+:ffi::Tensor:
+ ``ffi::Tensor`` is a completely onwing tensor pointer, pointing to TVM FFI tensor object. TVM FFI handles the lifetime of ``ffi::Tensor`` by retaining a strong reference.
+
+:ffi::TensorView:
+ ``ffi::TensorView`` is a light weight non-owning tnesor pointer, pointeing to a TVM FFI tensor or external tensor object. TVM FFI does not retain its reference. So users are responsible for ensuring the lifetime of tensor object to which the ``ffi::TensorView`` points.
+
+TVM FFI can automatically convert the input tensor at Python side, e.g. ``torch.Tensor``, to both ``ffi::Tensor`` or ``ffi::TensorView`` at C++ side, depends on the C++ function arguments. However, for more flexibility and better compatibility, we **recommand** to use ``ffi::TensorView`` in practice. Since some frameworks, like JAX, cannot provide strong referenced tensor, as ``ffi::Tensor`` expected.
```
yzh119: > Since some frameworks, like JAX, cannot provide strong referenced tensor,

Can you provide reference?

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Tensor as Argument
+------------------
+
+Typically, we expect that all tensors are pre-allocated at Python side and passed in via TVM FFI, including the output tensor. And TVM FFI will convert them into ``ffi::TensorView`` at runtime. For the optional arguments, ``ffi::Optional`` is the best practice. Here is an example of a kernel definition at C++ side and calling at Python side.
```
yzh119: might be off-topic, but does tvm-ffi supports std::optional?

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+:ffi::Tensor:
+ ``ffi::Tensor`` is a completely onwing tensor pointer, pointing to TVM FFI tensor object. TVM FFI handles the lifetime of ``ffi::Tensor`` by retaining a strong reference.
+
+:ffi::TensorView:
+ ``ffi::TensorView`` is a light weight non-owning tnesor pointer, pointeing to a TVM FFI tensor or external tensor object. TVM FFI does not retain its reference. So users are responsible for ensuring the lifetime of tensor object to which the ``ffi::TensorView`` points.
+
+TVM FFI can automatically convert the input tensor at Python side, e.g. ``torch.Tensor``, to both ``ffi::Tensor`` or ``ffi::TensorView`` at C++ side, depends on the C++ function arguments. However, for more flexibility and better compatibility, we **recommand** to use ``ffi::TensorView`` in practice. Since some frameworks, like JAX, cannot provide strong referenced tensor, as ``ffi::Tensor`` expected.
+
+Tensor as Argument
```
yzh119: The title looks confusing to me, how about passing style? (e.g. if output tensor is passed to the function, it's destination passing style).

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+    void FreeData(DLTensor* tensor) { nvshmem_free(tensor->data); }
+  };
+
+  // Allocator usage
+  ffi::Tensor cpu_tensor = ffi::Tensor::FromNDAlloc(CPUNDAlloc(), ...);
+  ffi::Tensor cuda_tensor = ffi::Tensor::FromNDAlloc(CUDANDAlloc(), ...);
+  ffi::Tensor nvshmem_tensor = ffi::Tensor::FromNDAlloc(NVSHMEMNDAlloc(), ...);
+
+:FromEnvAlloc:
+ For the case of using external tensor allocator, like``at::empty`` in PyTorch C++ extensions, ``FromEnvAlloc`` is the better choice. Besides of the basic attributes like shape, data type and device, it requires a thread-local environmental allocator ``TVMFFIEnvTensorAlloc``. ``TVMFFIEnvTensorAlloc`` gets the global tensor allocator in the current context. The context can be switched based on the arguments of the kernel.
```
yzh119: "external tensor allocator" -> "framework tensor allocator"

It's better to mention the advantage of using framework allocator:
1. take advantage of framework's native caching allocator: https://zdevito.github.io/2022/08/04/cuda-caching-allocator.html 
2. it's easier to track memory usage: if we use `cudaMalloc`, then memory allocated from tvm-ffi is not visible to torch.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+:FromDLPack:
+ ``FromDLPack`` enables creating ``ffi::Tensor`` from ``DLManagedTensor*``.
+
+:FromDLPackVersioned:
```
yzh119: Why do we need this, please explain.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+======
+
+TVM FFI maintains the stream context per device type and index. Use ``TVMFFIEnvGetStream`` to get the current stream on device:
+
+.. code-block:: c++
+
+ ffi::DLDevice device = input.device();
+ cudaStream_t stream = reinterpret_cast<cudaStream_t>(TVMFFIEnvGetStream(device.device_type, device.device_id));
+
+Similar to ``TVMFFIEnvTensorAlloc``, TVM FFI updates the context stream based on the arguments of the kernel, by calling ``TVMFFIEnvSetStream``.
```
yzh119: Provide some examples

docs/guides/kernel_library_guide.rst
tqchen: we can support std::optional, `ffi::Optional` is mainly to stablize the ABI of the optional and saves some space for now, but we can update to also support std::optional as well as arguments.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+.. "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+
+====================
+Kernel Library Guide
+====================
+
+This guide serves as a quick start for composing a kernel from scratch, or migrating a kernel from externel frameworks. It covers the core concepts in TVM FFI, such as tensor, stream.
```
tqchen: quick start for shipping python version and framework agnostic kernel libraries with TVM FFI

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Tensor
+======
+
+Tensor is the most important input for a kernel libaray. In PyTorch C++ extensions, kernel library usually takes ``at::Tensor`` as tensor input. In TVM FFI, we introduce two types of tensor, ``ffi::Tensor`` and ``ffi::TensorView``.
```
tqchen: In TVM FFI, we support two types of tensor constructs,  ``ffi::Tensor`` and ``ffi::TensorView`` that can be used to represent a tensor from machine learning frameworks

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Tensor and TensorView
+---------------------
+
+Both ``ffi::Tensor`` and ``ffi::TensorView`` are designed to represent tensors in TVM FFI eco-system. The main difference is whether it is an owning tensor pointer.
```
tqchen: represent tensors from ML frameworks that interact with the TVM FFI ABI.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+:ffi::Tensor:
+ ``ffi::Tensor`` is a completely onwing tensor pointer, pointing to TVM FFI tensor object. TVM FFI handles the lifetime of ``ffi::Tensor`` by retaining a strong reference.
+
+:ffi::TensorView:
+ ``ffi::TensorView`` is a light weight non-owning tnesor pointer, pointeing to a TVM FFI tensor or external tensor object. TVM FFI does not retain its reference. So users are responsible for ensuring the lifetime of tensor object to which the ``ffi::TensorView`` points.
```
tqchen: non-owning view of an existing tensor. It is backed by DLTensor structure in DLPack

docs/guides/kernel_library_guide.rst
tqchen: we **recommend* to use ``ffi::TensorView`` when possible, that helps us to support more cases, including cases where only view but not strong reference are passed. It is also more lightweight.

tqchen: XLA Buffer is weak reference, we can "fake" a strong reference by `pass_owning_tensor=True` now, but the tensor cannot be retained or returned.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+.. code-block:: python
+
+ # Kernel calling
+ input = torch.tensor(...)
+ output = torch.empty(...)
+ workspace = torch.empty(...)
+ func(input, None, output, workspace)
+
+Ideally, we expect the kernel function to have ``void`` as return type. However, if it is necessary to return the ``ffi::Tensor`` anyway, please pay attention to convert the output ``ffi::Tensor`` to original tensor type at Python side, like ``torch.from_dlpack``.
```
tqchen: the FFI automatically convert the returning tensor to torch.Tensor when input contains torch arguments

docs/guides/kernel_library_guide.rst
tqchen: I agree, let us move `FromEnvAlloc` first, we encourage `FromEnvAlloc` when possible, since it makes use of the allocator from framework tensor allocator. Give an example code

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+:FromDLPack:
+ ``FromDLPack`` enables creating ``ffi::Tensor`` from ``DLManagedTensor*``.
+
+:FromDLPackVersioned:
+ ``FromDLPackVersioned`` enables creating ``ffi::Tensor`` from ``DLManagedTensorVersioned*``.
+
+Stream
+======
+
+TVM FFI maintains the stream context per device type and index. Use ``TVMFFIEnvGetStream`` to get the current stream on device:
```
tqchen: When calling from frameworks like PyTorch, `TVMFFIEnvGetStream` will return the stream under the current torch stream context

tqchen (changes requested):

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,161 @@
...
+DLDataType
+ In TVM FFI, tensor data types are stored as ``DLDataType`` which is defined by DLPack protocol.
+
+DLDevice
+ In TVM FFI, tensor device information are stored as ``DLDevice`` which is defined by DLPack protocol.
+
+ShapeView
+ In TVM FFI, tensor shapes and strides attributes retrieval are returned as ``ShapeView``. It is an iterate-able data structure storing the shapes or strides data as ``int64_t`` array.
+
+Tensor Allocation
```
yaoyaoding, line 60: Do we support allocate some zero-initialized tensor?

I ask this question because there is a usage that requires the workspace to be zero-initailized before we launch the kernel - like the semaphores used in split_k optimization of matmul in cutlass. If we have an extra memset-async kernel, it will introduce some overhead. I don't have some good idea on this. Maybe manually set to zero when first launch the kernel and reserve it to the kernel provider.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,161 @@
...
+-----------------
+
+TVM FFI provides several methods to allocate tensors, when dynamic tensor allocation is necessary.
+
+FromEnvAlloc
```
yaoyaoding, line 68: Any details related to stream? Since the kernel launch will be async, same as the memory allocation. The alloc will be issued on the current cuda stream of the input tensor's device? What happens if they are on different devices?

docs/guides/kernel_library_guide.rst
yzh119: When creating a zero tensor in frameworks, the most common behavior is to first allocate the tensor and then fill it with zeros (either using a kernel or memset), correct? I'm inclined to let users apply zero_() in their own ways (users can modify tensors however they want once allocated), which would be more flexible (e.g. they can choose whether to use cumemset or kernel call, which stream to use and whether using pdl, etc) than having the TVM FFI handle it automatically.

yaoyaoding: I see, sounds good to me! I also feel that passing the tensor explicitly by the user is a good design choice.

Another approach that I can think is allow the kernel provider to manage it. Currently, in hidet/tilus, we have a runtime library shipped with hidet package to have such functionality (like request a zero-initialized workspace). I want to remove the hidet runtime library completely in the future to make kernels easy to ship tilus-generated kernels. It would be good to have a general mechanism in tvm-ffi to provide a general key-based storage to the kernel providers to use like:
```c++
std::optional<tvm::ffi::Any> tvm::ffi::getStorageObject(std::string key);
void tvm::ffi::setStorageObject(std::string key, std::ffi::Any);
```
so that we can manage the internal storage easily:
```c++
// before kernel launch
...  
workspace = tvm::ffi::getStorageObject("tilus.workspace") 
if (not workspace.has_value()) { 
   workspace = ffi::Tensor::FromEnvAlloc(TVMFFIEnvTensorAlloc, ...);
   // any way to memset via tvm-ffi?
   tvm::ffi::setStorageObject("tilus.workspace", workspace);
}
...
```
Our current recommendation still requires a runtime library (https://tvm.apache.org/ffi/guides/compiler_integration.html#recommended-approach-for-state-management)

cc @tqchen

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,192 @@
...
+which is equivalent to:
+
+.. code-block:: c++
+
+ at::Tensor tensor = at::empty(...);
+
+FromNDAlloc
+^^^^^^^^^^^
+
+``FromNDAlloc`` is the most basic tensor allocator. It is designed for simple cases where framework tensor allocator is no longer needed. ``FromNDAlloc`` just requires a custom allocator struct to handle the tensor allocation and free, with fixed interface ``void AllocData(DLTensor*)`` and ``void FreeData(DLTensor*)`` methods. Here are the examples of CPU, CUDA and NVSHMEM allocation:
```
tqchen: add a note stating that if we are returning arrays that are allocated by `FromNDAlloc` to caller, we need to make sure the array does not outlive the runtime.Module, because its deleter points to a function pointer in the DLL. This can typically be done by retaining the runtime.Module globally, or for the period of time. 
We always recommend using `ffi::Tensor::FromEnvAlloc(TVMFFIEnvTensorAlloc, ...)` when possible

yaoyaoding:
I polished the doc a little bit:
1. add more hyperlinks
2. move the example of NDAlloc from the guide to the function comment, to show that we recommend the first approach to allocate tensor.

tqchen:
looks good, @junrushao @yzh119 please take another look would be good to add a note to FromNDAlloc(c++ doc) to remark about we encourage using FromEnvAlloc when defining kernels. Let us fix the docs error and get it landed

cyx-6:
doc errors have been fixed

docs/guides/kernel_library_guide.rst
cyx-6: added

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,153 @@
...
+
+Actually, it is **recommended** to pre-allocated input and output tensors from framework at Python side alreadly. So that the return type of kernel functions at C++ side should be ``void`` always.
+
+Context Inherit
+---------------
+
+Also, when calling our kernel library at Python side, we usually need to pass the important context to the kernel library, for example, the CUDA stream context from ``torch.cuda.stream`` or ``torch.cuda.graph``. So that the kernels can be dispatched to the expected CUDA stream. TVM FFI has already made it by maintaining the stream context table per device type and index. And when converting the framework tensors as mentioned above, TVM FFI automatically updates the stream context table, by the device on which the converted framework tensors. For example, if there is an framework tensor as ``torch.Tensor(device="cuda:3")``, TVM FFI will automatically update the current stream of cuda device 3 to torch current context stream, by ``TVMFFIEnvSetStream``. And then at C++ side, we just use ``TVMFFIEnvGetStream`` to get the updated current stream on the specific device. Here is an example:
```
junrushao: Let's have a section explicitly talking about streams right after `Tensor Allocation`

cyx-6: updated and decomposed the stream section into minor subsections

include/tvm/ffi/container/tensor.h
```diff
@@ -350,6 +350,50 @@ class Tensor : public ObjectRef {
   bool IsAligned(size_t alignment) const { return tvm::ffi::IsAligned(*get(), alignment); }
   /*!
    * \brief Create a Tensor from a NDAllocator.
+   *
+   *
+   * Example usage:
```
tqchen, line 360: Add note here:

When building a kernel library, we always recommend use FromEnvAlloc when possible to allocate intermediate Tensors. When a loaded module returns an allocated Tensor to the caller, we need to keep the module alive before the returned tensor get deallocated, because its deleter is defined within the module. FromNDAlloc can be used by C++ applications and runtimes to create Tensors.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. code-block:: c++
+
+ at::Tensor tensor = at::empty(...);
+
+FromNDAlloc
+^^^^^^^^^^^
+
+:cpp:func:`tvm::ffi::Tensor::FromNDAlloc` can be used to create a tensor with custom memory allocator. It's used by the kernel provider if they don't want to rely on the framework tensor allocator. Instead, they provide their own custom allocator and deleter for tensor allocation and free. However, the tensors allocated by ``FromNDAlloc`` only retain the function pointer to its custom deleter for deconstruction. The custom deletes are all owned by the kernel library still. So it is important to make sure the loaded kernel library, :py:class:`tvm_ffi.Module`, outlives the tensors allocated by ``FromNDAlloc``. Otherwise, the function pointers to the custom deleter will be invalid. Here a typical approach is to retain the loaded :py:class:`tvm_ffi.Module` globally or for the period of time. But ``FromEnvAlloc`` is free of this issue, which is more **recommended** in practice.
+
```
tqchen, line 94: FromNDAlloc can be used in linked runtime libraries and c++ applications -- these scenarios keeps the library alive globally throughout the entire lifetime of the process

tqchen:
@junrushao @yzh119 @yaoyaoding please take a final pass reviewing, let us get it in

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+.. "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+
+====================
+Kernel Library Guide
+====================
+
+This guide serves as a quick start for shipping python version and machine learning(ML) framework agnostic kernel libraries with TVM FFI. With the help of TVM FFI, we can connect the kernel libraries to multiple ML framework, such as PyTorch, XLA, JAX, together with the minimal efforts.
```
yaoyaoding:
```suggestion
This guide serves as a quick start for shipping kernel libraries with TVM FFI. The shipped kernel libraries are python version and ML framework agnostic. With the help of TVM FFI, we can connect the kernel libraries to multiple ML framework, such as PyTorch, XLA, JAX, together with the minimal efforts.
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+Tensor
+======
+
+Almost all kernel libraries are about tensor computation and manipulation. For better adaptation to different ML frameworks, TVM FFI provides a minimal set of data structures to represent tensors from ML frameworks, including the tensor basic attributes and storage pointer.
+To be specific, in TVM FFI, two types of tensor constructs, :cpp:class:`~tvm::ffi::Tensor` and :cpp:class:`~tvm::ffi::TensorView`, can be used to represent a tensor from ML frameworks.
+
+Tensor and TensorView
+---------------------
+
+Though both :cpp:class:`~tvm::ffi::Tensor` and :cpp:class:`~tvm::ffi::TensorView` are designed to represent tensors from ML frameworks that interact with the TVM FFI ABI. They are backed by the `DLTensor` in DLPack in practice. The main difference is whether it is an owning tensor structure.
```
yaoyaoding:
```suggestion
Both :cpp:class:`~tvm::ffi::Tensor` and :cpp:class:`~tvm::ffi::TensorView` are designed to represent tensors from ML frameworks that interact with the TVM FFI ABI. They are backed by the `DLTensor` in DLPack in practice. The main difference is whether it is an owning tensor structure.
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+:cpp:class:`tvm::ffi::Tensor`
+ :cpp:class:`~tvm::ffi::Tensor` is a completely owning tensor with reference counting. It can be created and passed between C++ and Python side safely. When the counting reference goes to zero, its underlying deleter function will be called to free the tensor storage.
```
yaoyaoding:
```suggestion
 :cpp:class:`~tvm::ffi::Tensor` is a completely owning tensor with reference counting. It can be created on either C++ or Python side, and we use a reference count to track the number of references. It can be passed between C++ and Python. When the reference count goes to zero, its underlying deleter function will be called to free the tensor storage.
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+:cpp:class:`tvm::ffi::TensorView`
+ :cpp:class:`~tvm::ffi::TensorView` is a non-owning view of an existing tensor, pointing to an existing tensor (e.g., a tensor allocated by PyTorch).
+
+It is **recommended** to use :cpp:class:`~tvm::ffi::TensorView` when possible, that helps us to support more cases, including cases where only view but not strong reference are passed, like XLA buffer.
+It is also more lightweight. However, since :cpp:class:`~tvm::ffi::TensorView` is a non-owning view, it is the user's responsibility to ensure the lifetime of underlying tensor data and attributes of the viewed tensor object.
```
yaoyaoding:
```suggestion
It is also more lightweight. However, since :cpp:class:`~tvm::ffi::TensorView` is a non-owning view, it is the user's responsibility to ensure the lifetime of underlying tensor.
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Tensor Attributes
+-----------------
+
+For the sake of convenience, :cpp:class:`~tvm::ffi::TensorView` and :cpp:class:`~tvm::ffi::Tensor` align the following attributes retrieval mehtods to :cpp:class:`at::Tensor` interface, to obtain tensor basic attributes and storage pointer:
+``dim``, ``dtype``, ``sizes``, ``size``, ``strides``, ``stride``, ``numel``, ``data_ptr``, ``device``, ``is_contiguous``
```
yaoyaoding, line 47:
```suggestion
``dim``, ``dtype``, ``sizes``, ``size``, ``strides``, ``stride``, ``numel``, ``data_ptr``, ``device``, ``is_contiguous``. Please refer to the documentation of both tensor classes for their details. We want to highlight the following attributes:
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+FromEnvAlloc
+^^^^^^^^^^^^
+
+To better adapt to the ML framework, it is **recommended** to reuse the framework tensor allocator anyway, instead of directly allocating the tensors via CUDA runtime API, like ``cudaMalloc``. Since reusing the framework tensor allocator:
+
+* Benefit from the framework's native caching allocator or related allocation mechanism.
+* Help framework tracking memory usage and planning globally.
+
+For this case, TVM FFI provides :cpp:func:`tvm::ffi::Tensor::FromEnvAlloc`. It internally calls the framework tensor allocator. To determine which framework tensor allocator, TVM FFI infers it from the passed-in framework tensors. For example, when calling the kernel library at Python side, there is an input framework tensor if of type ``torch.Tensor``, TVM FFI will automatically bind the :cpp:func:`at::empty` as the current framework tensor allocator by ``TVMFFIEnvTensorAlloc``. And then the :cpp:func:`~tvm::ffi::Tensor::FromEnvAlloc` is calling the :cpp:class:`at::empty` actually:
```
yaoyaoding:
```suggestion
TVM FFI provides :cpp:func:`tvm::ffi::Tensor::FromEnvAlloc` to allocate a tensor with the framework tensor allocator. To determine which framework tensor allocator, TVM FFI infers it from the passed-in framework tensors. For example, when calling the kernel library at Python side, there is an input framework tensor if of type ``torch.Tensor``, TVM FFI will automatically bind the :cpp:func:`at::empty` as the current framework tensor allocator by ``TVMFFIEnvTensorAlloc``. And then the :cpp:func:`~tvm::ffi::Tensor::FromEnvAlloc` is calling the :cpp:class:`at::empty` actually:
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+which is equivalent to:
+
+.. code-block:: c++
+
+ at::Tensor tensor = at::empty(...);
+
+FromNDAlloc
+^^^^^^^^^^^
+
+:cpp:func:`tvm::ffi::Tensor::FromNDAlloc` can be used to create a tensor with custom memory allocator. It's used by the kernel provider if they don't want to rely on the framework tensor allocator. Instead, they provide their own custom allocator and deleter for tensor allocation and free. However, the tensors allocated by ``FromNDAlloc`` only retain the function pointer to its custom deleter for deconstruction. The custom deletes are all owned by the kernel library still. So it is important to make sure the loaded kernel library, :py:class:`tvm_ffi.Module`, outlives the tensors allocated by ``FromNDAlloc``. Otherwise, the function pointers to the custom deleter will be invalid. Here a typical approach is to retain the loaded :py:class:`tvm_ffi.Module` globally or for the period of time. But ``FromEnvAlloc`` is free of this issue, which is more **recommended** in practice.
```
yaoyaoding:
```suggestion
:cpp:func:`tvm::ffi::Tensor::FromNDAlloc` can be used to create a tensor with custom memory allocator. It's used by the kernel provider if they don't want to rely on the framework tensor allocator. Instead, they provide their own custom allocator and deleter for tensor allocation and free. However, the tensors allocated by ``FromNDAlloc`` only retain the function pointer to its custom deleter for deconstruction. The custom deletes are all owned by the kernel library still. So it is important to make sure the loaded kernel library, :py:class:`tvm_ffi.Module`, outlives the tensors allocated by ``FromNDAlloc``. Otherwise, the function pointers to the custom deleter will be invalid. Here a typical approach is to retain the loaded :py:class:`tvm_ffi.Module` globally or for the period of time. 
:cpp:func:`tvm::ffi::Tensor::FromNDAlloc` can also be used in linked runtime libraries of the kernel provider. As long as the runtime library is not unloaded, the deleter will be valid. This allows the kernel provider to implement custom tensor allocator.
In general, ``FromEnvAlloc`` is free of this issue, which is more **recommended** in practice.
```

yaoyaoding: Merged TQ's suggestion into this paragraph.

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Function Exporting
+==================
+
+As we already have our kernel library wrapped with TVM FFI interface, our next and final step is exporting kernel library to Python side. TVM FFI provides macro :c:macro:`TVM_FFI_DLL_EXPORT_TYPED_FUNC` for exporting the kernel functions to the output library files. So that at Python side, it is possible to load the library files and call the kernel functions directly. For example, we export our kernels as:
+
+.. code-block:: c++
+
+ void func(ffi::TensorView input, ffi::TensorView output);
+ TVM_FFI_DLL_EXPORT_TYPED_FUNC(func, func);
```
yaoyaoding:
```suggestion
 // first `func` is the exported name
 // second `func` is the function to be exported
 TVM_FFI_DLL_EXPORT_TYPED_FUNC(func, func);  
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,167 @@
...
+And then we compile the sources into ``func.so``, or ``func.dylib`` for macOS, or ``func.dll`` for Windows. Finally, we can load and call our kernel functions at Python side as:
+
+.. code-block:: python
+
+ mod = tvm_ffi.load_module("func.so")
+ x = ...
+ y = ...
+ mod.func(x, y)
+
+``x`` and ``y`` here can be any ML framework tensors, such as ``torch.Tensor``, ``numpy.NDArray``, ``cupy.ndarray``, or other tensors as long as TVM FFI supports. TVM FFI detects the tensor types in arguments and converts them into :cpp:class:`~tvm::ffi::TensorView` automatically. So that we do not have to write the specific conversion codes per framework.
```
yaoyaoding:
```suggestion
``x`` and ``y`` here can be any ML framework tensors, such as ``torch.Tensor``, ``numpy.NDArray``, ``cupy.ndarray``, or other tensors as long as TVM FFI supports. TVM FFI detects the tensor types in arguments and converts them into :cpp:class:`~tvm::ffi::TensorView` or  :cpp:class:`~tvm::ffi::Tensor`  automatically. So that we do not have to write the specific conversion codes per framework.
```

Do we need add Tensor here?

docs/conf.py
```diff
@@ -463,3 +465,6 @@ def footer_html() -> str:
 
 
 html_css_files = ["custom.css"]
+
+
+show_warning_types = True
```
yaoyaoding:
```suggestion
```

We don't need this anymore. I was using it to debug.

include/tvm/ffi/container/tensor.h
cyx-6: added

docs/guides/kernel_library_guide.rst
cyx-6: updated

docs/guides/kernel_library_guide.rst
cyx-6: updated

docs/guides/kernel_library_guide.rst
cyx-6: removed

docs/guides/kernel_library_guide.rst
cyx-6: updated

docs/guides/kernel_library_guide.rst
cyx-6: removed

docs/guides/kernel_library_guide.rst
cyx-6: updated

docs/guides/kernel_library_guide.rst
cyx-6: refactored

docs/guides/kernel_library_guide.rst
cyx-6: renamed to reduce confusion

docs/guides/kernel_library_guide.rst
cyx-6: added

docs/conf.py
cyx-6: removed

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,171 @@
...
+
+:cpp:class:`tvm::ffi::TensorView`
+ :cpp:class:`~tvm::ffi::TensorView` is a non-owning view of an existing tensor, pointing to an existing tensor (e.g., a tensor allocated by PyTorch).
+
+It is **recommended** to use :cpp:class:`~tvm::ffi::TensorView` when possible, that helps us to support more cases, including cases where only view but not strong reference are passed, like XLA buffer. It is also more lightweight. However, since :cpp:class:`~tvm::ffi::TensorView` is a non-owning view, it is the user's responsibility to ensure the lifetime of underlying tensor.
+
+Tensor Attributes
+-----------------
+
+For the sake of convenience, :cpp:class:`~tvm::ffi::TensorView` and :cpp:class:`~tvm::ffi::Tensor` align the following attributes retrieval mehtods to :cpp:class:`at::Tensor` interface, to obtain tensor basic attributes and storage pointer:
```
junrushao:
```suggestion
For convenience, :cpp:class:`~tvm::ffi::TensorView` and :cpp:class:`~tvm::ffi::Tensor` align the following attributes retrieval mehtods to :cpp:class:`at::Tensor` interface, to obtain tensor basic attributes and storage pointer:
```

docs/guides/kernel_library_guide.rst
```diff
@@ -0,0 +1,171 @@
...
+``dim``, ``dtype``, ``sizes``, ``size``, ``strides``, ``stride``, ``numel``, ``data_ptr``, ``device``, ``is_contiguous``
+
+Please refer to the documentation of both tensor classes for their details. Here  highlight some non-primitive attributes:
+
+:c:struct:`DLDataType`
```
junrushao, line 51: Just to confirm - did you try offline and is this going to render properly? I met some issue a few months ago so just to double check

docs/guides/kernel_library_guide.rst
cyx-6: removed

docs/guides/kernel_library_guide.rst
cyx-6: yes, I have verified locally and it rendered correctly. Here, to obtain the correct tags, just using
`python -m sphinx.ext.intersphinx https://docs.pytorch.org/cppdocs/objects.inv`
and make sure `conf.py` is updated as well.

PR #177, opened 2025-10-20
doc: Handle builtin and inheritance

junrushao:
This PR does two things to the doc:
- **Change A**. Skip member methods of builtin classes;
- **Change B**. Clearly mark which base class a method inherits from.

**Change A**. To give an example, `tvm_ffi.dtype`, as a subclass of Python's native `str`, is shown to have the following methods inherited from `str`:

| Method | Description |
|---|---|
| `capitalize()` | Return a capitalized version of the string. |
| `casefold()` | Return a version of the string suitable for caseless comparisons. |
| `center(width[, fillchar])` | Return a centered string of length `width`. |
| `count(sub[, start[, end]])` | Return the number of non-overlapping occurrences of substring `sub` in `S[start:end]`. |
| `encode([encoding, errors])` | Encode the string using the codec registered for `encoding`. |
| `endswith(suffix[, start[, end]])` | Return `True` if the string ends with the specified `suffix`, `False` otherwise. |
| `expandtabs([tabsize])` | Return a copy where all tab characters are expanded using spaces. |
| `find(sub[, start[, end]])` | Return the lowest index in `S` where substring `sub` is found, such that `sub` is contained within `S[start:end]`. |
| `format(*args, **kwargs)` | Return a formatted version of the string, using substitutions from `args` and `kwargs`. |
| `format_map(mapping, /)` | Return a formatted version of the string, using substitutions from `mapping`. |

, which adds quite a lot of noise because they are irrelevant to `dtype` usecases.

**Change B**. For inherited methods, e.g. `Tensor.same_as` which is from `Object.same_as` - currently it's not clickable and doesn't have a detailed documentation for it, which is quite inconvenient. This PR replaces its description to:

```
Defined in [Object](link/to/reference/python/generated/tvm_ffi.Object.html#tvm_ffi.Object).`
```

### DIFF
--- a/python/tvm_ffi/__init__.py
+++ b/python/tvm_ffi/__init__.py
@@ -101,7 +101,11 @@ def _update_module() -> None:
     for name in __all__:
         obj = globals()[name]
         if not getattr(obj, "__module__", "tvm_ffi").startswith("tvm_ffi"):
-            obj.__module__ = "tvm_ffi"
+            try:
+                obj.__module__ = "tvm_ffi"
+            except (AttributeError, TypeError):
+                # some types don't allow setting __module__
+                pass
 
 
 _update_module()


python/tvm_ffi/__init__.py
```diff
@@ -95,3 +95,9 @@
     "use_raw_stream",
     "use_torch_stream",
 ]
+
+for _name in __all__:
+    try:
+        setattr(globals()[_name], "__module__", "tvm_ffi")
+    except AttributeError:
+        pass
```
junrushao: @tqchen I will need to normalize `__module__` of objects like `core.Object` to `tvm_ffi.Object` so that the "as method XXX" part can find the method properly (see screenshot). Let me know if this makes sense!

Image

python/tvm_ffi/__init__.py
```diff
@@ -95,3 +95,13 @@
...
+
+
+def _update_module() -> None:
+    for name in __all__:
+        obj = globals()[name]
+        if not getattr(obj, "__module__", "tvm_ffi").startswith("tvm_ffi"):
+            obj.__module__ = "tvm_ffi"
+
+
+_update_module()
```
junrushao, line 111: @tqchen I will need to normalize `__module__` of objects like `core.Object` to `tvm_ffi.Object` so that the "as method XXX" part can find the method properly (see screenshot). Let me know if it's a reasonable change!

Image

tqchen: that sounds good

PR #178, opened 2025-10-20
Introduce dtype protocol

tqchen:
This PR adds support for `__dlpack_data_type__` to enable exchange of dtype from arguments. Note the intention is to use DLDataType to ingest dtype into tvm ffi since it is the most close to metal repr.

Likely we also can look into numpy dtype protocol once that materializes.

### DIFF
--- a/python/tvm_ffi/_dtype.py
+++ b/python/tvm_ffi/_dtype.py
@@ -60,15 +60,39 @@ class dtype(str):
 
     """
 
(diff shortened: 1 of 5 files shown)


tqchen:
also changes the python private field of dtype (to be consistent that dunder is always a function), given it is private, we consider it is a compatible change

junrushao (approved):

PR #187, opened 2025-10-22
feat: Enhance load_module to accept PathLike objects

### LINKED ISSUE #158: [Feature Request] Support path-like objects for `tvm_ffi.load_module`
As title.

Most of python's builtin functions (e.g. [`open`](https://docs.python.org/3.8/library/functions.html#open), [`os.path`](https://docs.python.org/3.8/library/os.path.html#module-os.path)) take [path-like objects](https://docs.python.org/3/glossary.html#term-path-like-object), not only path string, as a path. Supporting this can reduce one conversion when searching targeted lib with `pathlib`.

Tunghohin:
Closes #158

### DIFF
--- a/python/tvm_ffi/module.py
+++ b/python/tvm_ffi/module.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 from collections.abc import Sequence
 from enum import IntEnum
+from os import PathLike, fspath
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 from . import _ffi_api, core
@@ -305,12 +306,12 @@ def system_lib(symbol_prefix: str = "") -> Module:
     return _ffi_api.SystemLib(symbol_prefix)
 
 
-def load_module(path: str) -> Module:
+def load_module(path: str | PathLike) -> Module:
     """Load module from file.
 
     Parameters
     ----------
-    path
+    path : str | PathLike
         The path to the module file.
 
     Returns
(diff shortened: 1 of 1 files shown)


oraluben:
You can use following format in your description to link your PR with the issue :)
Closes: #xxx 

Ref: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

python/tvm_ffi/module.py
```diff
@@ -305,29 +307,36 @@ def system_lib(symbol_prefix: str = "") -> Module:
     return _ffi_api.SystemLib(symbol_prefix)
 
 
-def load_module(path: str) -> Module:
+def load_module(path: str | PathLike) -> Module:
     """Load module from file.
-
+
```
oraluben: I suggest to add [`'W'`](https://docs.astral.sh/ruff/rules/#warning-w) rule to ruff so we can check for these extra spaces cc @junrushao

junrushao: I thought we already did with pre-commit hooks’s trailing white space checking? @oraluben if not, happy to accept a PR doing that!

python/tvm_ffi/module.py
```diff
@@ -305,12 +306,12 @@ def system_lib(symbol_prefix: str = "") -> Module:
...
 
 
-def load_module(path: str) -> Module:
+def load_module(path: str | PathLike) -> Module:
     """Load module from file.
 
     Parameters
     ----------
-    path
+    path : str or PathLike
```
oraluben: I think it's better to have a valid type expression for python here, e.g. `str | PathLike` like the real signature above, unless there's other format in docstring?

python/tvm_ffi/module.py
```diff
@@ -322,12 +323,20 @@ def load_module(path: str) -> Module:
...
+        mod = tvm_ffi.load_module(Path("path/to/module.so"))
+        mod.func_name(*args)
 
     See Also
     --------
     :py:class:`tvm_ffi.Module`
 
     """
-    return _ffi_api.ModuleLoadFromFile(path)
+    # Convert PathLike objects to string
```
oraluben: It should be sufficient to just use `path = fspath(path)` as in python's stdlib:

https://github.com/python/cpython/blob/bd2c7e8c8b10f4d31eab971781de13844bcd07fe/Lib/posixpath.py#L62-L64

PR #189, opened 2025-10-24
feat: Build PyTorch DLPack extension ahead of time

yzh119:
This PR adds option to build PyTorch DLPack extension ahead of time.

# Changes

- Environment variable control: Set `TVM_FFI_BUILD_TORCH_EXT=1` to pre-compile the extension at build time
- Local version tagging: Set `TVM_FFI_LOCAL_VERSION=torch2.9.0` to create version-tagged wheels (e.g., `apache_tvm_ffi-0.1.0+torch2.9.0-...whl`)
- Dynamic build dependencies: PyTorch is only required as a build dependency when `TVM_FFI_BUILD_TORCH_EXT=1`
- Automatic fallback: Runtime automatically uses pre-compiled extension if available, otherwise falls back to JIT compilation

# Usage

## Default: JIT compilation (no torch build dependency)
```bash
python -m build --wheel
```

## AOT compilation with version tag
```bash
TVM_FFI_BUILD_TORCH_EXT=1 TVM_FFI_LOCAL_VERSION="torch2.5.0" python -m build --wheel
```

Backward compatible: Default behavior unchanged, opt-in only.

Remaining work includes adding the release workflow and @cyx-6 could take over, you can refer to https://github.com/flashinfer-ai/flashinfer/blob/main/.github/workflows/release.yml

### DIFF
--- a/addons/torch_c_dlpack_ext/build_backend.py
+++ b/addons/torch_c_dlpack_ext/build_backend.py
@@ -52,9 +52,8 @@ def get_requires_for_build_wheel(
     """Get build requirements for wheel, conditionally including torch and apache-tvm-ffi."""
     requires = orig.get_requires_for_build_wheel(config_settings)
     if not _is_lib_prebuilt():
-        # build wheel from sdist package, requires torch and apache-tvm-ffi
-        requires.append("torch")
-        requires.append("apache-tvm-ffi")
+        # build wheel from sdist package, requires apache-tvm-ffi>=0.1.1 to build lib
+        requires.append("apache-tvm-ffi>=0.1.1")
     return requires
 
 
@@ -68,15 +67,21 @@ def build_wheel(
         # build wheel from sdist package, compile the torch c dlpack ext library locally.
         import torch  # noqa: PLC0415
 
-        subprocess.run(
-            [
-                sys.executable,
-                "-m",
-                "tvm_ffi.utils._build_optional_torch_c_dlpack",
-                "--output-dir",
-                str(_package_path),
-                "--build-with-cuda" if torch.cuda.is_available() else "",
-            ],
-            check=True,
-        )
+        if hasattr(torch.Tensor, "__c_dlpack_exchange_api__"):
+            print(
+                "torch.Tensor already has attribute __c_dlpack_exchange_api__. "
+                "No need to build any torch c dlpackc libs."
+            )
+        else:
+            subprocess.run(
+                [
+                    sys.executable,
+                    "-m",
+                    "tvm_ffi.utils._build_optional_torch_c_dlpack",
+                    "--output-dir",
+                    str(_package_path),
+                    "--build-with-cuda" if torch.cuda.is_available() else "",
+                ],
+                check=True,
+            )
     return orig.build_wheel(wheel_directory, config_settings, metadata_directory)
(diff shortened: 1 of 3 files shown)


tqchen:
some notes:
- We can split the wheel into root/addons/torch_c_dlpack_ext
- No need to duplicate the c_dlpack_ext.cc, instead, we can update invoke tvm_ffi._optional_c_dlpack.get_source()
    - just need to split out the ffi specific part later

yzh119 (commented):
Comments:
1. we need both x86_64 and aarch64 wheels
2. once this package is ready, consider make it (at least optional) dependency of tvm-ffi
3. try https://github.com/jlumbroso/free-disk-space to clean up the disk space of github runners.
4. using pyproject.toml instead of setup.py
5. release wheel instead of sdist
6. <del> using torch library and Py_LIMITED_API (if possible) </del>

cc @cyx-6 @yaoyaoding @tqchen for viz.

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -54,6 +54,15 @@ def load_torch_c_dlpack_extension() -> Any:
         return None
 
     """Load the torch c dlpack extension."""
+    try:
+        import torch_c_dlpack_ext  # type: ignore  # noqa: PLC0415
+
+        func = torch_c_dlpack_ext.load_torch_c_dlpack_lib(torch.__version__)
+        setattr(torch.Tensor, "__c_dlpack_exchange_api__", func())
+        return None
+    except ImportError:
```
tqchen, line 61: note that this may still raise a ValueError if in the case of load_torch_c_dlpack_lib failed, we should support such case as well, to relax assumption here

addons/torch_c_dlpack_ext/setup.py
```diff
@@ -0,0 +1,28 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""setup script."""
+
+from setuptools import find_packages, setup
+
+setup(
+    name="torch_c_dlpack_ext",
```
tqchen: would still be useful to have a pyproject.toml, we should have build dep on apache-tvm-ffi. 

On setup, it would be useful to detect if the build so is already available. If so, we proceed with what we have. 

If not, we perform a sdist build, by calling into tvm_ffi's build so and place it onto the right place. This can be helpful for users installing sdist

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -54,6 +54,15 @@ def load_torch_c_dlpack_extension() -> Any:
         return None
 
     """Load the torch c dlpack extension."""
+    try:
+        import torch_c_dlpack_ext  # type: ignore  # noqa: PLC0415
+
+        func = torch_c_dlpack_ext.load_torch_c_dlpack_lib(torch.__version__)
```
tqchen: do the loading and patching in `torch_c_dlpack_ext`, so we just try `import torch_c_dlpack_ext` and catch exception.

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -54,6 +54,15 @@ def load_torch_c_dlpack_extension() -> Any:
...
+        setattr(torch.Tensor, "__c_dlpack_exchange_api__", func())
```
tqchen: do the setattr in `torch_c_dlpack_ext`

PR #191, opened 2025-10-26
doc: Stable C ABI

junrushao:


examples/stable_c_abi/src/add_one_cpu.c
```diff
@@ -0,0 +1,56 @@
...
+// [example.begin]
+// File: src/add_one_cpu.cc
+TVM_FFI_DLL int __tvm_ffi_add_one_cpu(void* handle, const TVMFFIAny* args, int32_t num_args,
+                                      TVMFFIAny* result) {
+  // Step 1. Extract inputs from `Any`
+  // Step 1.1. Extract `x := args[0]`
+  DLTensor* x;
+  if (args[0].type_index == kTVMFFIDLTensorPtr) x = (DLTensor*)(args[0].v_ptr);
+  else if (args[0].type_index == kTVMFFITensor) x = (DLTensor*)(args[0].v_c_str + sizeof(TVMFFIObject));
+  else { TVMFFIErrorSetRaisedFromCStr("ValueError", "Expects a Tensor input"); return -1; }
```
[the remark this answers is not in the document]
junrushao: This is actually fine

examples/stable_c_abi/src/add_one_cpu.c
```diff
@@ -0,0 +1,56 @@
...
+  // Step 1.2. Extract `y := args[1]`
+  DLTensor* y;
+  if (args[1].type_index == kTVMFFIDLTensorPtr) y = (DLTensor*)(args[1].v_ptr);
+  else if (args[1].type_index == kTVMFFITensor) y = (DLTensor*)(args[1].v_c_str + sizeof(TVMFFIObject));
+  else { TVMFFIErrorSetRaisedFromCStr("ValueError", "Expects a Tensor output"); return -1; }
```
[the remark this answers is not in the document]
junrushao: This is actually fine

examples/stable_c_abi/src/load.c
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+int main() {
+  Initialize();
+  int ret_code = 0;
+  float x_data[5] = {1.0, 2.0, 3.0, 4.0, 5.0};
+  float y_data[5] = {0.0, 0.0, 0.0, 0.0, 0.0};
+  int64_t shape[1] = {5};
+  int64_t strides[1] = {0};
```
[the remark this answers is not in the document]
junrushao: Good catch!

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,190 @@
...
+The ``universal_c_abi`` function uses a stable layout for all the input and output arguments.
+
+Layout of ``Any``
+~~~~~~~~~~~~~~~~~
+
+TVM-FFI's ``Any`` is a fixed size (128-bit) tagged union that represents all supported types.
+
+- First 32 bits: represent the type index, indicating which type the value holds. By definition, it represents at most 2^32 different types.
+- Next 32 bits: null padding, or reserved as special flags in rare cases.
+- Last 64 bits: represent the value itself, which can be an 64-bit integer, floating pointer number, or pointer to a heap allocated object.
```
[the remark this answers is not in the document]
junrushao: Good catch!

examples/stable_c_abi/src/load.c
```diff
@@ -0,0 +1,118 @@
...
+  call_args[1] = (TVMFFIAny){.type_index = kTVMFFIRawStr, .v_c_str = "add_one_cpu"};
+  call_args[2] = (TVMFFIAny){.type_index = kTVMFFIBool, .v_int64 = 0};
+  if ((ret_code = TVMFFIFunctionCall(fn_get_function, call_args, 3, &func))) goto _RAII;
+
+  // Step 3. Call function `add_one_cpu(x, y)`
+  // Equivalent to:
+  //    func(x, y)
+  call_args[0] = (TVMFFIAny){.type_index = kTVMFFIDLTensorPtr, .v_ptr = x};
+  call_args[1] = (TVMFFIAny){.type_index = kTVMFFIDLTensorPtr, .v_ptr = y};
+  if ((ret_code = TVMFFIFunctionCall(func.v_ptr, call_args, 2, &((TVMFFIAny){})))) goto _RAII;
```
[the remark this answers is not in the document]
junrushao: ok it's a good idea!

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+.. admonition:: Key Idea
+  :class: important
+
+  Every function call can be represented by a single ABI stable C function:
+
+  .. code-block:: c
+
+      int universal_c_abi(  // returns 0 if succeed, error code if failure
+        void*  handle,      // library handle
+        Any*   args,        // inputs: args[0 ... N - 1]
```
tqchen: const TVMFFIAny*

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+        int    N,           // number of inputs
+        Any*   result,      // output: *result
```
tqchen: TVMFFIAny*

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+      );
+
+  where ``Any`` is a tagged union of all supported types, e.g. integers, floats, Tensors, strings, etc., and can be further extended to arbitrary user-defined types.
```
tqchen: TVMFFIAny

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+.. software distributed under the License is distributed on an
+.. "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+
+Stable C ABI
+============
+
+TVM-FFI centers on a single key idea:
```
tqchen: TVM FFI defines a common stable C ABI protocol for all functions

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+
+
+.. admonition:: Key Idea
```
tqchen: TVM FFI Function C ABI

docs/get_started/stable_c_abi.rst
tqchen: It is better to be consistent with the C ABI code, the simplification is less useful and could cause confusion with c++ ffi::Any(note that they re managed and unmanged version)

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+  :class: important
+
+  Every function call can be represented by a single ABI stable C function:
+
+  .. code-block:: c
+
+      int universal_c_abi(  // returns 0 if succeed, error code if failure
```
tqchen: tvm_ffi_func_c_abi

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+
+- The stable C layout and calling convention of ``universal_c_abi``;
+- C examples that implements the stable C ABI, and calls functions that follow this ABI.
+
+Stable C Layout
+---------------
+
+The ``universal_c_abi`` function uses a stable layout for all the input and output arguments.
+
+Layout of ``Any``
```
tqchen: TVMFFIAny

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+~~~~~~~~~~~~~~~~~
+
+TVM-FFI's ``Any`` is a fixed size (128-bit) tagged union that represents all supported types.
+
+- First 32 bits: type index indicating which value is stored (supports up to 2^32 types).
+- Next 32 bits: null padding, or reserved for special flags in rare cases.
```
tqchen: reserved used for small string length and padded to zero in other cases

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+The following conventions apply when representing values in ``Any``:
+
+- Primitive types: the last 64 bits directly store the value, for example:
+
+  * Integers
+  * Floating-point numbers
+
+- Heap-allocated objects: the last 64 bits store a pointer to the actual object, for example:
+
+  * `DLPack tensors <https://data-apis.org/array-api/2024.12/design_topics/data_interchange.html#dlpack-an-in-memory-tensor-structure>`_
```
tqchen: Managed Tensor objects that follows DLPack layout. "DLPack tensors can be confusing since ppl may mistake it with DLManagedTensor"

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+  * C-style strings
```
tqchen: Other objects( C-style string can cause some confusion), since string can either be small string(on-stack) or on heap, it is ok to remove this

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+  * :py:class:`tvm_ffi.Function`, representing all functions, such as Python/C++ functions/lambdas, etc.;
+  * :py:class:`tvm_ffi.Array` and :py:class:`tvm_ffi.Map` as list/dictionary containers of all ``Any`` values;
+  * Up to 2^32 types can be represented in this type system.
+
+Function Calling Convention
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Function calls in TVM-FFI share the same calling convention, ``universal_c_abi``, as described above.
+
+- ``handle: void*``: a library handle that contains the function to be called. It is usually ``nullptr`` for most use cases, but some libraries (such as cuDNN) may use it to store global contexts.
```
tqchen: a library handle that contains the function to be called. It is usually `nullptr` for exported symbols. It can be used to enable closures that captures extra data.

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+Stable ABI in C Code
+--------------------
+
+.. note::
+
+  All the code in this tutorial can be found under `examples/stable_c_abi <https://github.com/apache/tvm-ffi/tree/main/examples/stable_c_abi>`_ in the repository, including commands to compile the code and run the examples.
+
+This section shows how to write C code that follows the stable C ABI. Specifically, we provide two examples:
+
+- A CPU ``add_one_cpu`` kernel in C that is equivalent to the :ref:`C++ example <cpp_add_one_kernel>`.
```
tqchen: change name to `add_one_c` or `add_one_c_abi`

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+
+.. figure:: ../_static/tvm-ffi-layout-func.svg
+   :alt: Calling convention for universal_c_abi
+   :name: fig:layout-func
+
+   Figure 2. Layout spec of ``universal_c_abi`` function.
+
+
+Stability and Interoperability
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tqchen, line 123: likely this section is redundant with readme and can be removed

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+**Stability.** The pure C layout and the calling convention are stable across compiler versions and independent of host languages or frameworks.
+
+**Cross-language.** TVM-FFI implements this calling convention in multiple languages (C, C++, Python, Rust, ...), enabling code written in one language—or generated by a DSL targeting the ABI—to be called from another language.
+
+**Cross-framework.** TVM-FFI uses standard data structures such as `DLPack tensors <https://data-apis.org/array-api/2024.12/design_topics/data_interchange.html#dlpack-an-in-memory-tensor-structure>`_ to represent arrays, so compiled functions can be used from any array framework that implements the DLPack protocol (NumPy, PyTorch, TensorFlow, CuPy, JAX, and others).
+
+
+Stable ABI in C Code
+--------------------
+
```
tqchen, line 134: We design the C ABI with DSL and ML compilers in mind. In the context of ML compilers, usually we do not have access to c++ templates and needs to operate directly in LLVM (MLIR dialect) or low-level C. In such cases, it is helpful to directly leverage and target the stable C ABI.

This section shows how we can expose a kernel directly using the C ABI. Because we write code directly in C without any boilerplate, the code can serve as a direct reference for DSL compilers to expose kernel functions following our C ABI standard.

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+.. literalinclude:: ../../examples/stable_c_abi/src/add_one_cpu.c
+   :language: c
+   :start-after: [example.begin]
+   :end-before: [example.end]
+
+
+Compared to the :ref:`C++ example <cpp_add_one_kernel>`, there are a few key differences:
+
+- The explicit marshalling in **Step 1** is only needed in C. In C++, templates hide these details.
```
tqchen, line 190: it is better to directly talk about this example, target audience here is compiler developers, we can add c++ template reference in end.

At high-level, we perform the following steps:

- Function symbols follow name int __tvm_ffi_<name>
- We go over and check the type_index in  TVMFFIAny and extract the input function arguments
- The function Return 0 for success, -1 for error 
    - When error happens, we set the error via `TVMFFIErrorSetRaisedFromCStr` or `TVMFFIErrorSetRaisedFromCStrParts`

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+
+  In TVM-FFI's C++ APIs, many invocables (functions, lambdas, functors) are automatically converted into the universal C ABI form by :cpp:class:`tvm::ffi::Function` and :cpp:class:`tvm::ffi::TypedFunction`.
+
+  Rule of thumb: if an invocable's arguments and result can be converted to/from :cpp:class:`tvm::ffi::Any`, it can be wrapped as a universal C ABI function.
+
+
+Load and Run the Kernel
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Next, a minimal C loader invokes the ``add_one_cpu`` kernel. It is functionally identical to the :ref:`C++ example <cpp_load>` and performs:
```
tqchen, line 203: C loader is less useeful, instead we can use a python loader. Since the main goal is to demonstrate we can use the same C ABI to expose functions that can be universally accessed.

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,189 @@
...
+- Optionally convert the output ``Any`` back to the desired type, if the function returns a value.
+
+What's Next
+-----------
+
+**ABI specification.** See the complete ABI specification in :doc:`../concepts/abi_overview`.
+
+**Convenient compiler target.** The stable C ABI is a simple, portable codegen target for DSL compilers. Emit C that follows this ABI to integrate with TVM-FFI and call the result from multiple languages and frameworks. See :doc:`../guides/compiler_integration`.
+
+**Rich and extensible type system.** TVM-FFI supports a rich set of types in the stable C ABI: primitive types (integers, floats), DLPack tensors, strings, built-in reference-counted objects (functions, arrays, maps), and user-defined reference-counted objects. See :doc:`../guides/cpp_guide`.
```
tqchen, line 254: Likely the two we should recommend reads are abi specification and compiler_integration

docs/get_started/stable_c_abi.rst
junrushao: On the other hand, TVMFFIAny is an over-complicated name, and it’s more about implementation details rather than key idea.

How about this? Let’s use “Any” in the “Key Idea” section, and closely note that it means :cpp:class:`TVMFFIAny`

docs/get_started/stable_c_abi.rst
junrushao: I don’t think it would be helpful for understanding if we enumerate all the details

docs/get_started/stable_c_abi.rst
junrushao: I worry less about absolute technical precision, comparing to how much time it takes impatient readers to get the high-level idea right.

More specifically in our case, as a “get started” tutorial, I’d prefer the wording starts gentle and zooming in gradually into details.

In this case, introducing a very high-level key idea (with less noise around impl details) seems more reasonable approach overall to me.

Therefore, I don’t like the idea of “common stable C ABI protocol for all” which is less comprehensible than “ok this is the key idea”

docs/get_started/stable_c_abi.rst
junrushao: How about this? 

Key idea: C ABI for TVM-FFI Functions

docs/get_started/stable_c_abi.rst
tqchen: I think in this case being technically precise is more important, and it is not that more complicated. The main reason is main target audience is DSL compiler writer, knowing exactly what it is is better

junrushao: ok i will use `TVMFFIAny` everywhere then

docs/get_started/stable_c_abi.rst
junrushao: changed it to

> reserved for special flags in rare cases (e.g. small string optimization)

docs/get_started/stable_c_abi.rst
junrushao: removed

docs/get_started/stable_c_abi.rst
junrushao: changed to this:

> ``handle: void*``: a library handle that contains the function to be called. It is usually ``nullptr`` for most use cases such as exported symbols. It can be used to enable closures or module-scoped functions that capture extra context.

junrushao: indeed this cuDNN library handle is pretty misleading and wrong

docs/get_started/stable_c_abi.rst
junrushao: it's fine to re-emphasize

docs/get_started/stable_c_abi.rst
junrushao: we did in the two bullet points. some additional reading on "type index extensibility" seems fine for curious readers, so i put the 3rd bullet point here just in case

docs/get_started/stable_c_abi.rst
junrushao: this section is designed as showing two sides of the ABI:
- callee side: using `add_one` kernel
- caller side: using `loader` code

the caller side example can be useful if we want to codegen graph IR calling kernel IRs

docs/get_started/stable_c_abi.rst
junrushao: adjusted a little bit:

> TVM FFI's :ref:`C ABI <tvm_ffi_c_abi>` is designed with DSL and ML compilers in mind. DSL codegen usually relies on MLIR, LLVM or low-level C as the compilation target, where no access to C++ features is available, and where stable C ABIs are preferred for simplicity and stability.
>
> This section shows how to write C code that follows the stable C ABI. Specifically, we provide two examples:
>
> - Callee side: A CPU ``add_one_cpu`` kernel in C that is equivalent to the :ref:`C++ example <cpp_add_one_kernel>`.
> - Caller side: A loader and runner in C that invokes the kernel, a direct C translation of the :ref:`C++ example <cpp_load>`.
>
> The C code is minimal without any other dependencies, and thus can serve as a direct reference for DSL compilers to expose or invoke kernels or other functions following our C ABI standard.

docs/get_started/stable_c_abi.rst
junrushao: i reworded it a little bit:

> TVM-FFI centers on a single key idea.
>
> ...code block
>
> Built on top of this stable C ABI, TVM-FFI defines a common C ABI protocol for all functions, and further provides an extensible, performant, and ecosystem-friendly open solution for all.

docs/get_started/stable_c_abi.rst
junrushao: i think naming wise it's fine. furthermore, i'd love to have a drop-in replacement of the cpp counterpart

docs/get_started/stable_c_abi.rst
junrushao: good point!

junrushao: adjusted a little bit:

> **Compiler codegen**. This C code serves as a direct reference for DSL compilers. To emit a function that follows the stable C ABI, the compiler needs to ensure the following aspects:
> 
> - Symbol naming: define the exported symbol name as ``__tvm_ffi_{func_name}``;
> - Type checking: check types of input arguments via :cpp:member:`TVMFFIAny::type_index`, and then marshal inputs from :cpp:class:`TVMFFIAny` to the desired types;
> - Exception handling: Return 0 on success, or a non-zero error code on failure. When error occurs, set appropriate error message via :cpp:func:`TVMFFIErrorSetRaisedFromCStr` or :cpp:func:`TVMFFIErrorSetRaisedFromCStrParts`.

docs/get_started/stable_c_abi.rst
```diff
@@ -0,0 +1,254 @@
...
+
+     .. code-block:: bash
+
+        pip install --reinstall --upgrade apache-tvm-ffi
+
+This guide introduces TVM-FFI's stable C ABI: a single, minimal and stable
+ABI that represents any cross-language calls, with DSL and ML compiler codegen
+in mind.
+
+TVM-FFI centers on a single key idea below:
```
tqchen: TVM-FFI builds on the following key idea

junrushao: updated!

PR #192, opened 2025-10-26
[Feature] Add a function to build c-dlpack for torch

yaoyaoding:
This PR adds a script `python/tvm_ffi/utils/_build_optional_c_dlpack.py`, which a standalone python module that is used to build a shared library as a patch to make old version of pytorch to support the dlpack protocol.

Two kinds of usage:

1. used as a standalone python module to build the shared library for each combination of `<python-version>` and `<pytorch-version>` and build a package to contain the built libraries.
2. used as a jit module in tvm-ffi when there is not a prebuilt patch found.

## AOT Usage

```bash
python python/tvm_ffi/utils/_build_optional_c_dlpack.py --build_dir <build-dir> [--build_with_cuda]
```

Then there will be a shared library under the given `<build-dir>` (either `libtorch_c_dlpack_addon.so` or `libtorch_c_dlpack_addon.dll`, depends on the platform). The built shared library is specific to the current
python version and pytorch version.

## JIT Usage

We launched a python interpreter to run the build script. It will store the built shared library in tvm-ffi cache. See `python/tvm-ffi/_optional_torch_c_dlpack.py`.


yzh119:
Do we have an ETA about when would this feature get ready? @yaoyaoding @tqchen

yaoyaoding:
@yzh119 this afternoon.

yaoyaoding:
Hi @yzh119, the PR is ready for review.

PR #203, opened 2025-10-29
[Lint] Lint the build torch c dlpack utility

yaoyaoding:
This PR lint the `_optional_build_torch_c_dlpack.py` utility by moving some definition to the inside of the main function.

Usage:
```
$ python -m tvm_ffi.utils._build_optional_torch_c_dlpack -h
usage: _build_optional_torch_c_dlpack.py [-h] [--build-dir BUILD_DIR] [--output-dir OUTPUT_DIR] [--build-with-cuda] [--libname LIBNAME]

Build the torch c dlpack extension. After building, a shared library will be placed in the output directory.

options:
  -h, --help            show this help message and exit
  --build-dir BUILD_DIR
                        Directory to store the built extension library. If not provided, a temporary directory will be used.
  --output-dir OUTPUT_DIR
                        Directory to store the built extension library. If not specified, the default cache directory of tvm-ffi will be used.
  --build-with-cuda     Build with CUDA support.
  --libname LIBNAME     The name of the generated library. It can be a name 'auto' to auto-generate a name following 'libtorch_c_dlpack_addon_torch{version.major}{version.minor}-cpu/cuda.{extension}'.
```

### DIFF
--- a/python/tvm_ffi/_optional_torch_c_dlpack.py
+++ b/python/tvm_ffi/_optional_torch_c_dlpack.py
@@ -60,17 +60,26 @@ def load_torch_c_dlpack_extension() -> Any:
 
         # check whether a JIT shared library is built in cache
         cache_dir = Path(os.environ.get("TVM_FFI_CACHE_DIR", "~/.cache/tvm-ffi")).expanduser()
-        addon_build_dir = cache_dir / "torch_c_dlpack_addon"
-        lib_path = addon_build_dir / (
-            "libtorch_c_dlpack_addon" + (".dll" if sys.platform == "win32" else ".so")
-        )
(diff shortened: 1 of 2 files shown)


python/tvm_ffi/utils/_build_optional_torch_c_dlpack.py
```diff
@@ -30,13 +30,6 @@
 import torch.torch_version
 import torch.utils.cpp_extension
```
tqchen, line 32: also document at the top. # Important: to avoid cyclic dependency, we avoid import tvm_ffi names at top level

yaoyaoding: Sure!

PR #206, opened 2025-10-29
doc(python): Add example coverage to docstrings

junrushao:
This PR refines Python-side documentation. It includes:

- Adding example code to every public API as much as possible
- Making existing examples copy-pastable as much as possible
- Polish writing
- Clarifies a few API usage, e.g. system_lib

### DIFF
--- a/python/tvm_ffi/module.py
+++ b/python/tvm_ffi/module.py
@@ -115,13 +115,14 @@ class Module(core.Object):
     def implements_function(self, name: str, query_imports: bool = False) -> bool:
         """Return True if the module defines a global function.
 
-        Note
-        ----
-        that has_function(name) does not imply get_function(name) is non-null since the module
-        that has_function(name) does not imply get_function(name) is non-null since the module
-        may be, eg, a CSourceModule which cannot supply a packed-func implementation of the function
-        without further compilation. However, get_function(name) non null should always imply
-        has_function(name).
+        Notes
+        -----
+        ``implements_function(name)`` does not guarantee that
+        ``get_function(name)`` will return a callable, because some module
+        kinds (e.g. a source-only module) may not provide a packed function
+        implementation until further compilation occurs. However, a non-null
+        result from ``get_function(name)`` should imply the module implements
+        the function.
 
         Parameters
         ----------
@@ -133,8 +134,7 @@ class Module(core.Object):
 
         Returns
         -------
-        b
-            True if module (or one of its imports) has a definition for name.
+        True if module (or one of its imports) has a definition for name.
 
         """
         return _ffi_api.ModuleImplementsFunction(self, name, query_imports)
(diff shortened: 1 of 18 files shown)


python/tvm_ffi/utils/lockfile.py
```diff
@@ -35,6 +35,19 @@ class FileLock:
 
     This class implements an advisory lock, which must be respected by all
     cooperating processes.
+
```
junrushao, line 38: @yaoyaoding could you enlighten me here - why aren't we using the standard [filelock](https://py-filelock.readthedocs.io/en/latest/) package instead?

src/ffi/extra/library_module_system_lib.cc
```diff
@@ -134,6 +142,8 @@ TVM_FFI_STATIC_INIT_BLOCK() {
     }
     *rv = SystemLibModuleRegistry::Global()->GetOrCreateModule(symbol_prefix);
   });
+  TVMFFIEnvModRegisterSystemLibSymbol("__tvm_ffi_test_symbol_add_one",
```
junrushao: @tqchen I added this symbol for documentation purpose

python/tvm_ffi/module.py
```diff
@@ -133,8 +134,7 @@ def implements_function(self, name: str, query_imports: bool = False) -> bool:
 
         Returns
         -------
-        b
-            True if module (or one of its imports) has a definition for name.
+        True if module (or one of its imports) has a definition for name.
```
junrushao, line 137: there's no need to name an extra variable - the return type is derived from function signature

python/tvm_ffi/module.py
```diff
@@ -311,25 +332,25 @@ def load_module(path: str | PathLike) -> Module:
 
     Parameters
     ----------
-    path : str | PathLike
+    path
```
junrushao, line 335: there's no need either to annotation type again - it's auto-derived from function signature

src/ffi/extra/library_module_system_lib.cc
```diff
@@ -125,6 +125,14 @@ class SystemLibModuleRegistry {
   Map<String, ffi::Module> lib_map_;
 };
 
+static int test_symbol_add_one(void*, const TVMFFIAny* args, int32_t num_args, TVMFFIAny* ret) {
+  TVM_FFI_SAFE_CALL_BEGIN();
```
tqchen: let us avoid this in the lib, maybe to testing? or in this case better to skip

junrushao: moved to `testing.cc` per discussion

junrushao:
@tqchen can you re-review?

python/tvm_ffi/utils/lockfile.py
yaoyaoding: The only reason is to reduce the 3rdparty dependency (cc @tqchen). If the community think the filelock package is ok to have, there is no blocker for us to switch to it.

junrushao: i see. the question is simply out of curiosity and im fine either way. happy to keep it as it is now as long as it doesn't cause any problem

tqchen: I think dependency reduction is good

Issue #184, opened 2025-10-21
[Feature Request] Better load/build inline


DarkSharpness:
Great work! I have a small suggestion. Currently, `tvm_ffi.cpp.build_inline` only accepts raw cpp/CUDA source code strings, rather than file paths. It would be useful to add a `tvm_ffi.cpp.load` function, similar to `torch.utils.cpp_extension.load`, which can directly compile from source files.

junrushao:
Seems like a good first issue! @DarkSharpness feel free to contribute if you have bandwidth, and I'm happy to help if you got any issues/questions.

Basically we need to extend this file: `python/tvm_ffi/cpp/load_inline.py`, and support Path-like objects.

yzh119:
FYI, in flashinfer the JitSpec basically implement this functionality on top of tvm-ffi: https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/jit/core.py#L180-L280 (with functions such as `build` (for AOT) and `build_and_load` (for JIT)).

DarkSharpness:
I was busy with my paper last week. If no one is working on this for now, I can handle this :)

PR #213, opened 2025-10-31
[Feature] Support build and build_and_load.

DarkSharpness:
Related issue #184 . This PR aims at improving the functionality of `build_inline`. We implement a new `build`/`build_and_load` function which accepts paths to cpp/cuda files instead of raw string of cpp/cuda files. They have an interface similar to [`torch.utils.cpp_extension.load`](https://docs.pytorch.org/docs/2.9/cpp_extension.html#torch.utils.cpp_extension.load).

### DIFF
--- a/tests/python/test_build.cc
+++ b/tests/python/test_build.cc
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "test_build.h"
+
+#include <tvm/ffi/container/tensor.h>
+#include <tvm/ffi/dtype.h>
+#include <tvm/ffi/error.h>
+#include <tvm/ffi/extra/c_env_api.h>
+#include <tvm/ffi/function.h>
+
+void add_one_cpu(tvm::ffi::TensorView x, tvm::ffi::TensorView y) {
+  TVM_FFI_ICHECK(x.ndim() == 1) << "x must be a 1D tensor";
+  DLDataType f32_dtype{kDLFloat, 32, 1};
+  TVM_FFI_ICHECK(x.dtype() == f32_dtype) << "x must be a float tensor";
+  TVM_FFI_ICHECK(y.ndim() == 1) << "y must be a 1D tensor";
+  TVM_FFI_ICHECK(y.dtype() == f32_dtype) << "y must be a float tensor";
+  TVM_FFI_ICHECK(x.size(0) == y.size(0)) << "x and y must have the same shape";
+  for (int i = 0; i < x.size(0); ++i) {
+    static_cast<float*>(y.data_ptr())[i] = static_cast<float*>(x.data_ptr())[i] + 1;
+  }
+}
+
+TVM_FFI_DLL_EXPORT_TYPED_FUNC(add_one_cpu, add_one_cpu);
--- a/tests/python/test_build.h
+++ b/tests/python/test_build.h
@@ -1,0 +1,25 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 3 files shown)


tqchen:
@yaoyaoding can you help review?

python/tvm_ffi/cpp/extension.py
```diff
@@ -702,3 +760,94 @@ def load_inline(
...
+    *,
+    cpp_files: Sequence[str] | str | None = None,
+    cuda_files: Sequence[str] | str | None = None,
+    extra_cflags: Sequence[str] | None = None,
+    extra_cuda_cflags: Sequence[str] | None = None,
+    extra_ldflags: Sequence[str] | None = None,
+    extra_include_paths: Sequence[str] | None = None,
+    build_directory: str | None = None,
+) -> str:
+    """Compile and build a C++/CUDA module from source files."""
```
yaoyaoding: Add more detailed documentation for this function, and `build_and_load` function. 
Then add these two new functions to https://github.com/apache/tvm-ffi/blob/main/docs/reference/python/index.rst?plain=1#L105-L114 section. 
It's also good to rename the section from "inline loading" to something like "cpp extension" or "cpp utility".

python/tvm_ffi/cpp/extension.py
```diff
@@ -360,7 +417,16 @@ def _decorate_with_tvm_ffi(source: str, functions: Mapping[str, str]) -> str:
...
+def _str_seq2list(seq: Sequence[str] | str | None) -> list[str]:
+    if seq is None:
+        return []
+    elif isinstance(seq, str):
+        return [seq]
+    else:
+        return list(seq)
+
+
+def build_inline(
```
yaoyaoding, line 452: with the new `build` function, this function can be refactored to
1. decorate the sources given functions to export
2. write the sources to build directory
3. use `build` function to build the shard library

python/tvm_ffi/cpp/load_inline.py
```diff
@@ -195,6 +206,52 @@ def _run_command_in_dev_prompt(
...
+            )
+        )
+        link_files.append("main.o")
+        if with_cuda:
+            ninja.append(
+                "build cuda.o: compile_cuda {}".format(
+                    str((Path(build_dir) / "cuda.cu").resolve()).replace(":", "$:")
+                )
+            )
+            link_files.append("cuda.o")
```
yaoyaoding: after refactor `build_inline`, this logic is not needed.

junrushao:
could you add this method to docs? it's under `./docs/reference/python/index.rst`

DarkSharpness:
Updated cc @yaoyaoding @junrushao.

I'm not sure about the backward compatibility. Without [this commit](https://github.com/apache/tvm-ffi/pull/213/commits/aabfb6892dd42a867f19a3545a4f985e9017dbf2), old code compiled via `load_inline` or `build_inline` will trigger a recompile, because the intermediate object name changed.

python/tvm_ffi/cpp/extension.py
DarkSharpness: I'm not sure how to implement this. The old `build_inline` placed the C++/CUDA source files directly in the build directory.  If we want to reuse the existing `build` function, we’d need to write the source code to disk before we even know the final build directory.

My suggestion would be to store the sources in a temporary file whose name is derived from a hash of its content, which works well with ninja cache compile.

yaoyaoding: I see.

We could hash the contents of the source code in `build_inline` to get a directory to store the source code like `<tvm-ffi-cache-root>/sources/<name>-<hash>`, and use the source files to call `build`.

yaoyaoding:
> Without [this commit](https://github.com/apache/tvm-ffi/pull/213/commits/aabfb6892dd42a867f19a3545a4f985e9017dbf2), old code compiled via load_inline or build_inline will trigger a recompile, because the intermediate object name changed.

I feel recompilation for `load_inline` is okay when we upgrade tvm-ffi to a new version.

tqchen:
>I feel recompilation for load_inline is okay when we upgrade tvm-ffi to a new version

agree

yzh119:
Would you mind also adding unittests for testing the filelock behavior (we met lots of deadlock issues with similar functions in flashinfer and they cause trouble in multi-GPU inference setting).

yaoyaoding (changes requested):
Thanks @DarkSharpness ! The implementation looks good to me.

It would be better if we could have some test for `build` and `build_and_load` under `tests/python/cpp_extension/test_build.py`. Also move the `tests/python/test_load_inline.py` and `tests/python/test_build_inline.py` to the test folder as well.

yaoyaoding:
> Would you mind also adding unittests for testing the filelock behavior (we met lots of deadlock issues with similar functions in flashinfer and they cause trouble in multi-GPU inference setting).

I can add test for `FileLock` in another PR. 

Update: added #216

tqchen:
@yaoyaoding did your comment get addressed? would be good to see if we can get another round of review

yaoyaoding:
hi @DarkSharpness , do you mind adding some tests to the new added functions, `build` and `build_and_load`?

DarkSharpness:
test added cc @yaoyaoding

yaoyaoding:
Thanks @DarkSharpness , LGTM!

PR #224, opened 2025-11-04
[EXTRA] Improve the Error propagation in alloc

tqchen:
This PR updates the env_conetxt.cc to check the ret code first before checking the dl_tensor being not null, this allows us to effectively propagate error message from the caller if any to the consumer.
Test case updated to guard this regression.


yzh119 (commented):
Can we capture the error from pytorch APIs (which should be more explicit)?

tqchen:
@yzh119 this is being done in https://github.com/apache/tvm-ffi/blob/main/python/tvm_ffi/utils/_build_optional_torch_c_dlpack.py#L534 which cacths exception thrown and set e.what as error message.

tqchen:
noted that we should set the error type to MemoryError, updated in patch, cc @Kathryn-cat

PR #228, opened 2025-11-05
[Feature] Support vanilla C++ STL in FFI.

DarkSharpness:
Similar to pybind, we add a `stl.h` which support `array`, `vector`, `tuple`, `optional` and `variant`. After this file is included, users can use native C++ components, which could hopefully improve compatibility and reduce manual effort in converting between tvm::ffi components to C++ STL components.

~~We also modify the `function_detail.h` a little, so that we support all kinds of argument type (`T`, `const T`, `T&`, `const T&`, `T&&`, `const T&&` have been tested) in C++ exported functions.~~

Example code:

```cpp
#include <tvm/ffi/container/array.h>
#include <tvm/ffi/dtype.h>
#include <tvm/ffi/error.h>
#include <tvm/ffi/extra/stl.h>
#include <tvm/ffi/function.h>

#include <algorithm>
#include <array>
#include <cstddef>
#include <numeric>
#include <optional>
#include <variant>
#include <vector>

namespace {

// optional, array, vector, tuple is supported
auto sum_row(std::optional<std::vector<std::array<int, 2>>> arg)
    -> std::tuple<bool, std::vector<int>> {
  if (arg) {
    auto result = std::vector<int>{};
    result.reserve(arg->size());
    for (const auto& row : *arg) {
      result.push_back(std::accumulate(row.begin(), row.end(), 0));
    }
    return {true, result};
  } else {
    return {false, {}};
  }
}

// const reference is also supported, though not recommended and won't bring performance gain
// all types must be cast to value, and then pass by reference
auto find_diff(const std::vector<int>& a, std::vector<int> b) -> std::size_t {
  const auto max_pos = std::min(a.size(), b.size());
  for (std::size_t i = 0; i < max_pos; ++i) {
    if (a[i] != b[i]) {
      return i;
    }
  }
  return max_pos;
}

auto test_variant(std::variant<int, float, std::vector<int>>&& arg)
    -> std::variant<int, std::vector<int>> {
  if (std::holds_alternative<int>(arg)) {
    std::vector<int> result;
    auto& value = std::get<int>(arg);
    result.reserve(value);
    for (int i = 0; i < value; ++i) {
      result.push_back(i);
    }
    return result;
  } else if (std::holds_alternative<float>(arg)) {
    return static_cast<int>(std::get<float>(arg));
  } else {
    auto& value = std::get<std::vector<int>>(arg);
    std::reverse(value.begin(), value.end());
    return std::move(value);
  }
}

TVM_FFI_DLL_EXPORT_TYPED_FUNC(sum_row, sum_row);
TVM_FFI_DLL_EXPORT_TYPED_FUNC(find_diff, find_diff);
TVM_FFI_DLL_EXPORT_TYPED_FUNC(test_variant, test_variant);

}  // namespace


}  // namespace

```

Python part:
```python
from __future__ import annotations

from tvm_ffi.cpp import load_inline
from pathlib import Path

cur_path = Path(__file__).parent

with open(cur_path / "stl.cpp") as f:
    cpp_source = f.read()

module = load_inline(
    "test_stl",
    cpp_sources = cpp_source,
)

print(module.sum_row([[1, 2], [3, 4]]))  # Expected output: (True, [3, 7])
print(module.sum_row(None))  # Expected output: (False, [])
print(module.find_diff([1, 2, 3, 4], [1, 2, 4, 3]))  # Expected output: 2 (index = 2)
print(module.test_variant(2))  # Expected output: [0, 1]
print(module.test_variant(3.1))  # Expected output: 3
print(module.test_variant([1, 2]))  # Expected output: [2, 1]
```


include/tvm/ffi/extra/stl.h
```diff
@@ -0,0 +1,421 @@
...
+      TypeTraits<std::nullptr_t>::CopyToAnyView(nullptr, result);
+    }
+  }
+  TVM_FFI_INLINE static void MoveToAny(Self&& src, TVMFFIAny* result) {
+    if (src.has_value()) {
+      TypeTraits<Self>::MoveToAny(std::move(*src), result);
+    } else {
+      TypeTraits<std::nullptr_t>::MoveToAny(nullptr, result);
+    }
+  }
```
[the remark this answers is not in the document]
DarkSharpness: true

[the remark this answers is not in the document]
DarkSharpness: fixed

tqchen:
please split the core changes `function_details` into a separate PR. 

likely stl should go into extra. The main reason is that while stl can be helpful sometimes, the tradeoff is when present in cases like object, we can no longer to near zero cost access in static languages like rust via directly ptr and access (otherwise it have to be a call to reflection getter that does conversion under the hood, so we should always encourage ffi types when possible (and document such rationales in stl.h).

DarkSharpness:
> please split the core changes `function_details` into a separate PR.
> 
> likely stl should go into extra. The main reason is that while stl can be helpful sometimes, the tradeoff is when present in cases like object, we can no longer to near zero cost access in static languages like rust via directly ptr and access (otherwise it have to be a call to reflection getter that does conversion under the hood, so we should always encourage ffi types when possible (and document such rationales in stl.h).

I agree that stl should go into extra. The main goal of this PR is to introduce a pybind/PyTorch-style interface that provides greater flexibility for user-defined conversions. In performance-critical paths, however, FFI types should always take priority. This change only performs a one-time transformation for convenience and potential backward compatibility with existing code.

tqchen:
@DarkSharpness that sounds good, can we still move function_details.h chaneg into a separate PR? mainly want to make sure core changes are clearly scoped

DarkSharpness:
@tqchen now in #229 . Also tried to fix the `dependent template lookup` error of in the [CI failure](https://github.com/apache/tvm-ffi/actions/runs/19114299926/job/54619569681?pr=228). I will rebase after #229 is merged.

More tests of this PR is working in progress.

DarkSharpness:
cc @tqchen . We now support `array`, `vector`, `map`, `unordered_map`, `optional`, `tuple`, `variant` and `function` in this PR.

For STL containers, when passed from Python to C++, we will always construct from the Python object. When passed from C++ to Python, we will construct the corresponding object (e.g. `std::array` -> `Array`) first, and then move it to `AnyView`.

We also allow types that doesn't enable storage (e.g. `std::string`) to appear in STL container (e.g. `std::vector<std::string>`) to provide better compatibility.

Finally, do you think I should split the `stl.h` into multiple files? I'm not sure whether it is too long and include too many unwanted headers (e.g. user may only need to use `vector`).

tqchen:
Thanks @DarkSharpness , would be good to get some reviews, cc @junrushao @Ubospica @Kathryn-cat

DarkSharpness:
rebased cc @junrushao . Yes, your understanding is right. Since STL is not native to tvm::ffi, type conversion will happen at each Python <-> C++. I'm not sure where to document this. I have added a warning here in `stl.h`:

```cpp
/*!
 * \file tvm/ffi/extra/stl.h
 * \brief STL container support.
 * \note This file is an extra extension of TVM FFI,
 * which provides support for STL containers in C++ exported functions.
 *
 * Whenever possible, prefer using tvm/ffi/container/ implementations,
 * such as `tvm::ffi::Array` and `tvm::ffi::Tuple`, over STL containers
 * in exported functions for better performance and compatibility.
 */
 ```

File: tests/python/cpp_src/test_stl.cc [hunk 2]
```diff
@@ -0,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
```
tqchen, line 9: move this to tests/python/cpp_src

include/tvm/ffi/extra/stl.h
```diff
@@ -0,0 +1,646 @@
...
+
+/*!
+ * \file tvm/ffi/extra/stl.h
+ * \brief STL container support.
+ * \note This file is an extra extension of TVM FFI,
+ * which provides support for STL containers in C++ exported functions.
+ *
+ * Whenever possible, prefer using tvm/ffi/container/ implementations,
+ * such as `tvm::ffi::Array` and `tvm::ffi::Tuple`, over STL containers
+ * in exported functions for better performance and compatibility.
```
tqchen: Add more notes:

Native ffi objects comes with stable data layout and can be directly accessed through compiled languages (Rust) and DSLs(via LLVM) with raw pointer access for better performance and compatibility

junrushao:
Should be good to go once CI greens up

DarkSharpness:
This CI failures (seg-fault) make no sense to me. The latest commit is just a doc-string update. It's rather tricky as I failed to replay the seg-fault in my local environment. Still investigating on the root cause.

DarkSharpness:
cc @junrushao. The seg-fault disappeared after I rebased to main...

Some observation: in the [old ci](https://github.com/apache/tvm-ffi/actions/runs/19532442048/job/55918436762), all 240 tests are either passed or skipped, but one of the runner encounter seg-fault. So I guess it may be related some lifetime span errors like this one #264 . Still, I can not reproduce that seg-fault locally.

junrushao:
@DarkSharpness if you may, would you like to find out (by bisecting):
- which tests lead to segfault
- which commit leads to segfault
You may run the tests on your fork, using free github actions

DarkSharpness:
All 240 tests passed in my [fork](https://github.com/DarkSharpness/tvm-ffi/actions/runs/19615209372/job/56166811736). Really confused.

junrushao:
Perhaps it’s quite flaky. How many times did you trigger the test?

DarkSharpness:
For github ci, around 5 times? (only failed once in the previous CI)
For local environment, many times (`pytest -vvs tests/python`), yet never failed.

junrushao:
It might help if you enable Address Sanitizer over this Cython extension. It’s not entirely straightforward but I managed to do this once on my local dev machine: https://github.com/apache/tvm-ffi/pull/55

DarkSharpness:
Thanks @Ubospica . `std::pair` is very easy to integrate (just a special form of tuple).

Supporting std::any, however, is much more complicated. For foreign objects we only have their `TVMFFIAny` representation and a corresponding `type_index`, but we do not know the actual C++ type information behind them. To support `std::any`, we would need to assign a default fallback C++ type for every possible `type_index`. This might be technically possible, but it definitely would require substantial additional logic and invasive changes.

Given this, I would recommend using `tvm::ffi::Any` or `tvm::ffi::AnyView` instead. They integrate more naturally with the existing runtime and are generally more powerful than C++ `std::any`.

tqchen:
i agree that digging deep into the issue would be helpful, having asan checks can help

junrushao:
I was not able to reproduce the segfault either on my macbook

junrushao:
Can't reproduce the segfault on my x86 ubuntu machine either :((

junrushao:
I was not able to reproduce the segfault with ASan enabled on the specific commit that fails on github actions: https://github.com/apache/tvm-ffi/actions/runs/19532442048/job/55918436762

junrushao:
gpt-5.1-codex-max says:

> I dug through the new commit `fa8e0e5` and it only adds the header `include/tvm/ffi/extra/stl.h`, a new C++ test source `tests/python/cpp_src/test_stl.cc`, the Python driver `tests/python/test_stl.py`, and a single friend line in `include/tvm/ffi/container/map.h` to allow the STL type traits to call `MapObj::CreateFromRange`. No core runtime or shutdown logic was touched.
>
> I tried to reproduce locally by running `pytest tests/python/test_stl.py`, but the first build of that test module didn’t finish within the timeout, so I couldn’t observe a crash. There’s nothing in the diff that runs at interpreter teardown; the only new runtime path is the `test_stl` module exercising the new STL type traits.
>
> Given that, the segfault at pytest exit is likely either (a) pre‑existing in the core runtime, or (b) triggered by `test_stl` exercising a latent bug, but not caused by structural changes in this commit. To pin it down quickly, I’d suggest:
>
> 1. rerun the failing command with `PYTHONFAULTHANDLER=1` (or under `gdb --args pytest -vvs tests/python`) to grab a C stack trace at the crash;
> 2. rerun the same pytest command on the parent commit (`7cd2e50`) to see if it still segfaults;
> 3. if you want a fast signal, run `pytest -vvs tests/python --deselect tests/python/test_stl.py` once; if the segfault still appears, it’s not tied to the new STL code.
>
> With a stack trace we can investigate the actual crashing thread/function and decide next fixes.

junrushao:
@tqchen I recommend that we merge this PR given the segfault is never reproducible, can't be detected by ASan, and very unlikely related to this particular PR itself.

junrushao:
[asan.patch](https://github.com/user-attachments/files/23835626/asan.patch)

The patch I used to enable ASan

tqchen:
sounds good to me, thanks for looking into things

junrushao:
Merging this PR - @DarkSharpness thank you for your cooperation and patience awaiting us walking through this FUD. We are super excited to see all your contribution!

junrushao:
Unfortunately clangd is complaining about potential memory leak 🤔 https://github.com/apache/tvm-ffi/actions/runs/19791217781/job/56704441914

include/tvm/ffi/extra/stl.h
```diff
@@ -0,0 +1,649 @@
...
+  static constexpr int32_t field_static_type_index = TypeIndex::kTVMFFIArray;
+
+ private:
+  template <std::size_t... Is, typename Tuple>
+  TVM_FFI_INLINE static ObjectPtr<ArrayObj> CopyToTupleImpl(std::index_sequence<Is...>,
+                                                            Tuple&& src) {
+    auto array = ArrayObj::Empty(static_cast<std::int64_t>(sizeof...(Is)));
+    auto dst = array->MutableBegin();
+    // increase size after each new to ensure exception safety
+    ((::new (dst++) Any(std::get<Is>(std::forward<Tuple>(src))), array->size_++), ...);
```
junrushao, line 125: 😯 Here's what clangd says:

```
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/extra/stl.h:125:58: error: 'src' used after it was forwarded [bugprone-use-after-move,-warnings-as-errors]
  125 |     ((::new (dst++) Any(std::get<Is>(std::forward<Tuple>(src))), array->size_++), ...);
      |                                                          ^
/Users/jshao/Projects/tvm-ffi/include/tvm/ffi/extra/stl.h:125:25: note: forward occurred here
  125 |     ((::new (dst++) Any(std::get<Is>(std::forward<Tuple>(src))), array->size_++), ...);
      |                         ^
```

Basically, we should not forward `Tuple` multiple times during index unpacking. Very subtle bug @DarkSharpness!

junrushao:
A few segfaults on GitHub action:
- The original one: https://github.com/apache/tvm-ffi/actions/runs/19532442048/job/55918436762
- On `main` nightly: https://github.com/apache/tvm-ffi/actions/runs/19796601863/job/56717832596#step:8:3516
- Latest one on my version bump PR: https://github.com/apache/tvm-ffi/actions/runs/19803545351/job/56734651884#step:11:288

tqchen:
it would be great to find out what is going on, maybe add asan? do we know if it is related to stl support?

tqchen:
some followup suspects notes from @junrushao, could be related to use-after-free in deleter
#297 https://github.com/apache/tvm-ffi/pull/298

PR #229, opened 2025-11-05
[misc] support various arg types

DarkSharpness:
Related PR #228.

We now support all kinds of argument type (T, const T, T&, const T&, T&&, const T&& have been tested) in C++ exported functions.

### DIFF
--- a/include/tvm/ffi/function_details.h
+++ b/include/tvm/ffi/function_details.h
@@ -144,6 +144,13 @@ using FGetFuncSignature = std::string (*)();
 
 /*!
  * \brief Auxilary argument value with context for error reporting
+ * \tparam Type The expected type of the argument.
+ * \note We use a template class with non-template operator conversion
+ * instead of a non-template class with template operator conversion.
+ * This is because template operator conversion doesn't play well with
+ * classes with template constructors.
+ * In this case, it may lead to some unintended compiler errors.
+ * An example of class can be `std::optional<T>`.
  */
 template <typename Type>
 class ArgValueWithContext {


tqchen:
likely the PR will benfit from careful reviews from a few folks cc @junrushao , @yaoyaoding @Ubospica

tqchen:
i guess another question for @DarkSharpness is how important is it to support all these signatures, like &&, since one thing we want to optimize for is compilation speed and reduce of template complexity is helpful.

For most cases I am aware of, I feel T and const T& might be just fine.

junrushao:
I do like having broader types supported in the system, in spirit that compilation time is not a top-1 concern.

Meanwhile, I'm quite curious though the effect of supporting like `&&` types, like, will our system transfer ownership from function inputs in function call? It's quite unlikely, because `AnyView` itself doesn't own anything

DarkSharpness:
@tqchen I think this support can remain optional. I agree that for FFI-exported functions, the only ownership marker that truly makes sense is `T` — even `const T&` doesn’t, since we cannot hold a direct reference (like Any&) to a foreign object. In fact, any type passed through FFI must first be converted to a value.

To clarify, the process is:

1. Convert the input to a value.
2. Pass that value to the inner function according to the function signature.

For `const T&` or `T&&`, C++ automatically extends the lifetime of the value until the end of the function call. For `T&`, we instead create temporary objects.

The purpose of this PR is mainly to improve compatibility and ease of use with native C++ interfaces, which aligns with the goal of PR #228. We complete the support for all possible signatures `T`, `const T`, `T &`, `const T &`, `T&&`, `const T&&`. If this behavior feels misleading, we can simply disable the nonsensical type markers (like T&) and instead provide a clear, user-friendly error message.

Currently, for `T&` arguments, we get the following error:

```
In file included from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function.h:30,
                 from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/dtype.h:29,
                 from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/container/tensor.h:29,
                 from /root/.cache/tvm-ffi/test_sbind_760f1ab214615fe2/main.cpp:1:
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h: In instantiation of ‘void tvm::ffi::details::unpack_call(std::index_sequence<Is ...>, const std::string*, const F&, const tvm::ffi::AnyView*, int32_t, tvm::ffi::Any*) [with R = int; long unsigned int ...Is = {0}; F = int(int&); std::index_sequence<Is ...> = std::integer_sequence<long unsigned int, 0>; std::string = std::__cxx11::basic_string<char>; int32_t = int]’:
/root/.cache/tvm-ffi/test_sbind_760f1ab214615fe2/main.cpp:17:1:   required from here
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h:207:15: error: cannot bind non-const lvalue reference of type ‘int&’ to an rvalue of type ‘int’
  207 |     *rv = R(f(ArgValueWithContext(args, Is, optional_name, f_sig)...));
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h:155:18: note:   after user-defined conversion: ‘tvm::ffi::details::ArgValueWithContext::operator Type() [with Type = int]’
  155 |   TVM_FFI_INLINE operator Type() {  // NOLINT(google-explicit-constructor)
      |                  ^~~~~~~~
ninja: build stopped: subcommand failed.
```

Another minor point of PR is that, for types with template constructors, current `ArgValue` will fail. We need to change non-template ArgValue + template operator into template ArgValue + non-template operator. [Online example](https://godbolt.org/z/18z5MbzTY). A real-world example is `std::optional` (yes, we met this problem in PR #228 ).

DarkSharpness:
@junrushao Regarding ownership, our current system does not transfer ownership to the callee. We always convert to value first, and then pass the value according to the function signature. The ownership transfer might be supported in the future.

yaoyaoding:
I asked Gemini to summarize the two design choices:
| Option | Pros ✅ | Cons ❌ |
| :--- | :--- | :--- |
| **1. Support All Signatures** (`T`, `const T`, `T&`, `const T&`, `T&&`, `const T&&`) | **Increased compatibility** and ease of use with existing native C++ interfaces. Provides **full C++ idiomatic support** for exported functions. | **Increased template complexity**, potentially slowing down compilation. **References (`T&`, `T&&`) are semantically misleading** because the FFI input is converted to a value first (no actual object to reference/move from). |
| **2. Limit Support** (`T`, `const T&`) | **Optimizes compilation speed** by reducing template complexity. Enforces **idiomatic FFI practice**, where input is a value or a read-only reference. Nonsensical signatures (`T&`, `T&&`) can be **explicitly disabled with clear errors**. | **Breaks compatibility** with native C++ interfaces using other signatures. May require more user adaptation/rewriting of existing C++ functions. |

I prefer the Option 2 that only support (`T` and `const T&`, and with more user-friendly error message for unsupported modifiers, as @DarkSharpness suggested):
1. Since we introduce new types like `ffi::TensorView`, etc for tensor passing. It's likely that the users need to write the TVM-FFI interface functions, instead of reusing existing functions. Thus, the advantage of option 1 is applicable to limited cases. 
2. It's a little confusion when using other modifiers because we first convert the arguments to intermediate values and then pass these intermediate values to the user's function. We can not actually own or modify the user provided argument with the original c++ semantics.

DarkSharpness:
It seems that restricting to `T`, `const T` and `const T &` will cause a compile error in tvm-ffi source code when `pip install -ve`. That's because some part of tvm-ffi source code (like [this](https://github.com/apache/tvm-ffi/blob/main/include/tvm/ffi/reflection/registry.h#L580)) passes a function that accepts a rvalue reference type. So I keeps the support for `T &&`. @yaoyaoding 

Now, for unsupported types, we directly give the following error (and that's the only error):

```
In file included from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function.h:30,
                 from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/dtype.h:29,
                 from /root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/container/tensor.h:29,
                 from /root/.cache/tvm-ffi/test_sbind_760f1ab214615fe2/main.cpp:1:
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h: In instantiation of ‘class tvm::ffi::details::ArgValueWithContext<int&, false>’:
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h:226:14:   required from ‘void tvm::ffi::details::unpack_call(std::index_sequence<Is ...>, const std::string*, const F&, const tvm::ffi::AnyView*, int32_t, tvm::ffi::Any*) [with R = int; long unsigned int ...Is = {0}; F = int(int&); std::index_sequence<Is ...> = std::integer_sequence<long unsigned int, 0>; std::string = std::__cxx11::basic_string<char>; int32_t = int]’
/root/.cache/tvm-ffi/test_sbind_760f1ab214615fe2/main.cpp:17:1:   required from here
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h:193:7: error: static assertion failed: Argument type in FFI function signature can only be T, const T, or const T& or T&&
  193 |       TypeSupported<Type>,
      |       ^~~~~~~~~~~~~~~~~~~
/root/tvm-ffi/python/tvm_ffi/../../include/tvm/ffi/function_details.h:193:7: note: ‘tvm::ffi::details::TypeSupported<int&>’ evaluates to false
ninja: build stopped: subcommand failed.
```

yaoyaoding:
Thanks @DarkSharpness ! 

@tqchen do you have any other thought?

include/tvm/ffi/function_details.h
```diff
@@ -64,6 +64,13 @@ template <typename T>
 static constexpr bool RetSupported =
     (std::is_same_v<T, Any> || std::is_void_v<T> || TypeTraits<T>::convert_enabled);
 
+/// NOTE: we only support `T`, `const T`, `const T&` and `T &&` as argument types
+template <typename T>
+static constexpr bool TypeSupported =
```
tqchen: can we fold into ArgSupported in this case?

include/tvm/ffi/function_details.h
```diff
@@ -134,11 +141,11 @@ struct FunctionInfo<R (Class::*)(Args...) const,
 /*! \brief Using static function to output typed function signature */
 using FGetFuncSignature = std::string (*)();
 
-/*!
- * \brief Auxilary argument value with context for error reporting
- */
+template <typename Type, bool = TypeSupported<Type>>
```
tqchen: I am mainly hesitant about this extra template generic here, if we restrict to `const T&, T, T&&` is this still necessary, or is it sufficient if we update ArgSupported to indicate arg is supported. The main consideration here is if we can reduce the extra template generic to reduce template complexity, it can help

include/tvm/ffi/function_details.h
DarkSharpness: Sure. I'm wondering how should we restrict the `ReturnType`. I think for ref-counted Object type, only returning value type `T` is useful. So we may only allow value type `T` and `const T` in return type (same as current implementation).

include/tvm/ffi/function_details.h
DarkSharpness: This is mainly for a cleaner error message. For unsupported type, we will only have one compile error here: `error: static assertion failed: Argument type in FFI function signature can only be T, const T, or const T& or T&&`. If we fold `TypeSupported` into `ArgSupported`, this is no longer needed.

include/tvm/ffi/function_details.h
DarkSharpness: anyway, I keep the original `RetTypeSupported` (which only support value type)

include/tvm/ffi/function_details.h
tqchen: Actually my original question was is the Type argument needed? 
 
Note that if we do ArgSupported right likely will trigger an message when we call FromTyped(that the particular function signature is not supported).

DarkSharpness: Yes, but it is not related to error message. The original `ArgValueWithContext` uses a template operator function, which may not be compatible with template constructor in some cases. You can see the example code [here](https://godbolt.org/z/KM5qzK5ME). For a template `ArgValueWithContext` with non-template operator function, this will no be a problem.

A real-world example of this can be `std::optional`.

Ubospica (commented):
@DarkSharpness Thanks for the contribution! Supporting more function parameter types make sense to me. Comments from @junrushao and @yaoyaoding are also very helpful.

I checked the impl in nanobind. Findings:
1. Nanobind supports `T`, `const T`, `T&`, and `T&&` as parameters. Its ownership model keeps the ownership in the Python side, and every python object holds an internal C++ object ptr. So it can actually use `T&` parameter to modify Python objects. 
2. It also supports `T*`, `T`, `T&` and `const T&` return types. `T*` and `T` return type transfers ownership to Python side, while `T&` and `const T&` return type copies the result as a new object to Python.

For us I think raising error for `T&` make sense since it is not semantically supported. Supporting `T&&` makes sense because we already support RValueRef in parameters. 

Regarding compilation time, I think here the main overhead is that we now need to instantiate ArgValueWithContext for every function parameter, while previously we instantiate it for every function. It should be fine if the function don't have a lot of parameters. We should be careful if the target project is very large, but the speed impact can be further evaluated.

Another difference of our impl vs nanobind is, nanobind automatically converts python obj to rvalue reference when calling functions with rvalue reference parameters. For us, we need to call `_move()`. This is a design decision and I am fine with both, but just wanted to point this out in case anyone has preference.

We should add more tests to this PR about accepting rvalue reference parameters and rejecting reference parameters. You can use static_assertion.

We should also add docs in cpp_guide.md to talk about parameter types.

Another discussion items is, shall we support more return types such as `T&`, `T*` and `const T&`? Now we only support return type `ObjectRef` and `Object*`. I don't feel it must be supported now, but supporting it, especially `T&` and `const T&` looks fine for me. 

cc @tqchen

include/tvm/ffi/function_details.h
```diff
@@ -137,8 +145,11 @@ using FGetFuncSignature = std::string (*)();
 /*!
  * \brief Auxilary argument value with context for error reporting
  */
+template <typename Type>
```
tqchen, line 155: Add Type to the doc `\tparam`, add comment on rationale why this is needed, because we need to support Type being template types such as `std::optional<T>`

tqchen:
For return types, we should be able to support anything that can be moved into an Any (so it is beyond ObjectRef).

tqchen:
Thanks everyone for discussions through this change, the core changes needs careful due delligence and it can helps us to have minimal, robust, and stable features for downstream

Ubospica:
@tqchen Thanks for merging this. I agree that any return type convertible to Any are supported now. I mainly wanted to mention the support for `const ObjectRef&` and `ObjectRef&` return type, which seems not supported for now.

PR #230, opened 2025-11-06
feat: Embed metadata for FFI exported functions

Kathryn-cat:
## Summary

This PR adds reflection metadata support for functions exported via `TVM_FFI_DLL_EXPORT_TYPED_FUNC`, enabling signature validation and memory effect analysis without invoking the function. This addresses the need for compiler frameworks (e.g., JAX/XLA, MLIR-based compilers) to validate third-party FFI modules at compile time.

## Core Changes

**1. `TVM_FFI_DLL_EXPORT_TYPED_FUNC` macro**
- Exports two symbols per function:
  - `__tvm_ffi_<name>`
  - `__tvm_ffi__metadata_<name>`

**Metadata format:**
```
{
  "type_schema": "{\"type\":\"ffi.Function\",\"args\":[...]}",
}
```
Based on discussion, we choose to not include C++ const information in metadata in this PR, and think about it as a followup.

**2. `TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC` macro**
- Exports three symbols per function:
  - `__tvm_ffi_<name>`
  - `__tvm_ffi__metadata_<name>`
  - `__tvm_ffi__doc_<name>`

**3. Module API** (C++)
- `Module::GetFunctionMetadata(name)` - Query metadata from loaded modules
- `Module::GetFunctionDoc(name)` - Query documentation
- Works with both static and dynamic linking

**4. Module API** (Python)
- `mod.get_function_metadata("func_name")` - Query metadata from loaded modules
- `mod.get_function_doc("func_name")` - Query documentation
- Might be good to keep methods private

## Tests

The tests include both unit tests and integration tests.

**C++ Tests:**
- `tests/cpp/test_metadata.cc` - Direct symbol tests
- `tests/cpp/test_function.cc` - Static linking scenario
- `tests/cpp/extra/test_module.cc` - Module API

**Python Tests:**
- `tests/python/test_metadata.py` - Unit test
- `tests/python/test_build.py` - Integration test

### DIFF
--- a/include/tvm/ffi/function.h
+++ b/include/tvm/ffi/function.h
@@ -888,6 +888,7 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
     TVM_FFI_SAFE_CALL_END();                                                           \
   }                                                                                    \
   }
+/// \endcond
 
 /*!
  * \brief Export typed function as a SafeCallType symbol that follows the FFI ABI.


include/tvm/ffi/function.h
```diff
@@ -878,6 +880,56 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+    static std::string metadata_json = []() {                                          \
+      std::ostringstream oss;                                                          \
+      oss << R"({"type_schema":)"                                                      \
+          << ::tvm::ffi::EscapeString(::tvm::ffi::String(FuncInfo::TypeSchema()));     \
+      oss << R"(,"arg_const":[)";                                                      \
+      ::tvm::ffi::details::ArgConstHelper<typename FuncInfo::ArgType>::Generate(oss);  \
+      oss << "]}";                                                                     \
+      return oss.str();                                                                \
+    }();                                                                               \
+    return metadata_json.c_str();                                                      \
```
tqchen: unfortunately this will leak, need to instead export a TVM_FFI_DLL_EXPORT_TYPED_FUNC that returns String, @junrushao do u know if we can use constexpr trick here?

include/tvm/ffi/function.h
```diff
@@ -878,6 +880,56 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
         reinterpret_cast<::tvm::ffi::Any*>(result));                                   \
     TVM_FFI_SAFE_CALL_END();                                                           \
   }                                                                                    \
+  TVM_FFI_DLL_EXPORT const char* __tvm_ffi__metadata_##ExportName() {                  \
```
tqchen: this function needs to sit behind a flag TVM_FFI_DLL_EXPORT_TYPED_FUNC_META_DATA, so it can be toggled on off, mainly for concerns of binary size etc.

include/tvm/ffi/function.h
```diff
@@ -878,6 +880,56 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+ *
+ * \note The final symbol names are:
+ *       - `__tvm_ffi_<ExportName>` (function)
+ *       - `__tvm_ffi__metadata_<ExportName>` (metadata)
+ *       - `__tvm_ffi__doc_<ExportName>` (docstring)
+ */
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC(ExportName, Function, DocString)           \
+  TVM_FFI_DLL_EXPORT_TYPED_FUNC(ExportName, Function)                                \
+  extern "C" {                                                                       \
+  TVM_FFI_DLL_EXPORT const char* __tvm_ffi__doc_##ExportName() { return DocString; } \
```
tqchen: if it is a constant var, we should instead do const char symbol directly instead of fn?

include/tvm/ffi/function.h
junrushao: Note that string escaping here is not constexpr, which means it’s not possible to make it constexpr.

include/tvm/ffi/function.h
Kathryn-cat: addressed

Kathryn-cat:
@junrushao let's get it reviewed and get it in

Kathryn-cat:
@junrushao a small thing to discuss on the Python API side, currently we expose 

- `get_function_metadata("func_name")`
-  `get_function_doc("func_name") `

on the Module level, per TQ's advice, we might think of hiding these two functions by private names and underscores. What's your opinion on this?

include/tvm/ffi/function.h
```diff
@@ -840,13 +840,56 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+    ::tvm::ffi::details::unpack_call<typename FuncInfo::RetType>(                      \
+        std::make_index_sequence<FuncInfo::num_args>{}, &name, Function,               \
+        reinterpret_cast<const ::tvm::ffi::AnyView*>(args), num_args,                  \
+        reinterpret_cast<::tvm::ffi::Any*>(result));                                   \
+    TVM_FFI_SAFE_CALL_END();                                                           \
+  }                                                                                    \
+  }
+
+// Internal implementation macro that optionally generates metadata export function
+#ifdef TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA
```
tqchen: TVM_FFI_DLL_EXPORT_INCLUDE_METADATA, predefine the macro as a flag value 0/1

include/tvm/ffi/function.h
```diff
@@ -840,13 +840,56 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA_IMPL_(ExportName, Function)         \
+  inline ::tvm::ffi::String __tvm_ffi_get_metadata_##ExportName() {                \
+    using FuncInfo = ::tvm::ffi::details::FunctionInfo<decltype(Function)>;        \
+    /* Generate metadata on demand */                                              \
+    std::ostringstream os;                                                         \
+    os << R"({"type_schema":)"                                                     \
```
tqchen: need to include sstream in function.h because we need to have this

include/tvm/ffi/function.h
tqchen: then we use `#if TVM_FFI_DLL_EXPORT_INCLUDE_METADATA`

include/tvm/ffi/function.h
```diff
@@ -863,22 +906,66 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+ *
+ * \endcode
+ *
+ * \note The final symbol names are:
+ *       - `__tvm_ffi_<ExportName>` (function)
+ *       - `__tvm_ffi__metadata_<ExportName>` (metadata - only when
+ * TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA is defined)
+ *       - `__tvm_ffi__doc_<ExportName>` (docstring)
+ */
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_WITH_DOC(ExportName, Function, DocString) \
```
tqchen: how about we just do TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC, ExportName, DocString

include/tvm/ffi/function_details.h
```diff
@@ -253,6 +253,41 @@ struct TypeSchemaImpl {
   }
 };
 
+/*!
+ * \brief Helper to detect const-ness of a type parameter.
+ * Used for memory effect annotation in function metadata.
+ */
+template <typename T>
+struct IsConstParam {
```
tqchen: this is the part worth discussing, since normally the arg const in c++ is not a good indicative factor of arg_const in desirable metadata (most libs won't obey this).
 
I would be conservative first and not include it, then think about asking users to explicit annotate. What we really want to say is an op take some as input and muates some others. That metadata perhaps should come from outside, or being explicitly annotated

Better to be conservative than having the wrong one. Let us first remove this part

cc @junrushao @Kathryn-cat

python/tvm_ffi/module.py
```diff
@@ -170,6 +180,93 @@ def get_function(self, name: str, query_imports: bool = False) -> core.Function:
...
+        :py:func:`tvm_ffi.get_global_func_metadata`
+            Get metadata for global registry functions.
+
+        """
+        metadata_str = _ffi_api.ModuleGetFunctionMetadata(self, name, query_imports)
+        if metadata_str is None:
+            return None
+        return json.loads(metadata_str)
+
+    def get_function_doc(self, name: str, query_imports: bool = False) -> str | None:
```
tqchen, line 235: make sure we can eager auto attach the doc to function, so likely python user do not need this function

python/tvm_ffi/module.py
```diff
@@ -170,6 +180,93 @@ def get_function(self, name: str, query_imports: bool = False) -> core.Function:
             raise AttributeError(f"Module has no function '{name}'")
         return func
 
+    def get_function_metadata(
```
tqchen, line 188: we can keep this function since metadata is not as common, make sure it aligns with function in global one

src/ffi/testing/testing.cc
```diff
@@ -282,6 +284,86 @@ TVM_FFI_STATIC_INIT_BLOCK() {
 namespace tvm {
 namespace ffi {
 
+// -----------------------------------------------------------------------------
+// Implementation functions for schema testing
+// These are registered via both GlobalDef and TVM_FFI_DLL_EXPORT_TYPED_FUNC
+// -----------------------------------------------------------------------------
+namespace schema_test_impl {
+
```
tqchen, line 293: move these tests files into a cpp.load_inline test instead, so we don't have to blow up the testing. Given we use the same mechanism as the global one, we don't need to test so many functions

tests/cpp/extra/test_module.cc
```diff
@@ -0,0 +1,103 @@
...
+#include <tvm/ffi/container/map.h>
+#include <tvm/ffi/extra/json.h>
+#include <tvm/ffi/extra/module.h>
+#include <tvm/ffi/function.h>
+
+namespace {
+
+using namespace tvm::ffi;
+
+String GetTestingLibFilename() {
```
tqchen: for this case, let us simply run the test from the python side, to keep things simple and remove this cxx test

tests/cpp/test_function.cc
```diff
@@ -285,6 +289,21 @@ int invoke_testing_add1(int x) {
 
 TEST(Func, InvokeExternC) { EXPECT_EQ(invoke_testing_add1(1), 2); }
 
+extern "C" int __tvm_ffi__metadata_testing_add1(void*, const TVMFFIAny*, int32_t, TVMFFIAny*);
+
+TEST(Func, StaticLinkingMetadata) {
+  String metadata_str =
+      Function::InvokeExternC(nullptr, __tvm_ffi__metadata_testing_add1).cast<String>();
+  Map<String, Any> metadata = json::Parse(metadata_str).cast<Map<String, Any>>();
+  EXPECT_TRUE(metadata.count("type_schema"));
```
tqchen, line 298: we can keep this one test here

tqchen:
Did one pass, main thing we should be careful is not using constness in cxx as a indicator of mutability(let us remove for now and discuss how to allow explicit annotation if needed).

Others are mostly nit

include/tvm/ffi/function_details.h
junrushao: Agreed that const-ness may be not reflective of the underlying needed.

However, I'd love to also bring up that idiomatic function signatures for ML kernels may look like:

```
SomeKernel(
  x1: IN, # const: input
  x2: IN, # const: input
  y1: IN-OUT, # mutable inputs
  y2: OUT, # output
)
```

where, having const-ness indicators would indeed help production needs. @christopherbate also mentioned this feature as desirable in our discussion.

Another example is `torchgen` package installed along side torch, which allows us to parse function schemas provided in [`native_functions.yaml`](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml) and potentially give const-ness and aliasing info:

```python
import torch
import torchgen
from torchgen.model import FunctionSchema

op = torch.ops.aten.add.Tensor

FunctionSchema.parse(str(op._schema))
```

My recommendation is that we don't have to include this feature in this PR, but follow-up discussion may be worthy.

python/tvm_ffi/module.py
```diff
@@ -170,6 +180,93 @@ def get_function(self, name: str, query_imports: bool = False) -> core.Function:
...
+        self, name: str, query_imports: bool = False
+    ) -> dict[str, Any] | None:
+        """Get metadata for a function exported from the module.
+
+        This retrieves metadata for functions exported via TVM_FFI_DLL_EXPORT_TYPED_FUNC
+        and when TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA is on, which includes type schema
```
junrushao: nit

```suggestion
        This retrieves metadata for functions exported via c:macro:`TVM_FFI_DLL_EXPORT_TYPED_FUNC`
        and when c:macro:`TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA` is on, which includes type schema
```

python/tvm_ffi/module.py
```diff
@@ -170,6 +180,93 @@ def get_function(self, name: str, query_imports: bool = False) -> core.Function:
...
+        metadata_str = _ffi_api.ModuleGetFunctionMetadata(self, name, query_imports)
+        if metadata_str is None:
+            return None
+        return json.loads(metadata_str)
+
+    def get_function_doc(self, name: str, query_imports: bool = False) -> str | None:
+        """Get documentation string for a function exported from the module.
+
+        This retrieves documentation for functions exported via TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC.
+        If the function was exported with TVM_FFI_DLL_EXPORT_TYPED_FUNC, this function will
```
junrushao: ditto

tests/cpp/test_function.cc
```diff
@@ -16,9 +16,13 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA
```
junrushao: I'm thinking if we want to make this `TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA` a compile-definition in CMake? CC: @tqchen

tests/python/test_build.py
```diff
@@ -31,11 +34,293 @@ def test_build_cpp() -> None:
...
+    mod.double_tensor(x, y)
+    numpy.testing.assert_allclose(y, x * 2.0)
+    metadata = mod.get_function_metadata("double_tensor")
+    assert metadata is not None
+    schema = TypeSchema.from_json_str(metadata["type_schema"])
+    assert str(schema) == "Callable[[Tensor, Tensor], None]"
+
+    # Explicitly cleanup all objects before module unload to avoid use-after-free
+    del metadata, schema, result, x, y, mod
+    gc.collect()
```
junrushao, line 201: I'm thinking if `gc` is not actually useful as TVM-FFI will cache the loaded DSOs IIRC

docs/get_started/quickstart.rst
```diff
@@ -76,7 +76,8 @@ Suppose we implement a C++ function ``AddOne`` that performs elementwise ``y = x
 
 
 The macro :c:macro:`TVM_FFI_DLL_EXPORT_TYPED_FUNC` exports the C++ function ``AddOne``
-as a TVM FFI compatible symbol with the name ``__tvm_ffi_add_one_cpu/cuda`` in the resulting library.
+as a TVM FFI compatible symbol ``__tvm_ffi_add_one_cpu/cuda``. If ``TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA`` is on,
```
junrushao: IIRC this way will get us hyperlink to the C macro doc

```suggestion
as a TVM FFI compatible symbol ``__tvm_ffi_add_one_cpu/cuda``. If c:macro:`TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA` is on,
```

include/tvm/ffi/function_details.h
tqchen: yes, we agree there are on the value of extra information in the schema, mainly on avoiding exposing the wrong information. So we should have followup on ways to expose these things correctly

include/tvm/ffi/function.h
Kathryn-cat: resolved.

include/tvm/ffi/function.h
Kathryn-cat: @tqchen as of now `TVM_FFI_DLL_EXPORT_TYPED_FUNC_WITH_DOC(ExportName, Function, DocString)` does both the functionalities of `TVM_FFI_DLL_EXPORT_TYPED_FUNC(ExportName, Function)` and export doc. 

Do you mean we separate export doc utility as `TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC(ExportName, DocString)` from `TVM_FFI_DLL_EXPORT_TYPED_FUNC(ExportName, Function)`?

include/tvm/ffi/function_details.h
Kathryn-cat: thanks for the discussion @tqchen @junrushao. Now I removed const-ness from this PR, and we can think about it as a followup.

python/tvm_ffi/module.py
Kathryn-cat: ok

tests/cpp/extra/test_module.cc
Kathryn-cat: removed

src/ffi/testing/testing.cc
Kathryn-cat: changed

tests/python/test_build.py
junrushao: Likely I’m wrong

Kathryn-cat: @junrushao i think `gc` is still required here to drop all references (schema, metadata, doc strings) while `mod` is alive, in order to avoid use-after-free issue.

I tested without `gc` and it leads to a segfault :(

junrushao: you are absolutely right kathy!

python/tvm_ffi/module.py
Kathryn-cat: @tqchen @junrushao I saw that `Function` currently has an attribute called `__doc__`. Do we overwrite it if we find get_function_doc(func_name)? Also, since `Function` is a class in Cython, I'm not sure if I could modify it without breaking things.

junrushao (approved):

include/tvm/ffi/function.h
```diff
@@ -881,22 +932,66 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+ * \sa ffi::TypedFunction, TVM_FFI_DLL_EXPORT_TYPED_FUNC
+ *
+ * \code
+ *
+ * int Add(int a, int b) {
+ *   return a + b;
+ * }
+ *
+ * TVM_FFI_DLL_EXPORT_TYPED_FUNC_WITH_DOC(
+ *     add,
```
tqchen, line 965: somehow my previous comment get missing somewhere, I think we can have separate macro 

TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC(ExportName, DocString) independent from the normal export. TVM_FFI_DLL_EXPORT_INCLUDE_METADATA can also togle this

python/tvm_ffi/cpp/extension.py
```diff
@@ -366,8 +371,14 @@ def _decorate_with_tvm_ffi(source: str, functions: Mapping[str, str]) -> str:
     ]
 
     for func_name, func_doc in functions.items():
-        sources.append(f"TVM_FFI_DLL_EXPORT_TYPED_FUNC({func_name}, {func_name});")
-        _ = func_doc  # todo: add support to embed function docstring to the tvm ffi functions.
+        if func_doc:
+            # Escape the docstring for C++ string literal
+            escaped_doc = func_doc.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n")
```
tqchen: Leave it out for now since not too sure if we need doc attachment here. Also replace is less robust, maybe we should have a function _escape_cpp_string_literal, but this is not needed here

python/tvm_ffi/cpp/extension.py
```diff
@@ -40,7 +40,7 @@ def _hash_sources(
     cuda_source: str | None,
     cpp_files: Sequence[str] | None,
     cuda_files: Sequence[str] | None,
-    functions: Sequence[str] | Mapping[str, str],
+    functions: Sequence[str] | Mapping[str, str | None],
```
tqchen: it is strange to have Optional None here, do we know why?

python/tvm_ffi/cpp/extension.py
```diff
@@ -366,8 +371,14 @@ def _decorate_with_tvm_ffi(source: str, functions: Mapping[str, str]) -> str:
     ]
 
     for func_name, func_doc in functions.items():
-        sources.append(f"TVM_FFI_DLL_EXPORT_TYPED_FUNC({func_name}, {func_name});")
-        _ = func_doc  # todo: add support to embed function docstring to the tvm ffi functions.
+        if func_doc:
+            # Escape the docstring for C++ string literal
+            escaped_doc = func_doc.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n")
+            sources.append(
```
tqchen: not too sure if we need to export docs here atm, perhaps @yaoyaoding can also comment, we can skip first and update it as a followup

tqchen:
@yaoyaoding would be good to get your take on exporting docs on the load_inline, for now we can split it out first and add it as a followup

python/tvm_ffi/cpp/extension.py
yaoyaoding: If we allow `None` here, it seems allow the user to only provide docstring for some functions while leave docs for other function be None. I feel it's not very necessary here, the user can use empty string as a placeholder for the functions do not need a docstring.

python/tvm_ffi/cpp/extension.py
yaoyaoding: Except the escaping as @tqchen mentioned, the current implementation looks good to me. I feel we can keep the implementation as it is, and fix if there is any issue.

yaoyaoding:
> @yaoyaoding would be good to get your take on exporting docs on the load_inline, for now we can split it out first and add it as a followup

I feel we can add the docstring to match the behavior of our load_inline and pytorch's. The current implementation is a good start and we can fix any issue when it occurred.

python/tvm_ffi/cpp/extension.py
yaoyaoding: If we allow `None` here, it seems allow the user to only provide docstring for some functions while leave docs for other function be None. But I feel it's not very necessary here to allow `None`, the user can use empty string as a placeholder for the functions do not need a docstring.

tqchen:
@Kathryn-cat can you do a pass on latest comment
- Explicit docs macro TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC (docs still toggled  by TVM_FFI_DLL_EXPORT_INCLUDE_METADATA)
- Use empty string for functions do not need docs
- explicit _escape_cpp_string_literal, cc @junrushao to cross check the best way of impl this func

python/tvm_ffi/cpp/extension.py
Kathryn-cat: addressed

include/tvm/ffi/function.h
Kathryn-cat: resolved

python/tvm_ffi/cpp/extension.py
Kathryn-cat: addressed, @junrushao can double check

include/tvm/ffi/function.h
```diff
@@ -881,22 +932,65 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+ *     "    Sum of a and b");
+ *
+ * \endcode
+ *
+ * \note The exported symbol name is `__tvm_ffi__doc_<ExportName>` (docstring getter function).
+ *       This symbol is only exported when TVM_FFI_DLL_EXPORT_INCLUDE_METADATA is defined.
+ */
+#if TVM_FFI_DLL_EXPORT_INCLUDE_METADATA
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_DOC(ExportName, DocString)      \
+  static inline ::tvm::ffi::String __tvm_ffi_get_doc_##ExportName() { \
```
tqchen: avoid static inline, just the function static, since inline can have different indication of linkage

include/tvm/ffi/function.h
```diff
@@ -858,13 +869,53 @@ inline int32_t TypeKeyToIndex(std::string_view type_key) {
...
+        reinterpret_cast<const ::tvm::ffi::AnyView*>(args), num_args,                  \
+        reinterpret_cast<::tvm::ffi::Any*>(result));                                   \
+    TVM_FFI_SAFE_CALL_END();                                                           \
+  }                                                                                    \
+  }
+
+// Internal implementation macro that optionally generates metadata export function
+#if TVM_FFI_DLL_EXPORT_INCLUDE_METADATA
+#define TVM_FFI_DLL_EXPORT_TYPED_FUNC_METADATA_IMPL_(ExportName, Function)             \
+  inline ::tvm::ffi::String __tvm_ffi_get_metadata_##ExportName() {                    \
```
tqchen: avoid inline, instead use static, so it won'e cause conflict, actually can we "inline" it 

```c++
extern "C" {                                                                         \
  TVM_FFI_DLL_EXPORT int __tvm_ffi__metadata_##ExportName(void* self, const TVMFFIAny* args,     \
                                                int32_t num_args, TVMFFIAny* result) { \
    TVM_FFI_SAFE_CALL_BEGIN();                                                         \
    *result = String();...\
    TVM_FFI_SAFE_CALL_END();                                                           \
  }                                                                                    \
  }
```

include/tvm/ffi/function.h
tqchen: do the samething to inline into the extern "C"

include/tvm/ffi/function.h
Kathryn-cat: @tqchen assigning String() to Any type seems not accepted, so I used `MoveToAny` for type conversion, let me know if my approach works.

Also I'm keeping `TVM_FFI_DLL_EXPORT_TYPED_FUNC_IMPL_` as a helper that called twice.

tqchen (approved):

PR #237, opened 2025-11-07
feat: Attach class definition site in metadata

junrushao:
This PR consists of a few changes:
- Make every `Object` subclass has a class attribute `__tvm_ffi_type_info__` of type `TypeInfo`, and make sure every `TypeInfo` contains complete information, including `parent_type_info`, `type_cls` and others;
- Allocate a `TypeAttrColumn` called `__metadata__`, which contains all Objects' metadata;
- Record `__FILE__` and `__LINE__` in every `Object` definition on C++ end as constexpr, which will not impact binary size;
- Expose type attribute retrieval, currently as a private API `tvm_ffi.core._lookup_type_attr`.


python/tvm_ffi/cython/core.pyx
```diff
@@ -19,10 +19,18 @@
...
+_register_object_by_index(kTVMFFIDataType, DataType)
+include "./device.pxi"
+_register_object_by_index(kTVMFFIDevice, Device)
 include "./string.pxi"
+_register_object_by_index(kTVMFFIStr, String)
+_register_object_by_index(kTVMFFIBytes, Bytes)
 include "./tensor.pxi"
+_register_object_by_index(kTVMFFITensor, Tensor)
 include "./function.pxi"
+_register_object_by_index(kTVMFFIFunction, Function)
```
tqchen, line 36: why move registration to here?

python/tvm_ffi/cython/object.pxi
```diff
@@ -487,6 +488,17 @@ cdef _update_registry(int type_index, object type_key, object type_info, object
...
     TYPE_INDEX_TO_INFO[type_index] = type_info
     TYPE_KEY_TO_INFO[type_key] = type_info
+    if type_cls is not None:
+        if "__tvm_ffi_type_info__" not in type_cls.__dict__:
+            setattr(type_cls, "__tvm_ffi_type_info__", type_info)
+        elif (dst_type_info := type_cls.__tvm_ffi_type_info__) is not type_info:
+            dst_type_info.type_cls = type_info.type_cls
+            dst_type_info.type_index = type_info.type_index
+            dst_type_info.type_key = type_info.type_key
+            dst_type_info.type_ancestors = type_info.type_ancestors
```
tqchen, line 498: given type info may not be as freq used, it maybe more desirable still keep the original query by index/class and key function and just let user query instead of eager population

tqchen (changes requested):

tqchen:
after discussing with @junrushao , perhaps we can be a bit conservative here and minimize the metadata info to the fn type info that we had for now

junrushao:
While we are not embedding def site per discussion, I split out two useful parts of the PR into:
- #246
- #247

PR #239, opened 2025-11-08
[FIX] Fix missing static registration for DLTensor*

tqchen:
This PR fixes the missing static registration for DLTensor*


junrushao (approved):
Thanks @christopherbate @tqchen

PR #246, opened 2025-11-08
fix(cython): Make sure `TypeInfo` is properly registered by all classes

junrushao:
Before this commit, `parent_type_info` of a few Cython classes is not properly propagated, because of the registration ordering. This PR fixes this behavior and adds tests to safeguard this case.


python/tvm_ffi/cython/core.pyx
```diff
@@ -19,10 +19,18 @@
 # under the License.
 include "./base.pxi"
 include "./type_info.pxi"
-include "./dtype.pxi"
-include "./device.pxi"
 include "./object.pxi"
+_register_object_by_index(kTVMFFIObject, Object)
```
tqchen, line 27: would be good to document the rationale here(why register here), is to ensure ordering

junrushao: updated

PR #249, opened 2025-11-09
feat: Introduce `tvm.registry.get_registered_type_keys()`

junrushao:
This API is going to be useful for library stub generation, where we need to query objects registered on C++ end.


python/tvm_ffi/registry.py
```diff
@@ -346,9 +346,22 @@ def __init__invalid(self: Any, *args: Any, **kwargs: Any) -> None:
     raise RuntimeError("The __init__ method of this class is not implemented.")
 
 
+def get_valid_type_keys() -> list[str]:
```
tqchen: get_registered_type_keys

junrushao: updated

junrushao:
Updated and added a new test

tqchen (approved):

PR #266, opened 2025-11-15
[Minor] use perfect forwarding for template types.

### LINKED ISSUE #261: [Question] Why don't we use perfect forward for TypedFunction?
https://github.com/apache/tvm-ffi/blob/7f3f8726156ab6e33f781562afafd9c6f219551f/include/tvm/ffi/function.h#L714-L718

When constructed from C++ functions, this will lead to one unnecessary copy. A quick example is the code in PR #228 (support for STL), where we construct `TypedFunction` from native C++ functions.

A (hopefully) better implementation can be use perfect forwarding for all types except for `TypeFunction` itself.

DarkSharpness:
We use copy-and-move for types cheap to copy (e.g. `Function`, `std::string`), and use perfect forwarding to avoid any unnecessary copy/move construction during forwarding chain.

close #261

### DIFF
--- a/include/tvm/ffi/function.h
+++ b/include/tvm/ffi/function.h
@@ -503,6 +503,7 @@ class Function : public ObjectRef {
   template <typename TCallable>
   static Function FromTyped(TCallable&& callable) {
     using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
+    // Callable is always captured by value here to avoid possible dangling reference
     auto call_packed = [callable = std::forward<TCallable>(callable)](
                            const AnyView* args, int32_t num_args, Any* rv) mutable -> void {
       details::unpack_call<typename FuncInfo::RetType>(
@@ -519,6 +520,7 @@ class Function : public ObjectRef {
   template <typename TCallable>
   static Function FromTyped(TCallable&& callable, std::string name) {
     using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
+    // Callable is always captured by value here to avoid possible dangling reference
     auto call_packed = [callable = std::forward<TCallable>(callable), name = std::move(name)](
                            const AnyView* args, int32_t num_args, Any* rv) mutable -> void {
       details::unpack_call<typename FuncInfo::RetType>(


include/tvm/ffi/function.h
```diff
@@ -145,14 +146,22 @@ namespace details {
 template <typename TCallable>
 class FunctionObjImpl : public FunctionObj {
  public:
-  using TStorage = std::remove_cv_t<std::remove_reference_t<TCallable>>;
```
tqchen, line 148: is this change intententionally? We want to explicitly avoid the case where TCallable is a reference and resulted a shallow capture of the closure in previous impl, would be good o confirm the impl logic here.

DarkSharpness: `TCallable` cannot be a reference type in our code. The only usage is here:

`using ObjType = typename details::FunctionObjImpl<std::decay_t<TCallable>>;`

We manually decay it to value type, so we no longer need to remove cv ref qualifier again in storage.

include/tvm/ffi/function.h
```diff
@@ -145,14 +146,22 @@ namespace details {
 template <typename TCallable>
```
tqchen, line 147: document: Invariance: TCallable cannot be reference

include/tvm/ffi/function.h
```diff
@@ -611,11 +621,10 @@ class Function : public ObjectRef {
    * \param packed_call The packed function signature
    */
   template <typename TCallable>
-  static Function FromPackedInternal(TCallable packed_call) {
-    using ObjType = typename details::FunctionObjImpl<TCallable>;
+  static Function FromPackedInternal(TCallable&& packed_call) {
+    using ObjType = typename details::FunctionObjImpl<std::decay_t<TCallable>>;
```
tqchen, line 632: document: need to make sure we create ObjectType that do not come with references

include/tvm/ffi/function.h
tqchen: I see, in this case, would be useful to add a static assert here to indicate it is not cv ref (for maintainability). As well as a document

include/tvm/ffi/function.h
```diff
@@ -487,10 +497,10 @@ class Function : public ObjectRef {
    * \param callable the internal container of packed function.
    */
   template <typename TCallable>
-  static Function FromTyped(TCallable callable) {
-    using FuncInfo = details::FunctionInfo<TCallable>;
-    auto call_packed = [callable = std::move(callable)](const AnyView* args, int32_t num_args,
-                                                        Any* rv) mutable -> void {
+  static Function FromTyped(TCallable&& callable) {
+    using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
+    auto call_packed = [callable = std::forward<TCallable>(callable)](
```
tqchen, line 507: what happens if TCallable here is a const reference? will the captured callable being a constant ref? Where is FuncInfo being used here

include/tvm/ffi/function.h
```diff
@@ -503,9 +513,9 @@ class Function : public ObjectRef {
    * \param name optional name attacked to the function.
    */
   template <typename TCallable>
-  static Function FromTyped(TCallable callable, std::string name) {
-    using FuncInfo = details::FunctionInfo<TCallable>;
-    auto call_packed = [callable = std::move(callable), name = std::move(name)](
+  static Function FromTyped(TCallable&& callable, std::string name) {
+    using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
+    auto call_packed = [callable = std::forward<TCallable>(callable), name = std::move(name)](
```
tqchen, line 524: what happens if TCallable here is a const reference? will the captured callable being a constant ref? Where is FuncInfo being used here

tqchen (changes requested):

include/tvm/ffi/function.h
DarkSharpness: In lambda expression, a capture without `&` will always store by value (init-capture in [cppreference](https://en.cppreference.com/w/cpp/language/lambda.html)). So callable capture is by value (just as `std::decay_t<TCallable>` instead of by reference. The `forward` here just means we will choose to copy or move construct the inner-captured value.

`FuncInfo` is inherited from the old code. It is used to get the return type and num of arguments:

```cpp
  template <typename TCallable>
  static Function FromTyped(TCallable&& callable) {
    using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
    auto call_packed = [callable = std::forward<TCallable>(callable)](
                           const AnyView* args, int32_t num_args, Any* rv) mutable -> void {
      details::unpack_call<typename FuncInfo::RetType>(
          std::make_index_sequence<FuncInfo::num_args>{}, nullptr, callable, args, num_args, rv);
    };
    return FromPackedInternal(std::move(call_packed));
  }
  ```

include/tvm/ffi/function.h
DarkSharpness: In lambda expression, a capture without `&` will always store by value (init-capture in [cppreference](https://en.cppreference.com/w/cpp/language/lambda.html)). So callable capture is by value (just as `std::decay_t<TCallable>` instead of by reference. The `forward` here just means we will choose to copy or move construct the inner-captured value.

`FuncInfo` is inherited from the old code. It is used to get the return type and num of arguments:

```cpp
  template <typename TCallable>
  static Function FromTyped(TCallable&& callable) {
    using FuncInfo = details::FunctionInfo<std::decay_t<TCallable>>;
    auto call_packed = [callable = std::forward<TCallable>(callable)](
                           const AnyView* args, int32_t num_args, Any* rv) mutable -> void {
      details::unpack_call<typename FuncInfo::RetType>(
          std::make_index_sequence<FuncInfo::num_args>{}, nullptr, callable, args, num_args, rv);
    };
    return FromPackedInternal(std::move(call_packed));
  }
  ```

include/tvm/ffi/function.h
tqchen: thanks for the explaination, can we add a comment here stating that callable is always pass by value

include/tvm/ffi/function.h
tqchen: thanks for the explaination, can we add a comment here stating that callable is always pass by value

include/tvm/ffi/function.h
DarkSharpness: done

tqchen:
Thanks @DarkSharpness

PR #272, opened 2025-11-17
[TORCH] Remove ninja dep on non-windows when JIT optional torch-c-dlpack

tqchen:
This PR removes ninja dep on non-windows platform when JITing torch-c-dlpack to reduce overall dependencies needed and streamline the experience.


tqchen:
cc @yaoyaoding @cyx-6 @Hzfengsy

PR #283, opened 2025-11-22
[Feature] Add cubin launcher utility as an extra header

yaoyaoding:
This PR adds `include/tvm/ffi/extra/cubin_launcher.h`, a header only utility, that loads cubin from byte buffer or a file and launch it. The utility is based on CUDA Runtime API. It's not compiled with the `libtvm_ffi.so`, but shipped with the package. 

--- 

**Usage**
```python
import torch
from tvm_ffi import cpp
from tvm_ffi.cpp import nvrtc

# Step 1: Define CUDA kernel source
cuda_source = """
extern "C" __global__ void add_one(float* x, float* y, int n) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if (idx < n) {
        y[idx] = x[idx] + 1.0f;
    }
}
"""

# Step 2: Compile to CUBIN using NVRTC
cubin_bytes = nvrtc.nvrtc_compile(cuda_source, name="kernel.cu")

# Step 3: Define C++ wrapper with embedded CUBIN
cpp_wrapper = """
#include <tvm/ffi/error.h>
#include <tvm/ffi/extra/cuda/cubin_launcher.h>
#include <tvm/ffi/function.h>

// Declare embedded CUBIN module
TVM_FFI_EMBED_CUBIN(my_cubin);

void AddOne(tvm::ffi::TensorView x, tvm::ffi::TensorView y) {
  // Get kernel from embedded CUBIN (cached for efficiency)
  static auto kernel = TVM_FFI_EMBED_CUBIN_GET_KERNEL(my_cubin, "add_one");

  // Prepare kernel arguments
  int64_t n = x.size(0);
  void* x_ptr = x.data_ptr();
  void* y_ptr = y.data_ptr();
  void* args[] = {&x_ptr, &y_ptr, &n};

  // Configure launch parameters
  tvm::ffi::dim3 grid((n + 255) / 256);
  tvm::ffi::dim3 block(256);

  // Get CUDA stream and launch
  DLDevice device = x.device();
  CUstream stream = static_cast<CUstream>(
      TVMFFIEnvGetStream(device.device_type, device.device_id));

  CUresult result = kernel.Launch(args, grid, block, stream);
  TVM_FFI_CHECK_CUDA_DRIVER_ERROR(result);
}

TVM_FFI_DLL_EXPORT_TYPED_FUNC(add_one, AddOne);
"""

# Step 4: Load module with embedded CUBIN
mod = cpp.load_inline(
    "my_module",
    cuda_sources=cpp_wrapper,
    embed_cubin={"my_cubin": cubin_bytes}
)

# Step 5: Use the kernel
x = torch.arange(1024, dtype=torch.float32, device="cuda")
y = torch.empty_like(x)
mod.add_one(x, y)

# Verify results
assert torch.allclose(y, x + 1)
```

**C++ Core Usage**
```c++
// Load CUBIN module from memory
tvm::ffi::CubinModule module(cubin_data);

// Get kernel by name
tvm::ffi::CubinKernel kernel = module["my_kernel"];

// Launch kernel (same as embedded example)
void* args[] = {...};
tvm::ffi::dim3 grid(...);
tvm::ffi::dim3 block(...);
CUstream stream = ...;

CUresult result = kernel.Launch(args, grid, block, stream);
TVM_FFI_CHECK_CUDA_DRIVER_ERROR(result);
```

- documentation: see the guide at `docs/guides/cubin_launcher.rst`
- more examples: see more examples at `examples/cubin_launcher`

--- 

**Benchmark**
```
Kernel Launch Overhead Benchmark: Triton vs TVM-FFI
============================================================
              CPU: Intel(R) Core(TM) Ultra 9 285K
      CUDA device: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
  PyTorch version: 2.9.1+cu128
   Triton version: 3.5.1

Compiling empty Triton kernel to CUBIN...
Compiled CUBIN: 7,480 bytes

Benchmarking kernel launch overhead (10,000 calls)...
============================================================
  Triton launch                 :    4.911 μs/call
  TVM-FFI launch                :    2.034 μs/call

  Overhead: -58.59%
  TVM-FFI is 2.41x faster
============================================================
```
Benchmark script at `examples/cubin_launcher/benchmark_overhead.py`

Note: we did not check the dtype, shape constraints in triton's launch case, thus the number above is the lower-bound of a typical triton launch overhead. When the constraints are checked, the launch overhead of Triton will be higher. On the other hand, TVM-FFI checks the constraints in C++ and is fast.

--- 

**About Triton Example**
The triton example is tricky, since the cubin's kernel parameters is not aligned with the kernel definition of the triton kernel in python. I checked the generated PTX, there are 5 parameters while there is only 3 in the kernel definition. The last two are not used. We need some Triton expert to write the host side code the prepare the kernel parameters. To make it widely used, we need better documentation on the triton's calling convention regarding the generated cubin. But this is out of the scope of this PR.

### DIFF
--- a/python/tvm_ffi/cpp/extension.py
+++ b/python/tvm_ffi/cpp/extension.py
@@ -325,7 +325,7 @@ def _generate_ninja_build(  # noqa: PLR0915, PLR0912
 
             ninja.append("rule embed_cubin")
             ninja.append(
-                "  command = python -m tvm_ffi.utils.embed_cubin --output-obj $out --input-obj $in --cubin $cubin --name $name"
+                f"  command = {sys.executable} -m tvm_ffi.utils.embed_cubin --output-obj $out --input-obj $in --cubin $cubin --name $name"
             )
             ninja.append("")
 


CMakeLists.txt
```diff
@@ -208,6 +208,10 @@ option(TVM_FFI_BUILD_PYTHON_MODULE "Adding python module." OFF)
 if (TVM_FFI_BUILD_PYTHON_MODULE)
   # Helper function to build the cython module
   message(STATUS "Building cython module..")
+
+  # prefer virtualenv when searching for python
+  set(Python_FIND_VIRTUALENV FIRST) # cmake-lint: disable=C0103
```
yaoyaoding, line 213: See the documentation of [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html). This flag allows cmake to first use the python in the virtual env to build, when the virtualenv is activated. It's the desired behavior when the user is working in a virtualenv (like venv created by uv).

I add this modification to this PR since it's a small change. Happy to create a standalone one if needed.

yaoyaoding:
The Triton kernel:
```python
    # Define the kernel dynamically
    @triton.jit
    def square_kernel(X_ptr, Y_ptr, n, BLOCK: tl.constexpr = 1024):  # noqa
        pid = tl.program_id(0)
        start = pid * BLOCK
        offsets = start + tl.arange(0, BLOCK)
        mask = offsets < n
        x = tl.load(X_ptr + offsets, mask=mask, other=0.0)
        y = x * x
        tl.store(Y_ptr + offsets, y, mask=mask)

    # Trigger kernel compilation by doing a dummy call
    x_dummy = torch.ones(1024, dtype=torch.float32, device="cuda")
    y_dummy = torch.empty(1024, dtype=torch.float32, device="cuda")
    square_kernel[1, 1](x_dummy, y_dummy, 1024)
```

The PTX:
```
.visible .entry square_kernel(
        .param .u64 .ptr .global .align 1 square_kernel_param_0,
        .param .u64 .ptr .global .align 1 square_kernel_param_1,
        .param .u32 square_kernel_param_2,
        .param .u64 .ptr .global .align 1 square_kernel_param_3,
        .param .u64 .ptr .global .align 1 square_kernel_param_4
)
.reqntid 128
{
        .reg .pred      %p<5>;
        .reg .b32       %r<33>;
        .reg .b64       %rd<8>;
        .loc    1 53 0                          // example_triton_cubin.py:53:0
$L__func_begin0:
        .loc    1 53 0                          // example_triton_cubin.py:53:0

// %bb.0:
        ld.param.b64    %rd5, [square_kernel_param_0];
        ld.param.b64    %rd6, [square_kernel_param_1];
$L__tmp0:
        .loc    1 54 24                         // example_triton_cubin.py:54:24
        mov.u32         %r25, %ctaid.x;
        .loc    1 55 18                         // example_triton_cubin.py:55:18
        shl.b32         %r26, %r25, 10;
        ld.param.b32    %r27, [square_kernel_param_2];
        .loc    1 56 35                         // example_triton_cubin.py:56:35
        mov.u32         %r28, %tid.x;
        shl.b32         %r29, %r28, 2;
        and.b32         %r30, %r29, 508;
        .loc    1 56 22                         // example_triton_cubin.py:56:22
        or.b32  %r31, %r30, %r26;
        or.b32  %r32, %r31, 512;
        .loc    1 57 21                         // example_triton_cubin.py:57:21
        setp.lt.s32     %p1, %r31, %r27;
        setp.lt.s32     %p2, %r32, %r27;
        .loc    1 58 24                         // example_triton_cubin.py:58:24
        mul.wide.s32    %rd7, %r31, 4;
        add.s64         %rd1, %rd5, %rd7;
        add.s64         %rd2, %rd1, 2048;
        mov.b32         %r5, 0;
        .loc    1 58 16                         // example_triton_cubin.py:58:16
        // begin inline asm
        mov.u32 %r1, %r5;
        mov.u32 %r2, %r5;
        mov.u32 %r3, %r5;
        mov.u32 %r4, %r5;
        @%p1 ld.global.v4.b32 { %r1, %r2, %r3, %r4 }, [ %rd1 + 0 ];
        // end inline asm
        // begin inline asm
        mov.u32 %r9, %r5;
        mov.u32 %r10, %r5;
        mov.u32 %r11, %r5;
        mov.u32 %r12, %r5;
        @%p2 ld.global.v4.b32 { %r9, %r10, %r11, %r12 }, [ %rd2 + 0 ];
        // end inline asm
        .loc    1 59 12                         // example_triton_cubin.py:59:12
        mul.f32         %r17, %r1, %r1;
        mul.f32         %r18, %r2, %r2;
        mul.f32         %r19, %r3, %r3;
        mul.f32         %r20, %r4, %r4;
        mul.f32         %r21, %r9, %r9;
        mul.f32         %r22, %r10, %r10;
        mul.f32         %r23, %r11, %r11;
        mul.f32         %r24, %r12, %r12;
        .loc    1 60 21                         // example_triton_cubin.py:60:21
        add.s64         %rd3, %rd6, %rd7;
        add.s64         %rd4, %rd3, 2048;
        .loc    1 60 30                         // example_triton_cubin.py:60:30
        // begin inline asm
        @%p1 st.global.v4.b32 [ %rd3 + 0 ], { %r17, %r18, %r19, %r20 };
        // end inline asm
        // begin inline asm
        @%p2 st.global.v4.b32 [ %rd4 + 0 ], { %r21, %r22, %r23, %r24 };
        // end inline asm
        .loc    1 60 4                          // example_triton_cubin.py:60:4
        ret;
$L__tmp1:
$L__func_end0:
                                        // -- End function
}
```

The two extra parameters are not used. Maybe it's used for some features that not used in this simple kernel.

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,306 @@
...
+ * This header provides a lightweight C++ wrapper around CUDA Driver API
+ * for loading CUBIN modules and launching kernels. It supports:
+ * - Loading CUBIN from memory (embedded data) or files
+ * - Multi-GPU execution using CUDA primary contexts
+ * - Kernel parameter management and launch configuration
+ */
+#ifndef TVM_FFI_EXTRA_CUBIN_LAUNCHER_H_
+#define TVM_FFI_EXTRA_CUBIN_LAUNCHER_H_
+
+#include <cuda.h>
```
tqchen: let us move to extra/cuda/cubin_launcher.h

include/tvm/ffi/extra/cubin_launcher.h
```diff
@@ -0,0 +1,306 @@
...
+    if (__err != CUDA_SUCCESS) {                                                                   \
+      const char* __err_name = nullptr;                                                            \
+      const char* __err_str = nullptr;                                                             \
+      cuGetErrorName(__err, &__err_name);                                                          \
+      cuGetErrorString(__err, &__err_str);                                                         \
+      TVM_FFI_THROW(RuntimeError) << "CUDA Driver Error: "                                         \
+                                  << (__err_name ? __err_name : "UNKNOWN") << " ("                 \
+                                  << static_cast<int>(__err)                                       \
+                                  << "): " << (__err_str ? __err_str : "No description") << " at " \
+                                  << __FILE__ << ":" << __LINE__;                                  \
```
tqchen: no need to have file and line since they are part of TVM_FFI_THROW

include/tvm/ffi/extra/cubin_launcher.h
```diff
@@ -0,0 +1,306 @@
...
+        cuLibraryLoadData(&library_, data, nullptr, nullptr, 0, nullptr, nullptr, 0));
+  }
+
+  /*!
+   * \brief Load CUBIN module from file.
+   *
+   * \param filename Path to the CUBIN file.
+   * \note This reads the entire file into memory and then loads it.
+   */
+  explicit CubinModule(const char* filename) {
```
tqchen: to keep dep minimal let us consider always use the load from inline binary, add a constructor from tvm::ffi::Bytes

include/tvm/ffi/extra/cubin_launcher.h
```diff
@@ -0,0 +1,306 @@
...
+   * \brief Load CUBIN module from memory.
+   *
+   * \param data Pointer to CUBIN binary data in memory.
+   * \param size Size of the CUBIN binary data in bytes.
+   * \note Calls cuInit(0) to ensure CUDA is initialized.
+   */
+  CubinModule(const void* data, uint64_t size) {
+    TVM_FFI_CHECK_CUDA_DRIVER_ERROR(cuInit(0));
+    TVM_FFI_CHECK_CUDA_DRIVER_ERROR(
+        cuLibraryLoadData(&library_, data, nullptr, nullptr, 0, nullptr, nullptr, 0));
```
tqchen: would be cross check if the data needs to be retained in memory, and document accordingly

examples/cubin_launcher/src/lib_embedded.cc
```diff
@@ -0,0 +1,130 @@
...
+
+namespace cubin_embedded {
+
+/*!
+ * \brief Launch add_one_cuda kernel on input tensor.
+ * \param x Input tensor (float32, 1D)
+ * \param y Output tensor (float32, 1D, same shape as x)
+ */
+void AddOne(tvm::ffi::TensorView x, tvm::ffi::TensorView y) {
+  InitializeCubinModule();
```
tqchen: consider use a different pattern via static single singleton

```c++
// maybe as macro TVM_FFI_EMBED_CUBIN(env);
// use env as the indicator key in case we want to enbed other cubins
extern "C" const char __tvm_ffi__cubin_env[];
extern "C" const char __tvm_ffi__cubin_env_end[];

namespace {
struct EmbedCubinModule_env {
    tvm::ffi::CubinModule mod {__tvm_ffi__cubin_env, __tvm_ffi__cubin_env_end};  

    static EmbedCubinModule_env* Global() {
         static EmbedCubinModule_env* inst;
          return &inst;
    }
};
}  // anonymous namespace to avoid symbol conflict

void MulTwo(tvm::ffi::TensorView x, tvm::ffi::TensorView y) {
   // maybe as macro 
   //  static auto kernel =TVM_FFI_EMBED_CUBIN_GET_KERNEL(env, "mul_two_cuda");
   static tvm::ffi::CubinKernel kernel = EmbedCubinModule_env::Global()->mod["mul_two_cuda"];
   kernel->add_two_cuda.launch(...);
}
```

examples/cubin_launcher/src/lib_embedded.cc
```diff
@@ -0,0 +1,130 @@
...
+#include <tvm/ffi/extra/c_env_api.h>
+#include <tvm/ffi/extra/cubin_launcher.h>
+#include <tvm/ffi/function.h>
+
+#include <cstdint>
+#include <memory>
+
+// External symbols for embedded CUBIN data (linked via objcopy)
+extern "C" const char __cubin_data[];
+extern "C" const char __cubin_data_end[];
```
tqchen: one thing to consider is how to avoid symbol conflict when we link multiple such embedded files,

here is what gemini suggest, do it in the following steps, maybe we can have a tool via python `tvm.ffi.cpp.embed_cubin(output_object, input_object, cubin, key="env")` to do that. we can also provide a cmake macro for those who like to use tvm-ffi cmake

- Step 1: Compile the C++ Source (source.o) Compile your C++ code normally. Ensure you declare the symbols as extern so the compiler creates an "undefined reference" (a hole to be filled later). g++ -c source.cc -o source.o
- Step 2: Convert Binary to Object (blob_raw.o) Use objcopy to wrap the raw binary file into a linkable object file. This creates Global symbols by default. objcopy -I binary -O elf64-x86-64 kernel.cubin blob_raw.o
- Step 3: Rename Symbols (blob_renamed.o) Change the auto-generated names (e.g., _binary_kernel_start) to the specific names your C++ code expects (__tvm_ffi__...). objcopy --redefine-sym old_name=new_name blob_raw.o blob_renamed.o
- Step 4: Partial Link / Merge (merged.o) Use ld -r to fuse the code (source.o) and the data (blob_renamed.o) together. This resolves the "undefined reference." ld -r source.o blob_renamed.o -o merged.o
- Step 5: Localize Symbols (final.o) Crucial Last Step: Now that the code and data are in the same file, use objcopy to change the symbols from Global to Local. This hides them from the outside world (Internal Linkage). objcopy --localize-symbol=__tvm_ffi__cubin_data merged.o final.o

```make
# ==========================================
# Configuration
# ==========================================

# Files
BINARY_FILE := kernel.cubin
SOURCE_FILE := source.cc
OUTPUT_OBJ  := final_module.o

# The symbol names your C++ code uses (extern "C")
SYM_NAME      := __tvm_ffi__cubin_data
SYM_NAME_END  := __tvm_ffi__cubin_data_end

# Compiler settings
CXX      := g++
CXXFLAGS := -O2 -Wall -fPIC
LD       := ld
OBJCOPY  := objcopy

# ------------------------------------------
# Internal Calculation for objcopy default names
# objcopy converts "kernel.cubin" -> "_binary_kernel_cubin_start"
# We replace dots and slashes with underscores to match objcopy's behavior.
# ------------------------------------------
BINARY_FLAT   := $(subst /,_,$(subst .,_,$(BINARY_FILE)))
DEFAULT_START := _binary_$(BINARY_FLAT)_start
DEFAULT_END   := _binary_$(BINARY_FLAT)_end
DEFAULT_SIZE  := _binary_$(BINARY_FLAT)_size

# ==========================================
# Rules
# ==========================================

.PHONY: all clean check

all: $(OUTPUT_OBJ)

# 1. Compile the C++ source into an object file.
#    (Contains undefined references to the symbols)
source.o: $(SOURCE_FILE)
	@echo "[1/5] Compiling Source..."
	$(CXX) $(CXXFLAGS) -c $< -o $@

# 2. Convert the raw binary into an ELF object file.
#    (Symbols are Global and named _binary_kernel_cubin_start)
blob_raw.o: $(BINARY_FILE)
	@echo "[2/5] Converting Binary to Object..."
	$(OBJCOPY) -I binary -O elf64-x86-64 $< $@

# 3. Rename the symbols to match your C++ declaration.
#    (Still Global, but names match __tvm_ffi__...)
blob_renamed.o: blob_raw.o
	@echo "[3/5] Renaming Symbols..."
	$(OBJCOPY) \
		--redefine-sym $(DEFAULT_START)=$(SYM_NAME) \
		--redefine-sym $(DEFAULT_END)=$(SYM_NAME_END) \
		--strip-symbol=$(DEFAULT_SIZE) \
		$< $@

# 4. Partial Link (Merge).
#    (Fuses source.o and blob.o. Code can now see Data.)
merged.o: source.o blob_renamed.o
	@echo "[4/5] Linking (Partial Merge)..."
	$(LD) -r source.o blob_renamed.o -o $@

# 5. Localize Symbols.
#    (Hides the symbols from the outside world. Global D -> Local d)
$(OUTPUT_OBJ): merged.o
	@echo "[5/5] Finalizing: Hiding Symbols..."
	$(OBJCOPY) \
		--localize-symbol=$(SYM_NAME) \
		--localize-symbol=$(SYM_NAME_END) \
		$< $@
	@echo "Success! Created $(OUTPUT_OBJ)"

# ==========================================
# Utilities
# ==========================================

# Helper to prove the symbols are local
check: $(OUTPUT_OBJ)
	@echo "Checking symbol visibility in $(OUTPUT_OBJ)..."
	@echo "Look for lowercase 'd' (local data) or 'r' (local read-only):"
	@nm $(OUTPUT_OBJ) | grep __tvm_ffi__

clean:
	rm -f *.o
```

examples/cubin_launcher/example_triton_cubin.py
```diff
@@ -0,0 +1,221 @@
...
+        print("PTX Source:")
+        print("=" * 60)
+        print(ptx_source)
+        print("=" * 60 + "\n")
+    except Exception as e:
+        print(f"[ERROR] Failed to compile Triton kernel: {e}")
+        traceback.print_exc()
+        return 2
+
+    # Write CUBIN to file
```
tqchen: would be nice to directly do tvm_ffi.cpp.load_inline(cpp_source, embed_cubin={"env": triton_cubin});

tqchen (changes requested):

tqchen:
API-wise, i think we can streamline it a bit further

```python
cpp_source = """
#include <tvm/ffi/extra/cuda/cubin_launcher.h>

TVM_FFI_EMBED_CUBIN(env);

void AddTwo(TensorView a, TensorView b) {
   static ffi::CubinKernel kernel = TVM_FFI_EMBED_CUBIN_GET_KERNEL(env, "add_two");
   kernel.launch(...);
}
"""

cubin : bytes = compile_cubin_from_nvrtc(cuda_source);

cubin : bytes = compile_cubin_from_triton();
tvm_ffi.cpp.load_inline(cpp_source, embed_cubin={"env": cubin});
```

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,306 @@
...
+ * \file tvm/ffi/extra/cubin_launcher.h
+ * \brief CUDA CUBIN launcher utility for loading and executing CUDA kernels.
+ *
+ * This header provides a lightweight C++ wrapper around CUDA Driver API
+ * for loading CUBIN modules and launching kernels. It supports:
+ * - Loading CUBIN from memory (embedded data) or files
+ * - Multi-GPU execution using CUDA primary contexts
+ * - Kernel parameter management and launch configuration
+ */
+#ifndef TVM_FFI_EXTRA_CUBIN_LAUNCHER_H_
```
junrushao, line 29: I'm not entirely sure if it may look like we are introducing extra cuda dependency (which we don't) by adding this extra file into `include/` directory. In fact, there are only a few driver APIs it is using:

- cuInit
- cuGetErrorName
- cuGetErrorString
- cuLibraryLoadData
- cuLibraryUnload
- cuLibraryGetKernel
- cuKernelGetFunction
- cuLaunchKernel

Is it a good way if we could potentially not have to expose this header? e.g. hiding this in `example/`, etc?

tqchen: this is a header only thing, so users who do not use it likely don't have such dep, and users who include it will have cuda dep. i think we can try to keep it in tvm/ffi/extra/cuda/cubin_launcher.h

tqchen: we can also document clearly in the file that it is header only and include this file will introduce cuda dep

junrushao: having a separate `tvm/ffi/extra/cuda/` header-only directory is likely good enough

vinx13:
The last two args of triton kernel are scratch memory, it need to be allocated using the size in metadata. See https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/driver.py#L700-L714
you will also need to handle constexpr args that are not in the cubin args

yaoyaoding:
> The last two args of triton kernel are scratch memory, it need to be allocated using the size in metadata. See https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/driver.py#L700-L714 you will also need to handle constexpr args that are not in the cubin args

Cool, thanks for the information!

Ubospica (commented):
Thanks for the PR! This is very helpful for further integrating Triton kernels.

Shall we consider adding cubin to hash_sources at
https://github.com/apache/tvm-ffi/blob/8dbd28112cddbaeeffa26420b1dee98a582a7cd7/python/tvm_ffi/cpp/extension.py#L38 ? This could avoid incorrect cache reuse for the same source code but different cubin content.

include/tvm/ffi/extra/cubin_launcher.h
yaoyaoding: I checked the behavior in driver documentation, and find that it's not documented. I asked Gemini and it told me that it's safe to free the code bytes after the `cuLibraryLoadData` call. I will do some small experiment to validate.

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: done

include/tvm/ffi/extra/cubin_launcher.h
yaoyaoding: done

examples/cubin_launcher/src/lib_embedded.cc
yaoyaoding: great! done.

examples/cubin_launcher/src/lib_embedded.cc
yaoyaoding: The claude sonnet gives a slightly different steps. See the documentation of the macro `TVM_FFI_EMBED_CUBIN`

examples/cubin_launcher/example_triton_cubin.py
yaoyaoding: sounds good! Added.

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: moved to `tvm/ffi/extra/cuda`.

python/tvm_ffi/cpp/nvrtc.py
```diff
@@ -0,0 +1,296 @@
...
+
+from __future__ import annotations
+
+import subprocess
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Sequence
+
+
+def nvrtc_compile(  # noqa: PLR0912, PLR0915
```
tqchen, line 24: not have to be today, would be nice to support fatbin bundles through multiple nvrtc compile and them bundle

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -93,56 +90,140 @@ struct dim3 {
...
   /*!
    * \brief Load CUBIN module from memory.
    *
-   * \param data Pointer to CUBIN binary data in memory.
-   * \param size Size of the CUBIN binary data in bytes.
+   * \param bytes CUBIN binary data as a Bytes object.
    * \note Calls cuInit(0) to ensure CUDA is initialized.
    */
-  CubinModule(const void* data, uint64_t size) {
+  explicit CubinModule(const Bytes& bytes) {
```
tqchen, line 311: actually not explicitly creating Bytes is better, in the case of large view, maybe we can start by changing to `const char* data_begin, const char*data_end`

examples/cubin_launcher/src/lib_dynamic.cc
```diff
@@ -27,27 +27,42 @@
...
  * \brief Load CUBIN from file path.
  * \param path Path to the CUBIN file.
  */
 void LoadCubin(const tvm::ffi::String& path) {
-  // Create new module and kernels
-  g_cubin_module = std::make_unique<tvm::ffi::CubinModule>(path.c_str());
+  // Read file into memory
+  std::ifstream file(std::string(path), std::ios::binary | std::ios::ate);
+  TVM_FFI_CHECK(file.is_open(), RuntimeError) << "Failed to open CUBIN file: " << path;
+
```
tqchen: maybe we can remove lib_dynamic assume we have robust embedding path

tqchen (changes requested):

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: > in the case of large view

The underlying driver expects a null-terminated byte stream. Maybe we just accept a "const char* data"? this will hint that it must be null-terminated.

python/tvm_ffi/cpp/nvrtc.py
yaoyaoding: a quick survey find that:
- nvFatbin is used to manage cuda fatbin. (https://docs.nvidia.com/cuda/nvfatbin/index.html)
- the fatbin related API (nvFatbin) is not exposed in cuda-python (https://nvidia.github.io/cuda-python/cuda-bindings/latest/)

include/tvm/ffi/extra/cuda/cubin_launcher.h
tqchen: i think for binary that is dangerous, let us do begin and end

docs/guides/cubin_launcher.md
```diff
@@ -0,0 +1,471 @@
...
+- The `embed_cubin` parameter is a dictionary mapping CUBIN names to their binary data
+- CUBIN names in `embed_cubin` must match names in {c:macro}`TVM_FFI_EMBED_CUBIN`
+- Use `cuda_sources` parameter (instead of `cpp_sources`) to automatically link with CUDA libraries
+- The C++ wrapper handles device management, stream handling, and kernel launching
+
+### Example: Using Triton Kernels
+
+You can compile Triton kernels to CUBIN and launch them through TVM-FFI:
+
+```python
```
tqchen: i remember @junrushao mentioned there is a way for us to refer to existing file in the repo from rst so we don't have to copy code, checkout quick start

tqchen: cc @vinx13

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: I feel it's better to keep the current tvm::ffi::Bytes since it's also very lightlweight. Is the overhead to passing a bytes from python to c++ large?

docs/guides/cubin_launcher.md
yaoyaoding: Cool, I've updated to use the technique!

examples/cubin_launcher/src/lib_embedded.cc
yaoyaoding: I've added a python script to embed cubin to an object file:
```
usage: python -m tvm_ffi.utils.embed_cubin [-h] --output-obj PATH --input-obj PATH --cubin PATH --name NAME [-v]

Embed CUBIN data into existing object files that use TVM_FFI_EMBED_CUBIN macro

options:
  -h, --help         show this help message and exit
  --output-obj PATH  Output object file path (e.g., new.o)
  --input-obj PATH   Input object file path containing TVM_FFI_EMBED_CUBIN usage (e.g., old.o)
  --cubin PATH       Input CUBIN file path (e.g., kernel.cubin)
  --name NAME        Name used in TVM_FFI_EMBED_CUBIN macro (e.g., my_kernels)
  -v, --verbose      Print detailed command output

Examples:
  # Basic usage
  python -m tvm_ffi.utils.embed_cubin \
      --output-obj new.o \
      --input-obj old.o \
      --cubin kernel.cubin \
      --name my_kernels

  # With verbose output
  python -m tvm_ffi.utils.embed_cubin \
      --output-obj new.o \
      --input-obj old.o \
      --cubin kernel.cubin \
      --name my_kernels \
      --verbose

Workflow:
  1. Compile C++ code that uses TVM_FFI_EMBED_CUBIN to create old.o
  2. Compile CUDA kernel to CUBIN (e.g., using nvcc or NVRTC)
  3. Use this tool to merge them into new.o
  4. Link new.o into your final shared library

Usage in C++ code (source compiled to old.o):
  TVM_FFI_EMBED_CUBIN(my_kernels);
  auto kernel = TVM_FFI_EMBED_CUBIN_GET_KERNEL(my_kernels, "kernel_name");

Requirements:
  - GNU binutils (ld and objcopy) must be available in PATH
  - Linux/Unix platform (Windows uses different embedding mechanisms)
```

For cmake, we have two utility functions at `cmake/Utils/EmbedCubin.cmake`
```cmake
tvm_ffi_generate_cubin(
  OUTPUT <output_cubin_file>
  SOURCE <cuda_source_file>
  [ARCH <architecture>]
  [OPTIONS <extra_nvcc_options>...]
  [DEPENDS <additional_dependencies>...]
)
tvm_ffi_embed_cubin(
  OUTPUT <output_object_file>
  SOURCE <source_file>
  CUBIN <cubin_file>
  NAME <symbol_name>
  [DEPENDS <additional_dependencies>...]
)
```

yaoyaoding:
Hi @tqchen @junrushao @Ubospica , the PR is ready for review, could you have one pass?

python/tvm_ffi/cpp/nvrtc.py
yaoyaoding: Will leave to a future PR if needed.

include/tvm/ffi/extra/cuda/cubin_launcher.h
tqchen: Agree that we can keep tvm::ffi::Bytes constructor along with the non-owning constructor `(begin, end)` and `const TVMFFIByteArray* data`

Main consideration here is the overload of loading from embedded symbol when cubin is large, so we don't need to first do an extra copy

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,503 @@
...
+ * a singleton struct to manage the CubinModule instance. The CUBIN data
+ * symbols should be named `__tvm_ffi__cubin_<name>` and `__tvm_ffi__cubin_<name>_end`,
+ * typically created using objcopy and ld.
+ *
+ * \par Creating Embedded CUBIN Symbols
+ * To embed a CUBIN file into your binary, follow these steps:
+ *
+ * \par Step 1: Compile CUDA kernel to CUBIN
+ * \code{.bash}
+ * nvcc --cubin -arch=sm_75 kernel.cu -o kernel.cubin
```
tqchen: consider now refer to our python util and cmake util instead, and add explanations in these places

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,503 @@
...
+ */
+class CubinModule {
+ public:
+  /*!
+   * \brief Load CUBIN module from memory.
+   *
+   * \param bytes CUBIN binary data as a Bytes object.
+   * \note Calls cuInit(0) to ensure CUDA is initialized.
+   */
+  explicit CubinModule(const Bytes& bytes) {
```
tqchen, line 311: also add alternative constructor `CubinModule(const char* data_begin, const char* data_end);` for code that embeds in the module to avoid extra memcpy, in case cubin is too big

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,503 @@
...
+    TVM_FFI_CHECK_CUDA_DRIVER_ERROR(cuInit(0));
+    TVM_FFI_CHECK_CUDA_DRIVER_ERROR(
+        cuLibraryLoadData(&library_, bytes.data(), nullptr, nullptr, 0, nullptr, nullptr, 0));
```
tqchen: Looking again, seems now it is possible to do dynamic loading and execution fully from the cuda runtime API, this might simplify the linking process(as we only need to link against cudart) and user expectations so we don't have to work on driver api and driver init likely

Checkout functions

- cudaLibraryLoadData
- cudaLibraryGetKernel
- cudaLaunchKernel. (no need to get function, directly launch kernel)
- cudaLibraryUnload

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,503 @@
...
+    // Get function handle for the current context from the kernel
+    CUresult result = cuKernelGetFunction(&function, kernel_);
+    if (result != CUDA_SUCCESS) {
+      return result;
+    }
+
+    return cuLaunchKernel(function, grid.x, grid.y, grid.z, block.x, block.y, block.z,
+                          dyn_smem_bytes, stream, args, nullptr);
+  }
+
```
tqchen, line 495: add a helper function kernel.
```c++
kernel.adjustDynamicSharedMemorySize(size_t static_smem_bytes = 0) {
   size_t dynamic_max = device_attr_query_via(cudaDevAttrMaxSharedMemoryPerBlockOptin);
   size_t dynamic_limit = dynamic_max - static_smem_bytes;
   for (int i = 0; i < device_count; ++i) {
       cudaKernelSetAttributeForDevice(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, dynamic_limit);
   }
}

// in cubin module, allow fetch kernel as follows:
static kernel = cubin_mod.GetKernelWithMaxDynamicSharedMemory(name, static_smem_bytes);
```
This is typically needed in high-perf gemm kernels since default dynamic smem setting is not as large

python/tvm_ffi/cpp/extension.py
```diff
@@ -255,7 +269,8 @@ def _generate_ninja_build(  # noqa: PLR0915
             default_cuda_cflags += [_get_cuda_target()]
             default_ldflags += [
                 "-L{}".format(str(Path(_find_cuda_home()) / "lib64")),
-                "-lcudart",
+                "-lcudart",  # cuda runtime library
+                "-lcuda",  # cuda driver library
```
tqchen: assuming we switch to cuda runtime, likely we don't need lcuda anymore

tqchen (commented):
looking great, some final comments

python/tvm_ffi/cpp/extension.py
yaoyaoding: fixed

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: Added the code during kernel loading.

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: Cool! Updated.

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: Added!

include/tvm/ffi/extra/cuda/cubin_launcher.h
yaoyaoding: Updated!

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,562 @@
...
+    // Set max dynamic shared memory for all devices during initialization
+    // This allows the kernel to use maximum available shared memory when needed
+    int device_count = 0;
+    cudaError_t err = cudaGetDeviceCount(&device_count);
+    if (err == cudaSuccess && device_count > 0) {
+      bool any_success = false;
+      for (int device_id = 0; device_id < device_count; ++device_id) {
+        // Query device's maximum shared memory per block
+        cudaDeviceProp prop;
+        err = cudaGetDeviceProperties(&prop, device_id);
```
tqchen: use https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gb22e8256592b836df9a9cc36c9db7151 instead via cudaDevAttrMaxSharedMemoryPerBlockOptin

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,562 @@
...
+   *
+   * \param library The cudaLibrary_t handle.
+   * \param name Name of the kernel function.
+   */
+  CubinKernel(cudaLibrary_t library, const char* name) {
+    TVM_FFI_CHECK_CUDA_ERROR(cudaLibraryGetKernel(&kernel_, library, name));
+
+    // Set max dynamic shared memory for all devices during initialization
+    // This allows the kernel to use maximum available shared memory when needed
+    int device_count = 0;
```
tqchen: move this to a function kernel.SetMaxDynamicSharedMemory(size_t static_mem_size, int64_t dynamic_smem_max=-1); consider make private and friend to CubinModule

where -1 deduce max from max value mininus static_mem_size

tqchen: Add CubinModule.GetKernelWithMaxDynamicSharedMemory(name, static_mem_size, dynamic_smem_max);

This is advanced mode since not all kernels need it

docs/guides/cubin_launcher.rst
```diff
@@ -0,0 +1,422 @@
...
+   objcopy --rename-section .data=.rodata,alloc,load,readonly,data,contents \
+           --redefine-sym _binary_kernel_cubin_start=__tvm_ffi__cubin_my_kernels \
+           --redefine-sym _binary_kernel_cubin_end=__tvm_ffi__cubin_my_kernels_end \
+           kernel_data.o
+
+**Step 4: Link with your library**
+
+.. code-block:: bash
+
+   g++ -o mylib.so -shared mycode.cc kernel_data.o -Wl,-z,noexecstack -lcuda
```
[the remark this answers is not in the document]
yaoyaoding: good catch

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,623 @@
...
+  }
+
+  /*!
+   * \brief Load CUBIN module from raw memory buffer.
+   *
+   * \param code Pointer to CUBIN binary data.
+   * \note CUDA Runtime API automatically initializes on first use.
+   * \note This constructor is primarily used by TVM_FFI_EMBED_CUBIN macro.
+   * \note The code buffer must be null-terminated; size parameter is not required
+   *       as cudaLibraryLoadData can determine the size from the data itself.
```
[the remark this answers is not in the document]
yaoyaoding: fixed!

include/tvm/ffi/extra/cuda/cubin_launcher.h
```diff
@@ -0,0 +1,619 @@
...
+      err = cudaDeviceGetAttribute(&max_shared_mem, cudaDevAttrMaxSharedMemoryPerBlock, device_id);
+      if (err != cudaSuccess) {
+        continue;  // Skip this device if we can't get its attribute
+      }
+
+      int shared_mem_to_set;
+      if (dynamic_smem_max == -1) {
+        // Query the kernel's static shared memory usage
+        cudaFuncAttributes func_attr;
+        err = cudaFuncGetAttributes(&func_attr, reinterpret_cast<const void*>(kernel_));
```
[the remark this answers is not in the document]
yaoyaoding: After looking up the documentation:
- there is no runtime API function named `cudaKernelGetAttributes`
- the documentation says we can use `cudaFuncGetAttributes` to query attribute of cudaKernel_t.

python/tvm_ffi/cpp/extension.py
```diff
@@ -45,35 +46,47 @@ def _hash_sources(
...
+                _hash(item)
+        else:
+            raise ValueError(f"Unsupported type: {type(obj)}")
+
+    _hash(
+        (
+            cpp_source,
+            cuda_source,
+            cpp_files,
+            cuda_files,
```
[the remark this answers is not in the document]
yaoyaoding: fixed

python/tvm_ffi/cpp/extension.py
```diff
@@ -45,35 +46,47 @@ def _hash_sources(
...
+            m.update(b"str")
+            m.update(obj.encode("utf-8"))
+        elif isinstance(obj, bytes):
+            m.update(b"bytes")
+            m.update(obj)
+        elif isinstance(obj, Mapping):
+            m.update(b"Mapping")
+            for key, item in obj.items():
+                _hash(key)
+                _hash(item)
```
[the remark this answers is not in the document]
yaoyaoding: fixed

include/tvm/ffi/extra/cuda/cubin_launcher.h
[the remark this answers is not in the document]
yaoyaoding: added a comment to mark this.

PR #284, opened 2025-11-24
[Fix] Cast Shape from DLTensor*

locnd182644:
# Description

- Type error: Cannot convert from type ' DLTensor* '  to ' ffi.Shape '

# Detail
- onnx  -> lib.so -> Run via RPC -> Happen error

DLTensorConvertShape

- Fixed type error: Cannot convert from type ' DLTensor* '  to ' ffi.Shape '


tqchen:
thanks @locnd182644 . I think in this case instead we should fix to explicitly pass in Shape for this case

locnd182644:
@tqchen Thank you for comment. I resolved this issue in the tvm repository.

Issue #265, opened 2025-11-15
[Feature Request] Function Overloading


Ubospica:
**Goal and motivation.** Supporting function overloading in tvm-ffi. This would be helpful for libraries that use overloading in their code. For example, [XGrammar](https://github.com/mlc-ai/xgrammar) has a lot of overloading functions. Supporting overloading can unblock it from migrating to tvm-ffi.

**Method.** 
1. According to our guideline that the C ABI should be stable and minimal, we should keep the C ABI unchanged.
2. Extend tvm::ffi::Function to implement a chain to maintain all function pointers in the order of registration.
3. When calling functions, try functions in order until the first matched.

This also aligns with nanobind's method to handle overloaded functions. It allocates a contiguous block of memory to store all function overloads. Its memory layout:
```
+------------------+  ← nb_func* (PyObject_VAR_HEAD)
| ob_refcnt        |
| ob_type          |
| ob_size          |  ← number of overloads (Py_SIZE)
+------------------+
| vectorcall       |  ← function pointer: call dispatcher
| max_nargs        |  ← maximum number of arguments among all overloads
| complex_call     |  ← boolean: whether a complex call path is required
| doc_uniform      |  ← boolean: whether all overloads share the same docstring
+------------------+  ← sizeof(nb_func)
| func_data[0]     |  ← metadata for the first overload
|   (overload 0)   |
+------------------+
| func_data[1]     |  ← metadata for the second overload
|   (overload 1)   |
+------------------+
| ...              |
+------------------+
| func_data[N-1]   |  ← metadata for the last overload
|   (overload N-1) |
+------------------+
```

cc @junrushao @tqchen @DarkSharpness

tqchen:
Because the main goal of the design is to keep minimalism, we should not change the implementation of `ffi::Function` ideally, however, this is something that maybe build inside reflection namespace as a implementation of the Function(where callable is backed by something like this), where there is a single entry and dispatch happens under the hood

junrushao:
It's indeed a reasonable ask. I was curious to learn thoughts if it's a necessity though, if it comes at the cost of extra design complexity and runtime overhead at default path, particularly when frontend languages like Python don't really support overloading

tqchen:
I think the main goal is to NOT change the default path. Instead allow overload hidden under an ffi::Function that does dispatch inside

Ubospica:
Thanks @junrushao and @tqchen for the insightful comments!

[...]

PR #286, opened 2025-11-25
[Feature] Support dynamic-style overload for FFI object types

DarkSharpness:
Related discussion here #265 .

Modification in short:

1. Add `Function::FromPackedInplace` in `function.h` and generalize some methods in `registry.h`
2. Implement FFI object types overload method (all in `extra/overload.h`)
3. Add a simple test.

### DIFF
--- a/include/tvm/ffi/reflection/overload.h
+++ b/include/tvm/ffi/reflection/overload.h
@@ -17,7 +17,7 @@
  * under the License.
  */
 /*!
- * \file tvm/ffi/extra/overload.h
+ * \file tvm/ffi/reflection/overload.h
  * \brief Registry of reflection metadata, supporting function overloading.
  */
 #ifndef TVM_FFI_EXTRA_OVERLOAD_H
--- a/tests/cpp/test_overload.cc
+++ b/tests/cpp/test_overload.cc
@@ -18,11 +18,11 @@
  */
 #include <gtest/gtest.h>
 #include <tvm/ffi/container/map.h>
-#include <tvm/ffi/extra/overload.h>
 #include <tvm/ffi/object.h>
 #include <tvm/ffi/reflection/access_path.h>
 #include <tvm/ffi/reflection/accessor.h>
 #include <tvm/ffi/reflection/creator.h>
+#include <tvm/ffi/reflection/overload.h>
 #include <tvm/ffi/reflection/registry.h>
 #include <tvm/ffi/string.h>
 
@@ -33,11 +33,12 @@ using namespace tvm::ffi;
 struct TestOverloadObj : public Object {
   explicit TestOverloadObj(int32_t x) : type(Type::INT) {}
   explicit TestOverloadObj(float y) : type(Type::FLOAT) {}
-  static int add_one_int(int x) { return x + 1; }
-  static float add_one_float(float x) { return x + 1.0f; }
+
+  static int AddOneInt(int x) { return x + 1; }
+  static float AddOneFloat(float x) { return x + 1.0f; }
 
   template <typename T>
-  auto holds(T) {
+  auto Holds(T) const {
     if constexpr (std::is_same_v<T, int32_t>) {
       return type == Type::INT;
     } else if constexpr (std::is_same_v<T, float>) {
@@ -56,10 +57,10 @@ TVM_FFI_STATIC_INIT_BLOCK() {
   refl::OverloadObjectDef<TestOverloadObj>()
       .def(refl::init<int32_t>())
       .def(refl::init<float>())
-      .def("hold_same_type", &TestOverloadObj::holds<int32_t>)
-      .def("hold_same_type", &TestOverloadObj::holds<float>)
-      .def_static("add_one_static", &TestOverloadObj::add_one_int)
-      .def_static("add_one_static", &TestOverloadObj::add_one_float);
+      .def("hold_same_type", &TestOverloadObj::Holds<int32_t>)
+      .def("hold_same_type", &TestOverloadObj::Holds<float>)
+      .def_static("add_one_static", &TestOverloadObj::AddOneInt)
+      .def_static("add_one_static", &TestOverloadObj::AddOneFloat);
 }
 
 TEST(Reflection, CallOverloadedInitMethod) {


include/tvm/ffi/function.h
```diff
@@ -33,15 +33,20 @@
 #define TVM_FFI_DLL_EXPORT_INCLUDE_METADATA 0
 #endif
 
+#if TVM_FFI_DLL_EXPORT_INCLUDE_METADATA
```
junrushao, line 36: I suppose it's fine to wildcard include `sstream` without the guards?

DarkSharpness: Wildcard include sstream is not friendly to clangd, which gives me an annoying warning "Included header sstream is not used directly". It's just a style fix and could be removed.

junrushao: I see.

I’ve got bunch of similar warnings too. I managed to run clangd on this repo on mainline CI, but it only applies to .cc files instead of headers. If you have any ideas how to improve, I’m happy to walk you through what the existing infra looks like.

junrushao:
Ping @Ubospica @cyx-6 if you guys would love to help review this PR!

include/tvm/ffi/function.h
```diff
@@ -166,21 +171,19 @@ class FunctionObjImpl : public FunctionObj {
...
+  template <typename... Args>
+  explicit FunctionObjImpl(Args&&... args) : callable_(std::forward<Args>(args)...) {
     this->safe_call = SafeCall;
     this->cpp_call = reinterpret_cast<void*>(CppCall);
   }
 
+  FunctionObjImpl(const FunctionObjImpl&) = delete;
+  FunctionObjImpl& operator=(const FunctionObjImpl&) = delete;
+
+  TCallable& GetCallable() { return callable_; }
```
junrushao: nit: For some reason our convention is to always return pointers instead of references, so it would be better if we do

```suggestion
  TCallable* GetCallable() { return &callable_; }
```

include/tvm/ffi/reflection/registry.h
```diff
@@ -270,52 +274,49 @@ class ReflectionDefBase {
...
+    return ffi::Function::FromTyped(WrapFunction(std::forward<Func>(func)), std::move(name));
+  }
+
+  template <typename Func>
+  TVM_FFI_INLINE static Func&& WrapFunction(Func&& func) {
+    return std::forward<Func>(func);
+  }
   template <typename Class, typename R, typename... Args>
-  TVM_FFI_INLINE static Function GetMethod(std::string name, R (Class::*func)(Args...)) {
+  TVM_FFI_INLINE static auto WrapFunction(R (Class::*func)(Args...)) {
```
junrushao, line 287: Curious what the return type is for small-brain human beings like me :)

tests/cpp/test_overload.cc
```diff
@@ -0,0 +1,95 @@
+
```
junrushao: nit: remove blank line

junrushao (commented):
I don't have sufficient evidence to form an opinion on the design below (copied from unittests), but indeed it's super interesting and creative.

```C++
struct TestOverloadObj : public Object {
  explicit TestOverloadObj(int32_t x) : type(Type::INT) {}
  explicit TestOverloadObj(float y) : type(Type::FLOAT) {}
  static int add_one_int(int x) { return x + 1; }
  static float add_one_float(float x) { return x + 1.0f; }

  template <typename T>
  auto holds(T) {
    if constexpr (std::is_same_v<T, int32_t>) {
      return type == Type::INT;
    } else if constexpr (std::is_same_v<T, float>) {
      return type == Type::FLOAT;
    } else {
      static_assert(sizeof(T) == 0, "Unsupported type");
    }
  }

  enum class Type { INT, FLOAT } type;
  TVM_FFI_DECLARE_OBJECT_INFO("test.TestOverloadObj", TestOverloadObj, Object);
};

TVM_FFI_STATIC_INIT_BLOCK() {
  namespace refl = tvm::ffi::reflection;
  refl::OverloadObjectDef<TestOverloadObj>()
      .def(refl::init<int32_t>())
      .def(refl::init<float>())
      .def("hold_same_type", &TestOverloadObj::holds<int32_t>)
      .def("hold_same_type", &TestOverloadObj::holds<float>)
      .def_static("add_one_static", &TestOverloadObj::add_one_int)
      .def_static("add_one_static", &TestOverloadObj::add_one_float);
}
```

include/tvm/ffi/reflection/registry.h
DarkSharpness: The return type here is a lambda function (so we can only return by `auto`). We just do some generalization here for code reuse in `OverloadObjectDef`.

DarkSharpness:
cc @junrushao .
Now got following error message for mismatched arg type:

```
C++ exception with description "No matching overload found when calling: `test.TestOverloadObj.hold_same_type` with 2 arguments:
Overload #0: Mismatched type on argument #1 when calling: `test.TestOverloadObj.hold_same_type (0: test.TestOverloadObj, 1: int) -> bool`. Expected `int` but got `const char*`
Overload #1: Mismatched type on argument #1 when calling: `test.TestOverloadObj.hold_same_type (0: test.TestOverloadObj, 1: float) -> bool`. Expected `float` but got `const char*`
```

for mismatched arg number:

```
C++ exception with description "No matching overload found when calling: `test.TestOverloadObj.hold_same_type` with 1 arguments:
Overload #0: Mismatched number of arguments when calling: `test.TestOverloadObj.hold_same_type (0: test.TestOverloadObj, 1: int) -> bool`. Expected 2 arguments
Overload #1: Mismatched number of arguments when calling: `test.TestOverloadObj.hold_same_type (0: test.TestOverloadObj, 1: float) -> bool`. Expected 2 arguments
```

include/tvm/ffi/extra/overload.h
```diff
@@ -0,0 +1,501 @@
...
+    for (const auto& [overload, fptr] : overloads_) {
+      if (overload->num_args_ != num_args) continue;
+      if (fptr(overload.get(), args, num_args, rv)) return;
+    }
+
+    this->handle_overload_failure(args, num_args);
+  }
+
+ private:
+  void handle_overload_failure(const AnyView* args, int32_t num_args) {
```
tqchen: style, the codebase use CamelSyle per google C style

include/tvm/ffi/reflection/overload.h
```diff
@@ -0,0 +1,501 @@
...
+   * \tparam Extra The extra arguments.
+   *
+   * \param name The name of the method.
+   * \param func The function to be registered.
+   * \param extra The extra arguments that can be docstring.
+   *
+   * \return The reflection definition.
+   */
+  template <typename Func, typename... Extra>
+  TVM_FFI_INLINE OverloadObjectDef& def_static(const char* name, Func&& func, Extra&&... extra) {
```
tqchen, line 359: didn't realize it is soo tied to reflection, in this case,, perhaps it should go into reflection

DarkSharpness: agree. moved

Ubospica (approved):
Thanks, the changes LGTM! Supporting function overloading is be very helpful.

Based on our offline discussions regarding the design, this implementation looks solid. I especially like how we designed these features:
1. The `refl::OverloadObjectDef` is optional and independent of `ObjectDef`.
2. The code path for the first overload is aligned to the non-overload version to minimize overhead
3. Although we need to check arguments at runtime, the function signatures are known at compile time, and we are using templates to generate the checking code for certain function signatures. This reduces the runtime overhead.

One thing to note is that we are doing parameter type checking & conversion at the same time. If the conversion is slow (such as `Array<Array<Type>>`), we need to check and convert all array elements), this may introduce some overhead. But I think it's fine since these slow paths are currently relatively uncommon.

cc @tqchen @junrushao

Ubospica:
@junrushao Thanks so much for merging this PR!

PR #293, opened 2025-11-30
fix: Use-after-move when handling std::tuple

junrushao:
See: https://github.com/apache/tvm-ffi/pull/228#discussion_r2573338636


junrushao:
CC @tqchen @DarkSharpness

tqchen (approved):

junrushao:
Looks like this PR doesn't actually fix segfaults. See https://github.com/apache/tvm-ffi/actions/runs/19803545351/job/56734651884#step:11:288 which is run on commit https://github.com/apache/tvm-ffi/commits/f5adc30aa2f9520fd32b3f345d85e72ebe9b86bf/

PR #298, opened 2025-12-01
[TEST] Further fix potential use after free issue

tqchen:
This PR further fixes the potential use after free issue in the test


tqchen:
cc @junrushao @DarkSharpness

junrushao:
Thank you! Does this mean if we enclose every temp variable inside a function, they will be eagerly deallocated upon function exit?

junrushao (approved):

tqchen:
right, temp variables in function will be deallocated at function exit

PR #300, opened 2025-12-02
[Enhancement] Refactor cubin launcher

### LINKED ISSUE #292: [Proposal] Refactor cubin launcher
tvm-ffi just introduced [cubin launcher](https://github.com/apache/tvm-ffi/pull/283) to allow embed one cuda/cubin file into c++ source code, and call kernel via ffi. This feature allows a more flexible integration of cuda kernel. (mainly for jit compile, if I understand correctly)

However, while investigating the feature, we've found it not that close to the simplest cuda-development workflow (which requires more efforts while migrating from existing kernels), it also doesn't reuse the existing CUDA utils in cmake. Also, it's rely on cuda runtime library and sometimes this introduces unexpedted constraints to CUDA runtime version.

We've implemented a POC to resolve such quirks (https://github.com/oraluben/kernel-lib-template/commit/e973a2e256f46fb08d108d520e9797499d9c45b8) and would like to propose some changes to the current cubin launcher APIs, and would like to discuss with the team before coding.

Those changes might include:
1. cmake utils to allow user to generate header from cuda/cubin files;
2. tvm-ffi macros to load libraries from generated header;
3. switch between cudart api and cuda driver api, and necessary checks to detect major version mismatch;
4. other minor utils to support features above.

oraluben:
Resolves #292 

## Summary
This PR refactors the CUBIN launcher utilities to improve flexibility, modernization, and CMake integration. It introduces new macros for loading CUBINs from byte arrays (enabling C++23 `#embed` support or `bin2c` in CUDA Toolkit) and provides updated CMake utilities that better align with standard CMake practices.

### C++ API & Headers
- **New Macro**: `TVM_FFI_LOAD_LIBRARY_FROM_BYTES(name, imageBytes)`: Allows registering a CUBIN module from a raw byte array (e.g., `unsigned char[]`). This supports modern embedding techniques like C++23 `#embed` (or compiler extensions) and traditional header-based embedding (e.g. via `bin2c`).
- **Header Refactoring**:
    - `include/tvm/ffi/extra/cuda/unify_api.h`: specific low-level CUDA Driver/Runtime API wrappers moved here. This header introduces a unified abstraction layer over CUDA Driver and Runtime APIs. It provides:
        - **Unified Handles**: `LibraryHandle`, `KernelHandle`, `StreamHandle`, `ResultHandle`, etc. that map to their respective Driver (e.g., `CUfunction`) or Runtime (e.g., `cudaKernel_t`) counterparts.
        - **Unified Error Checking**: `TVM_FFI_CHECK_CUDA_ERROR` macro that handles return codes from either API.
        - **Seamless Switching**: Users can select the underlying API by defining `TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API` (defined true value for Driver, otherwise for Runtime) before inclusion, with automatic default selection based on CUDA version.
    - `include/tvm/ffi/extra/cuda/base.h`: `dim3` struct definition moved here.
    - `include/tvm/ffi/extra/cuda/cubin_launcher.h`: simplified to focus on `CubinModule` and `CubinKernel`.
- **API Update**: `CubinModule` constructor now accepts `const unsigned char*` to support byte arrays.

### CMake Utilities (`cmake/Utils/EmbedCubin.cmake`)
- **Updated Approach**: Transitioned from imperative command-based utilities (old `tvm_ffi_generate_cubin`/`tvm_ffi_embed_cubin`) to a declarative, target-based approach that is more idiomatic to modern CMake.
- **New Functions**:
    - `add_tvm_ffi_cubin`: Compatibility wrapper for compiling CUDA to CUBIN.
    - `add_tvm_ffi_fatbin`: Compatibility wrapper for compiling CUDA to FATBIN.
    - `tvm_ffi_embed_bin_into`: Helper to embed CUBIN/FATBIN into an existing object target (linking approach). It leverages `PRE_LINK` commands to inject binary data into object files just before linking. While using `PRE_LINK` to modify object files is non-trivial, it proves to be the most reliable method for deep integration with CMake's dependency graph and target system.
- **Removed**: Obsolete functions `tvm_ffi_generate_cubin` and `tvm_ffi_embed_cubin`.

### Examples
- **Restructured**: `examples/cubin_launcher/embedded_cubin/` is now split into sub-examples demonstrating different embedding techniques:
    - `embed_with_tvm_ffi`: Uses the CMake/object-linking approach with `TVM_FFI_EMBED_CUBIN`.
    - `cpp_embed`: Uses `#embed` (with C++ 23) with `TVM_FFI_LOAD_LIBRARY_FROM_BYTES`.
    - `include_bin2c`: Shows embedding via `bin2c` style (header file).

### Documentation
- Updated `docs/guides/cubin_launcher.rst` to document the new CMake functions, the `TVM_FFI_LOAD_LIBRARY_FROM_BYTES` macro, and the recommended embedding approaches.

## Test status

I run cubin launcher examples (2 migrated, 2 new) on a CUDA 13.1 env. For the `#embed` one, use clang-20 as host compiler.

I've confirmed that both Runtime and Driver API work, correct symbols from different APIs are used and kernel works as expected.

Notably, when compiling with cudart 13.1, it works with cudart 12.8 comes with torch.

### DIFF
--- a/docs/guides/cubin_launcher.rst
+++ b/docs/guides/cubin_launcher.rst
@@ -41,6 +41,19 @@ TVM-FFI provides utilities for loading and launching CUDA kernels from CUBIN mod
 
 By default, the implementation uses the Runtime API if compiled with CUDA >= 12.8, falling back to the Driver API for older versions. You can force the usage of the Driver API (or Runtime API) by defining the macro ``TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API`` (set to ``1`` for Driver API, ``0`` for Runtime API) before including the header.
 
+.. warning::
+
+   **CMAKE_CUDA_RUNTIME_LIBRARY and Driver API**
+
+   When using CMake, the default behavior (if ``CMAKE_CUDA_RUNTIME_LIBRARY`` is not set) is to link against the CUDA Runtime Library (``cudart``). TVM-FFI's CMake utility automatically defaults this variable to ``Shared`` if it is undefined. This introduces a dependency on the CUDA runtime version, requiring the system's driver to be compatible with that runtime version.
+
+   If you intend to use the Driver API only (e.g. by setting ``TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API=1``) to avoid this runtime dependency:
+
+   1. You must explicitly set ``CMAKE_CUDA_RUNTIME_LIBRARY`` to ``None`` in your CMake configuration to prevent linking ``cudart``.
+   2. You must manually link your target against the CUDA Driver library (usually ``cuda`` on Linux/Windows or `CUDA::cuda_driver` provided by CMake's ``FindCUDAToolkit``).
+
+   This ensures your application relies solely on the widely compatible CUDA Driver API (``libcuda.so.1``).
+
 The implementation is in ``tvm/ffi/extra/cuda/cubin_launcher.h`` and provides:
 
 - :cpp:class:`tvm::ffi::CubinModule`: RAII wrapper for loading CUBIN modules from memory
@@ -175,7 +188,7 @@ Using C++23 ``#embed`` (or compiler extensions like ``#embed`` in Clang/GCC) all
 - Kernel arguments must be pointers to the actual values (use ``&`` for addresses)
 - :cpp:type:`tvm::ffi::dim3` supports 1D, 2D, or 3D configurations: ``dim3(x)``, ``dim3(x, y)``, ``dim3(x, y, z)``
 - ``TVMFFIEnvGetStream`` retrieves the correct CUDA stream for the device
-- Always check kernel launch results with :c:macro:`TVM_FFI_CHECK_CUDA_ERROR` (which checks CUDA Runtime API errors)
+- Always check kernel launch results with :c:macro:`TVM_FFI_CHECK_CUDA_ERROR` (which checks CUDA Runtime API or Driver API errors depending on configuration)
 
 Loading CUBIN at Runtime
 ~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -223,6 +236,7 @@ TVM-FFI provides CMake utility functions that simplify the CUBIN embedding proce
    This prevents static linking of cudart, which requires an exact driver version match.
    If you intend to use the Driver API only (e.g., via ``TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API=1``),
    you should explicitly set ``CMAKE_CUDA_RUNTIME_LIBRARY`` to ``None`` in your CMake configuration before including this utility to avoid linking against the CUDA runtime library.
+   And link with CUDA Driver API.
 
 
 Embedding CUBIN with Python Utility
@@ -435,7 +449,7 @@ Python Utilities
 - ``python -m tvm_ffi.utils.embed_cubin``: Command-line utility to embed CUBIN into object files
 
   - ``--output-obj PATH``: Output combined object file path
-  - ``--input-obj PATH``: Input object file containing C++ code with ``TVM_FFI_EMBED_CUBIN``
+  - ``--input-obj PATH``: Input compiled object file containing C++ code with ``TVM_FFI_EMBED_CUBIN``
   - ``--cubin PATH``: Input CUBIN file to embed
   - ``--name NAME``: Symbol name matching ``TVM_FFI_EMBED_CUBIN(name)`` macro
   - ``--verbose``: Print detailed command output (optional)


oraluben:
This is ready for review @junrushao @yaoyaoding

cmake/Utils/EmbedCubin.cmake
```diff
@@ -15,205 +15,163 @@
 # specific language governing permissions and limitations
 # under the License.
 
+# Do not let cmake to link cudart.
+set(CMAKE_CUDA_RUNTIME_LIBRARY None)
```
yaoyaoding: consider restore this option after we created the targets in this utility since the downstream projects might configure this option.

cmake/Utils/EmbedCubin.cmake
```diff
@@ -15,205 +15,163 @@
...
+
+# We need this to simulate `CUDA_{CUBIN,FATBIN}_COMPILATION` in `add_tvm_ffi_{cubin,fatbin}`, to
+# copy `a.cu.o` to `a.cubin`/`a.fatbin`.
+set(COPY_SCRIPT "${CMAKE_BINARY_DIR}/cuda_copy_utils.cmake")
```
yaoyaoding: can this inlined script be a standalone script under cmake/Utils?

cmake/Utils/EmbedCubin.cmake
```diff
@@ -15,205 +15,163 @@
...
   add_custom_command(
-    OUTPUT "${INTERMEDIATE_OBJ}"
-    COMMAND ${CMAKE_CXX_COMPILER} -c -fPIC -std=c++17 ${INCLUDE_FLAGS} "${ARG_SOURCE_ABS}" -o
-            "${INTERMEDIATE_OBJ}"
-    DEPENDS "${ARG_SOURCE_ABS}"
-    COMMENT "Compiling ${ARG_SOURCE} to intermediate object file"
-    VERBATIM
+    TARGET ${target_name}
+    PRE_LINK
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different "$<TARGET_OBJECTS:${target_name}>"
```
yaoyaoding, line 131: Can we directly use `$<TARGET_OBJECTS:${target_name}>` in the subsequent `add_custom_command` to replace `${final_output}`? Some comment will be helpful for future developers.

cmake/Utils/EmbedCubin.cmake
```diff
@@ -15,205 +15,163 @@
...
-# The function will:
-#   1. Compile the SOURCE file to an intermediate object file
-#   2. Use the tvm_ffi.utils.embed_cubin Python utility to merge the object file
-#      with the CUBIN data
-#   3. Create symbols: __tvm_ffi__cubin_<NAME> and __tvm_ffi__cubin_<NAME>_end
+# Parameters:
+#   target_name: Name of the object library target
+#   library_name: Name of the kernel library
+#   BIN: CUBIN or FATBIN file
+#   INTERMEDIATE_FILE: Optional, location to copy original object file to.
```
yaoyaoding: what's the purpose of exposing `INTERMEDIATE_FILE` to the user?

cmake/Utils/EmbedCubin.cmake
```diff
@@ -15,205 +15,163 @@
...
+# The reason of this design is to integrate with cmake's workflow.
 #
-# The function will:
-#   1. Compile the SOURCE file to an intermediate object file
-#   2. Use the tvm_ffi.utils.embed_cubin Python utility to merge the object file
-#      with the CUBIN data
-#   3. Create symbols: __tvm_ffi__cubin_<NAME> and __tvm_ffi__cubin_<NAME>_end
+# Parameters:
+#   target_name: Name of the object library target
+#   library_name: Name of the kernel library
```
yaoyaoding:
```suggestion
#   symbol_name: Name of the symbol in TVM_FFI_EMBED_CUBIN macro.
```

yaoyaoding: Alternatively, we can make the signature like
```cmake
tvm_ffi_embed_bin_into(
    <target_name> 
    SYMBOL <symbol_name>
    BIN <cubin_or_fatbin>
)
```

docs/guides/cubin_launcher.rst
```diff
@@ -374,6 +411,7 @@ C++ Macros
 ~~~~~~~~~~
 
 - :c:macro:`TVM_FFI_EMBED_CUBIN`: Declare embedded CUBIN module
+- :c:macro:`TVM_FFI_LOAD_LIBRARY_FROM_BYTES`: Load CUBIN from byte array
```
yaoyaoding: In the global context of tvm-ffi, this name is a little confusing, consider adding `CUBIN` to the name so that others know it's cubin related utility macro.

yaoyaoding: Like `TVM_FFI_EMBED_CUBIN_FROM_BYTES`

examples/cubin_launcher/dynamic_cubin/src/lib_dynamic.cc
```diff
@@ -80,14 +80,19 @@ void AddOne(tvm::ffi::TensorView x, tvm::ffi::TensorView y) {
...
   // Get CUDA stream
   DLDevice device = x.device();
-  cudaStream_t stream =
-      static_cast<cudaStream_t>(TVMFFIEnvGetStream(device.device_type, device.device_id));
+  tvm::ffi::StreamHandle stream =
+      static_cast<tvm::ffi::StreamHandle>(TVMFFIEnvGetStream(device.device_type, device.device_id));
 
   // Launch kernel
-  cudaError_t result = g_add_one_kernel->Launch(args, grid, block, stream);
+  tvm::ffi::ResultHandle result = g_add_one_kernel->Launch(args, grid, block, stream);
```
yaoyaoding: The name `tvm::ffi::ResultHandle` is also too generic. Is it cuda specific? If so, better to add 'CUDA' in the name.

yaoyaoding: And the "Handle" is usually used for "pointer" / resource index.

include/tvm/ffi/extra/cuda/unify_api.h
```diff
@@ -0,0 +1,183 @@
...
+#else
+  return idx;
+#endif
+}
+
+static ResultHandle launch_kernel(KernelHandle kernel, void** args, tvm::ffi::dim3 grid,
+                                  tvm::ffi::dim3 block, StreamHandle stream,
+                                  uint32_t dyn_smem_bytes = 0) {
+#if TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API
+  return cuLaunchKernel(reinterpret_cast<CUfunction>(kernel), grid.x, grid.y, grid.z, block.x,
```
yaoyaoding: We can use [cuKernelGetFunction](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__LIBRARY.html#group__CUDA__LIBRARY_1ge4cf9abafaba338acb977585b0d7374a) to get a `CUfunction` from `CUkernel`. From the documentation, it seems that `CUfunction` is context-specific and `CUkernel` is context-agnostic. Thus, the reinterpret_cast here might not be proper since they are different things.

The `cuLaunchKernel` accepts both `CUkernel` and `CUfunction`, thus the kernel launch can work.

yaoyaoding:
```suggestion
  return cuLaunchKernel(kernel, grid.x, grid.y, grid.z, block.x,
```

include/tvm/ffi/extra/cuda/unify_api.h
```diff
@@ -0,0 +1,183 @@
+/*
```
yaoyaoding: It's not clear to me whether we should unify the cuda runtime api and driver api in a standalone header and expose to the users - people might rely on these headers but it might not be the intention for this PR.

I prefer to merge this file into the cubin_launcher.h, and add `_` as the prefix to the unified functions so that it keep internal and we can refactor the implementation in the future when needed.

@tqchen , how do you think the `unify_api.h`?

oraluben: My intension is to provide a layer that user could use same code (error handling, etc.) to produce different binaries that uses driver API and runtime API. To archive that I think we need to define some of the basic types used, e.g. `CUdevice`/`int`, `CUstream`/`cudaStream_t`.

For those underlying API calls, I agree with that we should make them internal.

cmake/Utils/EmbedCubin.cmake
oraluben: `tvm_ffi_embed_bin_into` injects into the building workflow of target `<target_name>`, overwrites the original object file before linking. `INTERMEDIATE_FILE` is the location where original object file would be copy to. I hope this could help users who want to access the original file.

include/tvm/ffi/extra/cuda/unify_api.h
oraluben: Based on my understanding, there's two categories of CUDA Driver API that could load and launch a cuda kernel independently: 
1. [Module API](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MODULE.html)
2. [Library API](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__LIBRARY.html)

The Library API is almost how cudart do in its library API ([6.35. Interactions with the CUDA Driver API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER.html)): 
> **Interactions between CUfunction and cudaFunction_t**
> 
> The types [CUfunction](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gba6128b948022f495706d93bc2cea9c8) and [cudaFunction_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g6ac3a22cc596d09ac07cdafb8a4638cf) represent the same data type and may be used interchangeably by casting the two types between each other.
> 
> In order to use a [cudaFunction_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g6ac3a22cc596d09ac07cdafb8a4638cf) in a CUDA Driver API function which takes a [CUfunction](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gba6128b948022f495706d93bc2cea9c8), it is necessary to explicitly cast the [cudaFunction_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g6ac3a22cc596d09ac07cdafb8a4638cf) to a [CUfunction](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gba6128b948022f495706d93bc2cea9c8).
> 
> **Interactions between CUkernel and cudaKernel_t**
> 
> The types [CUkernel](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g612028921e5736db673e4307589989ed) and [cudaKernel_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g0b33f204b307b3154aa4f005a3c8a46a) represent the same data type and may be used interchangeably by casting the two types between each other.
> 
> In order to use a [cudaKernel_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g0b33f204b307b3154aa4f005a3c8a46a) in a CUDA Driver API function which takes a [CUkernel](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g612028921e5736db673e4307589989ed), it is necessary to explicitly cast the [cudaKernel_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g0b33f204b307b3154aa4f005a3c8a46a) to a [CUkernel](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g612028921e5736db673e4307589989ed).

Since we were using the [library API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__LIBRARY.html) (starting from CUDA 12.8) here, I think it's safe that we use the driver version (library/kernel) also in the driver-api version.


-----

> the reinterpret_cast here might not be proper since they are different things.

In the [doc of `cuLaunchKernel`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1gb8f3dc3031b40da29d5f9a7139e52e15):
> Note that the API can also be used to launch context-less kernel [CUkernel](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g612028921e5736db673e4307589989ed) by querying the handle using [cuLibraryGetKernel()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__LIBRARY.html#group__CUDA__LIBRARY_1g15336d865f5abd63e3dc6004d5bc037a) and then passing it to the API by casting to [CUfunction](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gba6128b948022f495706d93bc2cea9c8). Here, the context to launch the kernel on will either be taken from the specified stream hStream or the current context in case of NULL stream.

I think this is the designed usage, and the signature of `cuLaunchKernel` only takes `CUfunction`, users should explicit convert `CUkernel` to `CUfunction` to call `cuLaunchKernel.

So I tend to keep this part unchanged unless there's other issue, e.g. performance difference comes from context-awareness.

cmake/Utils/EmbedCubin.cmake
oraluben: I don't think this is feasible with the current design.

First, when `CMAKE_CUDA_RUNTIME_LIBRARY` is not set, the default behavior is to statically link `cudart` (this isn’t clearly stated in the CMake docs). This imposes an extra constraint on the driver version - it requires the exact CUDA driver bundled with that `cudart`, not just a compatible one.

Second, when `CMAKE_CUDA_RUNTIME_LIBRARY` is set, it becomes the default value of `CUDA_RUNTIME_LIBRARY` for all future targets. However, our CMake utilities do not create targets for users — we only modify or use targets that users have already created. (We do define helper targets like `add_tvm_ffi_{cubin,fatbin}`, but those are compatibility utilities for CMake < 3.27; users are still encouraged to create their own CUDA targets and set `CUDA_{CUBIN,FATBIN}_COMPILATION` directly.) Therefore, `CMAKE_CUDA_RUNTIME_LIBRARY` must be configured before any targets are defined, and its value must persist throughout configuration.

Maybe we can explicitly document this behavior and log a warning in CMake to make it clearer?

examples/cubin_launcher/dynamic_cubin/src/lib_dynamic.cc
oraluben: Name updated for now, need more discuss about the high-level design of `unify_api.h`.

cmake/Utils/EmbedCubin.cmake
oraluben: `final_output` has a bad name, it's the copied original object file of that target before linking. it's not `$<TARGET_OBJECTS:${target_name}>`.

cmake/Utils/EmbedCubin.cmake
oraluben: Do you think there's need to expose `INTERMEDIATE_FILE`?

include/tvm/ffi/extra/cuda/unify_api.h
yaoyaoding: I see, then no problem here.

cmake/Utils/EmbedCubin.cmake
yaoyaoding: Thanks for the explanation! I'm curious about the point regarding the driver version constraints:
> This imposes an extra constraint on the driver version - it requires the exact CUDA driver bundled with that cudart, not just a compatible one.

Could you share where you found that static linking requires the exact bundled driver rather than just a compatible one? Some docs I found online: https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html#cuda-11-and-later-defaults-to-minor-version-compatibility

cmake/Utils/EmbedCubin.cmake
yaoyaoding: I feel most users will not need the object file. But it's okay to keep it if you feel it's useful.

include/tvm/ffi/extra/cuda/unify_api.h
yaoyaoding: Great! For the common handles/error code types, we can unified them and use them as the interface of cubin_launcher. For API calls, let's keep them in internal and merge `unify_api.h` into the `cubin_launcher.h` (or add a prefix to it as well like `_unify_api.h`). I feel merging it slightly better.

cmake/Utils/EmbedCubin.cmake
yaoyaoding: Never mind, I understand the issue now.

yaoyaoding: I still think a utility should not change the global configuration like CMAKE_CUDA_RUNTIME_LIBRARY.

Maybe we can add explicit remindar in the documentation of the macro
```
TVM_FFI_CUBIN_LAUNCHER_USE_DRIVER_API
```
to remind the user that the default behavior of cmake and CMAKE_CUDA_RUNTIME_LIBRARY to ask them to set it to None in their cmake when they want to use driver api only, otherwise their compiled binary will have a unnecessary runtime library and thus some dependency on driver version.

oraluben: > I'm curious about the point regarding the driver version constraints:
> 
> > This imposes an extra constraint on the driver version - it requires the exact CUDA driver bundled with that cudart, not just a compatible one.

Have a typo here, it requires **at least** the exact CUDA Driver bundled with that cudart, not all compatible ones.

cmake/Utils/EmbedCubin.cmake
oraluben: Removed to reduce confusion

include/tvm/ffi/extra/cuda/unify_api.h
oraluben: After reviewing the code again, I believe we still need a unified compatibility layer. 

Currently, we have `CUDADeviceGuard`, which works as `aten::DeviceGuard` - this could also be implemented using the driver API. On the other hand, if `CUDADeviceGuard` only works with the runtime API while `cubin_launcher` can switch between runtime/driver APIs, it introduces some inconsistency.

I suggest keeping a single internal compatibility header (perhaps renamed to `_unify_api.h` to reflect its internal use). For APIs like `CUDADeviceGuard` that are beyond the current PR’s scope, we can temporarily mark them as “not implemented” in `_unify_api.h` and throw an error when driver API is requested.

Looking forward to your thoughts.

cc @cyx-6 @junrushao @tqchen

oraluben: > this could also be implemented using the driver API

with `cuDeviceGet` & `cuDevicePrimaryCtxRetain` & `cuCtxSetCurrent`

cmake/Utils/EmbedCubin.cmake
oraluben: I agree with most of your points: we shouldn’t change the default configuration, and we should document the issues around `CMAKE_CUDA_RUNTIME_LIBRARY`.

But I’d like to discuss one additional possibility: if `CMAKE_CUDA_RUNTIME_LIBRARY` is not set, we set it to `shared`. In nearly all cases (maybe all), static linking is not what users actually want. For users unaware of this setting, defaulting to `shared` has no downside and may help in some cases. If the user has already set it, we won’t override their choice. We can also document that if users want to use only the driver API, they should explicitly set it to `None` and link against `libcuda` directly.

yaoyaoding: Sounds good!

include/tvm/ffi/extra/cuda/unify_api.h
yaoyaoding: > I suggest keeping a single internal compatibility header (perhaps renamed to _unify_api.h to reflect its internal use). For APIs like CUDADeviceGuard that are beyond the current PR’s scope, we can temporarily mark them as “not implemented” in _unify_api.h and throw an error when driver API is requested.

Sounds good to me as long as the compatibility layer is not exposed to the user as some public API (at least for now). I asked Gemini on the best parctice of indicating "internal-only" header and it suggests us to use some "detail" subfolder and define the internal functions in "details" namespace. Maybe we can create a "detail" subfolder like `/include/tvm/ffi/extra/cuda/detail` and define the compatibility layer in a "details" namespace.

Another minor suggestion is to avoid 
```c++
#ifdef USE_DRIVERAPI
#define some_func driver_func
#else  
#define some_func runtime_func
#endif
```
instead, using the following way
```c++
ret_type some_func(...) {
""" what this function does in the "unify" interface. """
#ifdef USE_DRIVERAPI
  driver_func(...)
#else  
  runtime_func(...)
#endif
}
```

> Currently, we have CUDADeviceGuard, which works as aten::DeviceGuard - this could also be implemented using the driver API. On the other hand, if CUDADeviceGuard only works with the runtime API while cubin_launcher can switch between runtime/driver APIs, it introduces some inconsistency.

The `CUDADeviceGurad` is orthogonal to cubin launcher. We can leave the support of `CUDADeviceGurad` with driver api in the future when needed.

oraluben:
> I asked Gemini on the best parctice of indicating "internal-only" header and it suggests us to use some "detail" subfolder and define the internal functions in "details" namespace. Maybe we can create a "detail" subfolder like /include/tvm/ffi/extra/cuda/detail and define the compatibility layer in a "details" namespace.

My gemini also prefers "detail", but I do think "internal" would be better :)

Currently we have `tvm/ffi/extras/cuda/internal/unified_api.h` and a new namespace `tvm::ffi::cuda_api` for all those internal type/impls.

I believe all comments have been resolved, PTAL again @yaoyaoding 

cc @junrushao

junrushao:
I’d love to delegate this PR to Yaoyao and just let me know when it’s ready to merge 🫡

yaoyaoding (approved):
The PR looks good to me now, thanks @oraluben ! 

cc @junrushao

junrushao:
Thank you both! This is huge amount of work 🫡

junrushao:
Looks like there's another issue from this PR:

```
[2026-01-12T20:01:26.601Z] /tmp/build-env-5u_p2hia/lib/python3.12/site-packages/tvm_ffi/include/tvm/ffi/extra/cuda/device_guard.h(52): error: a value of type "cudaError_t" (aka "cudaError") cannot be used to initialize an entity of type "tvm::ffi::cuda_api::ResultType" (aka "cudaError_enum")
[2026-01-12T20:01:26.601Z]       do { ::tvm::ffi::cuda_api::ResultType __err = (cudaGetDevice(&original_device_index_)); if (__err != ::tvm::ffi::cuda_api::kSuccess) { const char *__err_name, *__err_str; ::tvm::ffi::cuda_api::GetErrorString(__err, &__err_name, &__err_str); ::tvm::ffi::details::ErrorBuilder("RuntimeError", TVMFFIBacktrace("/tmp/build-env-5u_p2hia/lib/python3.12/site-packages/tvm_ffi/include/tvm/ffi/extra/cuda/device_guard.h", 52, __PRETTY_FUNCTION__, 0), 0) .stream() << "CUDA Error: " << __err_name << " (" << static_cast<int>(__err) << "): " << __err_str; } } while (0);
```

PR #306, opened 2025-12-03
chore: Use `importlib.metadata` to locate DSOs

### LINKED ISSUE #250: [Bug] DSO path finding should not depend on hardcoded paths
Current path finding for `libtvm_ffi.so` and `libtvm_ffi_testing.so` heavily depends on hardcoded paths. More concretely, assuming `build/lib` as part of the path for editable development.

https://github.com/apache/tvm-ffi/blob/53a7fe9f64a96eacdb044b1d7ac1ed1d17f3da70/python/tvm_ffi/libinfo.py#L55-L58

However, this is not an optimal strategy for editable builds, because:
1) build directory could be different than `build/`, e.g. `build-release`, `build/release`, etc.
2) the actual `.so` files are actually installed into Python's `site-packages` directory by `scikit-build-core`. The `.so` files can get outdated or mismatch inside `build/`

Let's switch to a more standard practice in this case for both `apache-tvm-ffi` and the example extension in `my_ffi_extension` under `examples/packaging`, by using the snippet below:

```python
import importlib.metadata as im

def find_files(
  package: str,  # pass in `__package__`
  stem: str  # pass in `tvm_ffi`, so that it finds `lib{stem}.so`, `lib{stem}.so`, `{stem}.dll`
):
    owners: list[str] = im.packages_distributions().get(package)
    pkg = owners[0] if owners else package.replace("_", "-")
    for pth in im.files(pkg):
        print("File: ", pth.locate())
    ...
```

junrushao:
Should fix #250. 

This PR introduces `tvm_ffi.lib_info.load_lib_ctypes`, which can be reused by downstream libraries to locate their DSOs if they are linked with tvm_ffi package.

For packages that are shipped with tvm-ffi, the one-liner below could load it into ctypes:

```
LIB = tvm_ffi.libinfo.load_lib_ctypes("apache-tvm-ffi", "tvm_ffi", "RTLD_LOCAL")
```

Here
- `apache-tvm-ffi` is the name of the package
- `tvm-ffi` is the target name of the shared library
- `mode` is ctypes opening mode


python/tvm_ffi/libinfo.py
```diff
@@ -22,78 +22,152 @@
...
+        ],
+        cond=lambda p: (p / "tvm_ffi_python_helpers.h").is_file(),
+    ):
+        return ret
+    raise RuntimeError("Cannot find python helper include path.")
+
+
+def include_paths() -> list[str]:
+    """Find all include paths needed for FFI related compilation."""
+    return sorted(
```
tqchen, line 136: why sorted? likely behavior is already deterministic

python/tvm_ffi/libinfo.py
```diff
@@ -22,78 +22,152 @@
...
 
 
-def find_library_by_basename(base: str) -> str:
+def load_lib_ctypes(package: str, base: str, mode: str) -> ctypes.CDLL:
+    """Load the tvm_ffi shared library by searching likely paths."""
+    lib_path: Path = _find_library_by_basename(package, base)
+    # The dll search path need to be added explicitly in windows
+    if sys.platform.startswith("win32"):
+        os.add_dll_directory(str(lib_path.parent))
+    return ctypes.CDLL(str(lib_path), getattr(ctypes, mode))
```
tqchen, line 169: not too sure why we need getattr here, but minor

tqchen (commented):
minor comments

python/tvm_ffi/libinfo.py
junrushao: its deduplicated using a `set`. Technically set is fine preserving orders in Python, but just want to make sure its deterministic

python/tvm_ffi/libinfo.py
junrushao: if `mode="RTLD_LOCAL"`, it is equivalent to accessing `ctypes.RTLD_LOCAL`.

The reason I don't love asking users to pass in `ctypes.RTLD_LOCAL` is that it is an `int`, which is semantically not very meaningful

python/tvm_ffi/libinfo.py
tqchen: i feel relying on set order preserving is fine

junrushao: sorted is even better given it's always deterministic across platforms and python versions

PR #309, opened 2025-12-04
[FEAT] kwargs wrapping utility

tqchen:
This PR introduces a kwargs_wrapper that can wrap an existing positional only function to a function with optional kwargs signatures through code generation. We carefully limit the code generation scope to not impact by untrusted input so generated code is safe.

### DIFF
--- a/python/tvm_ffi/utils/__init__.py
+++ b/python/tvm_ffi/utils/__init__.py
@@ -16,4 +16,5 @@
 # under the License.
 """Utilities used by the tvm_ffi Python package."""
 
+from . import kwargs_wrapper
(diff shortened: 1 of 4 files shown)


junrushao:
So I happened to have a similar implementation when trying to implement dataclass-style `__init__`: https://github.com/apache/tvm-ffi/blob/main/python/tvm_ffi/dataclasses/_utils.py#L101. Maybe could unify to this one as future work

junrushao (approved):
I'm getting this PR in as it looks perfect to me. The checks there are a bit too defensive, but fine

PR #324, opened 2025-12-08
[ADDON] bump torch c dlpack ext version to 0.1.4

cyx-6:
bump torch c dlpack ext version to 0.1.4


junrushao:
Nice catch @cyx-6 ! Could you please also update the linter so that it checks version numbers automatically for `torch_c_dlpack`? https://github.com/apache/tvm-ffi/blob/main/tests/lint/check_version.py

PR #327, opened 2025-12-10
[Feat] Remove reference cycle in error handling for faster GC

KEKE046:
When training neural networks, it is beneficial to free tensors as early as possible once they are no longer used. In CPython, as long as there is no reference cycle, tensors are freed immediately when their reference count drops to zero.

Torch blog: [Finding and Removing Reference Cycles](https://pytorch.org/blog/understanding-gpu-memory-2/)

However, in our TVM FFI error handling path, we create reference cycles that keep intermediate tensors (and the entire call chain) alive much longer than necessary. This slows down local GC, increases memory pressure, and hurts training throughput.

## Problem

The following code represents the problem.
1. We allocate a tensor in foo
2. We call `Map.get` with a non-existent key, which causes a KeyError and will be captured by Map
3. The KeyError causes a reference cycle, and makes the whole call chain require gc

``` py
import torch
import tvm_ffi
import gc
from torch.utils.viz._cycles import warn_tensor_cycles

m = tvm_ffi.Map({'a': 1})

def foo():
    a = torch.tensor([1], device='cuda')
    # map get a non-existent key will cause a KeyError and capture it
    _tmp = m.get('b', 0)

foo()

remove = warn_tensor_cycles()
gc.collect()
remove()
```

The following figure demonstrates the call chain and reference cycles:

image

The following figure shows the object reference graph from torch warn_tensor_cycles

image

## Approach

As shown in the figure, the 3 variables, `frame`, `py_error`, and `tb`, are the core of the reference cycle. To remove this error, we can manually delete the 3 variables in the loop.

https://github.com/apache/tvm-ffi/blob/f7e09d6a96b54554190bae0d7ba9ff7a6e9a109e/python/tvm_ffi/error.py#L124-L136

We can use `try ... finally ...` to manually delete these variables after the function returns.

```py
    try:
        tb = py_error.__traceback__
        for filename, lineno, func in _parse_backtrace(backtrace):
            tb = _TRACEBACK_MANAGER.append_traceback(tb, filename, lineno, func)
        return py_error.with_traceback(tb)
    finally:
        del py_error, tb
```

### DIFF
--- a/tests/python/test_error.py
+++ b/tests/python/test_error.py
@@ -138,7 +138,7 @@ def test_error_no_cyclic_reference() -> None:
             except KeyError:
                 pass
 
-        def may_create_cyclic_reference() -> weakref.ReferenceType[SampleObject]:
+        def may_create_cyclic_reference() -> weakref.ReferenceType:
             obj = SampleObject()
             trigger_cpp_side_error()
             return weakref.ref(obj)


python/tvm_ffi/error.py
```diff
@@ -121,19 +121,27 @@ def append_traceback(
...
 
         """
-        frame = self._create_frame(filename, lineno, func)
-        return types.TracebackType(tb, frame, frame.f_lasti, lineno)
+        # we manually delete frame to avoid reference cycle, making it faster to gc the locals inside the frame
+        try:
+            frame = self._create_frame(filename, lineno, func)
+            return types.TracebackType(tb, frame, frame.f_lasti, lineno)
+        finally:
+            del frame
```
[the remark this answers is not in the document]
KEKE046: Yes, you're right

KEKE046:
No @gemini-code-assist, it's wrong, it creates a `f_locals` reference

image


But I come up with another hack to prevent circular reference:

```py
  # Magic hack to prevent a circular reference
  def create(tb, frame, lineno):
      return types.TracebackType(tb, frame, frame.f_lasti, lineno)
  return create(tb, self._create_frame(filename, lineno, func), lineno)
```

LeiWang1999:
LGTM!

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,25 @@ def append_traceback(
             The new traceback with the appended frame.
 
         """
-        frame = self._create_frame(filename, lineno, func)
-        return types.TracebackType(tb, frame, frame.f_lasti, lineno)
+        # Magic hack to prevent a circular reference
+        def create(tb: types.TracebackType | None, frame: types.FrameType, lineno: int) -> types.TracebackType:
```
oraluben: Would it do the same with a anonymous lambda?

KEKE046: Absolutely!

But I think anonymous lambda is a little not clear, as shown below
```py
(lambda frame: types.TracebackType(tb, frame, frame.f_lasti, lineno))(self._create_frame(filename, lineno, func))
```

junrushao:
This is an incredible finding! CC: @tqchen

tqchen:
Thanks @KEKE046 , @junrushao let us prioritize merging this

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,29 @@ def append_traceback(
             The new traceback with the appended frame.
 
         """
-        frame = self._create_frame(filename, lineno, func)
-        return types.TracebackType(tb, frame, frame.f_lasti, lineno)
+
+        # Magic hack to prevent a circular reference
```
tqchen: can we add more comments on why this function prevent circular reference, would beuseful for future references.

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,29 @@ def append_traceback(
...
 _TRACEBACK_MANAGER = TracebackManager()
 
 
 def _with_append_backtrace(py_error: BaseException, backtrace: str) -> BaseException:
     """Append the backtrace to the py_error and return it."""
-    tb = py_error.__traceback__
-    for filename, lineno, func in _parse_backtrace(backtrace):
-        tb = _TRACEBACK_MANAGER.append_traceback(tb, filename, lineno, func)
-    return py_error.with_traceback(tb)
+    # we manually delete py_error and tb to avoid reference cycle, making it faster to gc the locals inside the frame
```
tqchen: would be great to add explainations of why deling py_error, tb helps

tqchen:
@KEKE046 would be good to go through one pass over gemini reviews and add comments about context, thank you so much!

python/tvm_ffi/error.py
KEKE046: 👌 I think I can draw a picture and call Gemini to convert it to a comments

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,65 @@ def append_traceback(
             The new traceback with the appended frame.
 
         """
-        frame = self._create_frame(filename, lineno, func)
-        return types.TracebackType(tb, frame, frame.f_lasti, lineno)
+
+        # The `create` function is a hack to prevent append_traceback from holding its child frame
+        # please see the reference cycle diagram in _with_append_backtrace and pull request #327 for more details
+        #
+        # This hack prevent binding `self._create_frame` to a writable local variable
```
tqchen: This hack => This approach avoid  binding `self._create_frame` to a writable local variable

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,65 @@ def append_traceback(
...
+    #     +-------------------+                              |
+    #     | _get_frame        | <----------------------------+
+    #     +-------------------+      (Cycle closes here)
+    try:
+        tb = py_error.__traceback__
+        for filename, lineno, func in _parse_backtrace(backtrace):
+            tb = _TRACEBACK_MANAGER.append_traceback(tb, filename, lineno, func)
+        return py_error.with_traceback(tb)
+    finally:
+        # this is a hack to break reference cycle
```
tqchen: "we explicitly break reference cycle here", this is deliberate and not a work around, so we do not need to say it is a "hack"

python/tvm_ffi/error.py
```diff
@@ -121,19 +121,29 @@ def append_traceback(
...
-        tb = _TRACEBACK_MANAGER.append_traceback(tb, filename, lineno, func)
-    return py_error.with_traceback(tb)
+    # we manually delete py_error and tb to avoid reference cycle, making it faster to gc the locals inside the frame
+    try:
+        tb = py_error.__traceback__
+        for filename, lineno, func in _parse_backtrace(backtrace):
+            tb = _TRACEBACK_MANAGER.append_traceback(tb, filename, lineno, func)
+        return py_error.with_traceback(tb)
+    finally:
+        del py_error, tb
```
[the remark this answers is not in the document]
tqchen: @KEKE046 shall we apply genimi suggestion here?

python/tvm_ffi/error.py
KEKE046: 👌, I'll add comprehensive comments

tqchen:
Thanks @KEKE046 i think we are close, please take a look at the latest comments

python/tvm_ffi/error.py
KEKE046: I'll draw a figure, and let Gemini convert it to an ASCII comment

tqchen:
Another minor question, is it possible to create a regression test about this? genai suggests that maybe one way to do is to create an weakref that calls to a callback, which get triggered in del, then we confirm in your example if the torch.Tensor or anything actually get recycled in an exception

```python
import gc
import weakref

# List to track if the object was collected
collected_list = []

def callback(wr):
    """Callback triggered when the weakref object is finalized."""
    collected_list.append(True)

def create_local_cycle_and_collect():
    # 1. Create an object (a list is simple)
    obj = [1, 2, 3]

    # 2. Create a weak reference to the object
    # This reference will trigger 'callback' when 'obj' is garbage collected.
    ref = weakref.ref(obj, callback)
```

python/tvm_ffi/error.py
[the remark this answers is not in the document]
KEKE046: That's good, but I think we can just move `tb = py_error.__traceback__` outside the try block. Because `py_error` always helds a `__traceback__`, this can prevent tb from undefined

KEKE046:
Sorry, @tqchen , you're so fast that my push may be outdated

KEKE046:
I'll try to create a test to verify the error

KEKE046:
Some lint errors don't occur in my environment, so I'll check it

tqchen (approved):

junrushao:
Incredible work!

tqchen:
Thanks @KEKE046 for great work!

PR #333, opened 2025-12-11
chore(lint): Run `torch_c_dlpack_ext` version check in linter

junrushao:
See also: https://github.com/apache/tvm-ffi/pull/324


junrushao:
@cyx-6 we will likely need v0.1.5 for `torch_c_dlpack_ext` if I understand correctly

junrushao:
Looks like the protocol for now is that `torch_c_dlpack_ext` doesn't have to follow `tvm_ffi`'s versioning

PR #347, opened 2025-12-17
Allow `tvm_ffi.device(..., id)` where id is numpy or torch scalar

wrongtest-intellif:
Similar to torch.device

```python
torch.device("cuda", 0)
torch.device("cuda", numpy.int32(1))
torch.device("cuda", torch.tensor(1, dtype=torch.int32))
```

```python
tvm_ffi.device("cuda", 0)
tvm_ffi.device("cuda", numpy.int32(1))
tvm_ffi.device("cuda", torch.tensor(1, dtype=torch.int32))
```

### DIFF
--- a/tests/python/test_device.py
+++ b/tests/python/test_device.py
@@ -22,7 +22,6 @@ import pickle
 
 import numpy
 import pytest
-import torch
 import tvm_ffi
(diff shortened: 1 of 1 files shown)


junrushao (approved):
Nice work!

PR #348, opened 2025-12-17
feat: add Rust binding for `Array<T>`

khj809:
This PR introduces a Rust implementation of the `Array` container.

### Key Features
* **Memory Safety**: Correctly handles reference counting for `ObjectRef` elements. It ensures `inc_ref` is called during retrieval (`get`) and `dec_ref` is called when elements are removed or the array is cleared.
* **Dynamic Mutation**: Implements `push`, `pop`, `insert`, `remove`, and `clear`. It handles internal growth and reallocation while maintaining compatibility with the underlying C++ memory layout.
* **FFI Compatibility**: Explicitly manages the `data` pointer within `ArrayObj` to allow C++ TVM functions to traverse the array using standard pointer arithmetic.
* **Type System Integration**:
    * Implements `AnyCompatible`, allowing `Array<T>` to be erased into `Any` and `AnyView` and recovered via `TryFrom`.
    * Implements `FromIterator` and `Extend`, enabling seamless integration with Rust's iterator ecosystem.

## Tests
Verified with a comprehensive test suite in `tvm-ffi/tests/test_array.rs` covering:
- [x] Basic creation and iteration.
- [x] Out-of-bounds safety.
- [x] Dynamic growth and reallocation (push/insert).
- [x] Memory integrity after element shifting (remove/insert).
- [x] Roundtrip conversions through `Any` and `AnyView`.
- [x] Parametric support for both `Tensor` and `Shape` types.


junrushao:
Nice work! Likely we will need Apache license headers on each file :)

```
ERROR: Missing ASF header in rust/tvm-ffi/src/collections/array.rs
Run `python tests/lint/check_asf_header.py --fix` to add the header
ERROR: Missing ASF header in rust/tvm-ffi/tests/test_array.rs
Run `python tests/lint/check_asf_header.py --fix` to add the header
```

khj809:
@junrushao Thanks. I added it :)

junrushao:
Looks like there's a test failure:

```
---- test_array_parametric_heterogeneity stdout ----

thread 'test_array_parametric_heterogeneity' (10146) panicked at tvm-ffi/tests/test_array.rs:145:46:
called `Result::unwrap()` on an `Err` value: Traceback (most recent call last):

  File "tvm-ffi/src/function.rs", line 190, in tvm_ffi::function::Function::get_global
RuntimeError: Function testing.echo not found
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_array_parametric_heterogeneity

error: test failed, to rerun pass `-p tvm-ffi --test test_array`
```

khj809:
@junrushao Hmm it's strange because `tvm_ffi_testing` target is already linked in `tvm-ffi-sys` crate, and `testing.echo` is already used in `test_function.rs` too. Is there anything I've missed to enable testing functions during cargo test?

khj809:
@junrushao Due to the unexpected link errors of testing functions on CI environment, I decided to use global functions which definitely exist in libtvm_ffi dylib.

junrushao:
CC @tqchen please take a look when you got time, as I'm not very experienced in Rust

tqchen:
Thanks a lot for contributions, i maybe a bit off recently but should be able to come back and take closer look around Jan. would be good to also get some eyes for rust reviews. cc @DarkSharpness @Ubospica if u have bandwidth

Some high-level feedbacks:

- Likely we can use Index overload, so we can do array[i]
- Avoid doing the object-type-index generic conversion in get(or index), since sometimes T can be backed by both special and Object, instead use AnyCompatible trait, which is meant to define conversion rule of T from/to any
- Keep array immutable for now, as of now we are still discussing mutable handling, so let us keep array immutable

khj809:
@tqchen Thanks for the review.
- About Index trait, the fn `index` requires to return the reference of output type, but it's technically impossible to return `&T` because `ArrayObj` does not store the buffer of `T` directly. Instead it's possible to return a reference of either `TVMFFIAny` or `AnyView`, so I implemented it to return `&AnyView` for now.
- As you mentioned, I tried to use `AnyCompatible` instead of the previous `ObjectRefCore`. While doing this, I found the issue that `Any::into_raw_ffi_any` drops the data immediately and makes the array of objects having invalid data. Hence I added `ManuallyDrop` to prevent the destructor from running at the end of the scope.
- As you mentioned, I removed all the mutable methods for now.

khj809:
@tqchen @junrushao I'd like to kindly remind you about this PR. Could you take a look if you have time?

junrushao:
@tqchen please take a look when you have some bandwidth

PR #364, opened 2025-12-24
feat: Restrict `__slots__=()` for subclasses of `tvm_ffi.Object` by default

junrushao:
This PR makes `__slots__=()` for all subclasses of `tvm_ffi.Object` if not specified. Here's all the changes:

- Rename the existing Cython class `Object` to `CObject`
- Create a new `Object` whose metaclass is `_ObjectSlotsMeta`
- Encourage inheriting from the new `tvm_ffi.Object`


tqchen:
this may have issues with slot equals () by default since downstream subclass may rely on the slot. it would be useful to have users opt-in, instead of opt-out (just like default python class behavior)

junrushao:
Closing - will open a new PR once unblocked.

Issue #357, opened 2025-12-22
[BUG] int64_t cannot represent the full range of size_t


nan2088:
```python
import tvm_ffi
mod = tvm_ffi.cpp.load_inline(name='hello',
                              cpp_sources="""
                          #include <limits>
                          tvm::ffi::Any example_func() {
                             return std::numeric_limits<size_t>::max();
                          }
                          """,
                              functions=['example_func'])

result = mod.example_func()

print(f'result={result} type={type(result)}')
```

result=-1 type=<class 'int'>

tqchen:
Previously we made this deliberate choice to simplify integer passing. As a possible workaround, one could have a custom wrapping of int and reinterpret_cast uint64 as int64 and return that

nan2088:
That makes sense. By the way, would it be a good idea to add a load parameter (e.g., strict_u64_cast=True) and a corresponding macro to add a compile-time check and prevent silent overflow?

junrushao:
> add a compile-time check and prevent silent overflow

That would make more sense if we check overflow strictly - we used to have this check in TVM IIRC. More specifically, you may want to add an `if constexpr (Int is uint64)` here:

https://github.com/apache/tvm-ffi/blob/0d8fec88eca5acaeaa7e771003d183c31711574a/include/tvm/ffi/type_traits.h#L281-L285

and if `Int` is `uint64`, it checks overflow

nan2088:
> > add a compile-time check and prevent silent overflow
> 
> That would make more sense if we check overflow strictly - we used to have this check in TVM IIRC. More specifically, you may want to add an `if constexpr (Int is uint64)` here:
> 
> [tvm-ffi/include/tvm/ffi/type_traits.h](https://github.com/apache/tvm-ffi/blob/0d8fec88eca5acaeaa7e771003d183c31711574a/include/tvm/ffi/type_traits.h#L281-L285)
> 
> Lines 281 to 285 in [0d8fec8](/apache/tvm-ffi/commit/0d8fec88eca5acaeaa7e771003d183c31711574a)
> 
>  TVM_FFI_INLINE static void CopyToAnyView(const Int& src, TVMFFIAny* result) { 
>    result->type_index = TypeIndex::kTVMFFIInt; 
>    result->zero_padding = 0; 
>    result->v_int64 = static_cast<int64_t>(src); 
>  } 
> and if `Int` is `uint64`, it checks overflow

[...]

PR #370, opened 2025-12-29
[feat] Add overflow check for uint64_t/size_t in TypeTraits<Int>::CopyToAnyView

nan2088:
Related discussion here  #357 
Adds runtime overflow detection when converting unsigned 64-bit integers ([uint64_t or size_t](https://stackoverflow.com/questions/73820966/stdis-same-vsize-t-uint64-t-evaluates-to-false-when-both-types-are-8-by
)) to int64_t in TVM FFI Any. 

128-bit signed integers (if exists) are not considered.

### DIFF
--- a/src/ffi/extra/structural_hash.cc
+++ b/src/ffi/extra/structural_hash.cc
@@ -152,6 +152,9 @@ class StructuralHashHandler {
                 std::swap(allow_free_var, map_free_vars_);
                 return static_cast<int64_t>(details::StableHashCombine(init_hash, hash_value));
               } else {
+                // we explicitly bitcast the result from `uint64_t` to `int64_t`.
+                // The range of `uint64_t` is too large to fit as `int64_t`, so if we don't bitcast,
+                // it will trigger an overflow error in `uint64_t` -> `Any` conversion.
                 return static_cast<int64_t>(details::StableHashCombine(init_hash, HashAny(val)));
               }
             });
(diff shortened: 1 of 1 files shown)


junrushao:
@nan2088 I updated your branch and fixed the C++ unittest. It should work now.

CC: @tqchen for review

src/ffi/extra/structural_hash.cc
```diff
@@ -150,15 +150,16 @@ class StructuralHashHandler {
                 std::swap(allow_free_var, map_free_vars_);
                 uint64_t hash_value = HashAny(val);
                 std::swap(allow_free_var, map_free_vars_);
-                return details::StableHashCombine(init_hash, hash_value);
+                return static_cast<int64_t>(details::StableHashCombine(init_hash, hash_value));
```
tqchen, line 153: Need explicit comment.

// static_cast from int64_t to uint64_t will behave exactly like bit cast

tqchen: do the same for other places of the static cast

src/ffi/extra/structural_hash.cc
```diff
@@ -311,9 +312,14 @@ uint64_t StructuralHash::Hash(const Any& value, bool map_free_vars, bool skip_te
   return handler.HashAny(value);
 }
 
+static int64_t FFIStructuralHash(const Any& value, bool map_free_vars, bool skip_tensor_content) {
```
tqchen, line 318: add comment on why here, we explicitly bitcast(static cast effectively does bit cast) to int64 so returned value bypasses the signed check

src/ffi/extra/structural_hash.cc
tqchen, line 153: comment, we explicitly bitcast to i64 so return value bypasses the bound check

PR #375, opened 2026-01-01
fix: add bounds checking for size() and stride() methods

guan404ming:
## Why

size() and stride() methods could access memory out of bounds when given invalid indices

## How

- Add bounds checking after adjusting negative indices
- Throw IndexError with descriptive message when index is out of bounds
- Add tests for both Tensor and TensorView classes


junrushao (approved):
Looks good to me!

junrushao:
Given that we already have conditionals in `size()`, I'd say it won't add to too much overhead and happy to get it in

guan404ming:
Thanks!

PR #376, opened 2026-01-02
fix: add negative index bounds check in ArrayObj

guan404ming:
## Why

`ArrayObj::operator[]` and `ArrayObj::SetItem` were missing negative index validation, causing undefined behavior when accessing arrays with negative indices.

## How

- Add `i < 0` check to `ArrayObj::operator[]` and `ArrayObj::SetItem` bounds validation
- Add test case


guan404ming:
Sure, I've resolved it. Thanks!

PR #379, opened 2026-01-02
docs: document structural_equal and structural_hash for Array and Map

guan404ming:
## Why

Users need guidance on how to perform structural equality comparison for containers.

## How

- Add notes to Array and Map docstrings pointing to structural_equal and structural_hash APIs

### DIFF
--- a/python/tvm_ffi/container.py
+++ b/python/tvm_ffi/container.py
@@ -144,6 +144,10 @@ class Array(core.Object, Sequence[T]):
         a = tvm_ffi.Array([1, 2, 3])
         assert tuple(a) == (1, 2, 3)
 
+    Notes
+    -----
+    For structural equality and hashing, use ``structural_equal`` and ``structural_hash`` APIs.
+
     See Also
     --------
     :py:func:`tvm_ffi.convert`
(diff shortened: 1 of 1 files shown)


junrushao:
I'm not totally sure about using structural equality/hash for array/map comparison - It's a recursively defined equality based on IR structure, which is rather compute intensive. Does Python define it's tuple/dict equality recursively?

guan404ming:
> I'm not totally sure about using structural equality/hash for array/map comparison - It's a recursively defined equality based on IR structure, which is rather compute intensive. Does Python define it's tuple/dict equality recursively?

Thanks for the question! Based on my review of CPython's implementation, Python does use recursive equality for tuple/dict. For example, tuple_richcompare (tupleobject.c:689) calls PyObject_RichCompareBool recursively on each element. Dict equality follows the same pattern. So I think `StructuralEqual` matches Python's semantics here.

Reference

- Tuple equality: https://github.com/python/cpython/blob/main/Objects/tupleobject.c - Uses PyObject_RichCompareBool() to recursively compare each element
- Dict equality: https://github.com/python/cpython/blob/main/Objects/dictobject.c - Similar recursive element-by-element comparison

tqchen:
Thanks for the contribution. For this particular case, i feel we should defer a bit, since structural equalty is already supported through explicitly.  While classes like list do not necessarily support nested comparison directly via operators. For immutable objects there is some merits for having the feature. This is mainly a consideration of minimalism and also consistency with wrt to instances like c++.



As a middle ground, perhaps let us first document structural_equal and structural_hash APIs and refer users to them. Personally i think keeping it simple here might better

guan404ming:
Thanks for the review and suggestion! I agree that keeping it simple and explicit is the better approach. I've updated the PR to just add documentation pointing users to structural_equal and structural_hash APIs.

PR #380, opened 2026-01-03
feat: add __bool__ support for Array and Map

guan404ming:
## Why

Python's built-in list and dict return False when empty. However, Array and Map containers always returned True because they didn't implement __bool__, making bool(Array([])) and bool(Map({})) return True unexpectedly.

## How

- Python: Implemented __bool__ for Array and Map returning len(self) > 0
- Tests: Added parametrized tests verifying truthy/falsy behavior for empty and non-empty containers

### DIFF
--- a/python/tvm_ffi/container.py
+++ b/python/tvm_ffi/container.py
@@ -192,6 +192,10 @@ class Array(core.Object, Sequence[T]):
         """Check if the array contains a value."""
         return _ffi_api.ArrayContains(self, value)
 
+    def __bool__(self) -> bool:
+        """Return True if the array is non-empty."""
+        return len(self) > 0
+
     def __add__(self, other: Iterable[T]) -> Array[T]:
         """Concatenate two arrays."""
         return type(self)(itertools.chain(self, other))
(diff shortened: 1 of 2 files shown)


python/tvm_ffi/container.py
```diff
@@ -192,6 +192,10 @@ def __contains__(self, value: object) -> bool:
         """Check if the array contains a value."""
         return _ffi_api.ArrayContains(self, value)
 
+    def __bool__(self) -> bool:
+        """Return True if the array is non-empty."""
+        return self.__chandle__() != 0 and len(self) > 0
```
junrushao: I don't think there's a need to check `__chandle__()`

```suggestion
        return len(self) > 0
```

python/tvm_ffi/container.py
```diff
@@ -337,6 +341,10 @@ def __len__(self) -> int:
         """Return the number of items in the map."""
         return _ffi_api.MapSize(self)
 
+    def __bool__(self) -> bool:
+        """Return True if the map is non-empty."""
+        return self.__chandle__() != 0 and len(self) > 0
```
junrushao: ditto

python/tvm_ffi/container.py
guan404ming: Sure, I've removed it. Thanks!

junrushao (approved):

guan404ming:
Thanks!

PR #383, opened 2026-01-06
feat: add find and substr methods to String

guan404ming:
## Why

String class lacks essential string manipulation methods that are commonly expected in standard library equivalents, requiring users to convert to std::string for basic operations like substring search and extraction.

## How

- Add find() method with three overloads (String, const char*, const char* with count) to search for substrings
- Add substr() method to extract substrings with bounds checking


guan404ming:
Thanks!

Issue #356, opened 2025-12-22
How far are we from `@dataclass`


junrushao:
Resolved by
- https://github.com/apache/tvm-ffi/pull/477
- https://github.com/apache/tvm-ffi/pull/499
- https://github.com/apache/tvm-ffi/pull/555
- https://github.com/apache/tvm-ffi/pull/556

-----------------------------------------------------------------------

- [x] Init: generates `__init__` (we implemented some of the logics in @c_class)
  - ✅ flag `init=True` in `@dataclass`
  - ✅ flag `init=True` in `dataclasses.field`
  - ✅ run `__post_init__` if presents done 
  - ✅ default=MISSING, default_factory=MISSING in `dataclasses.field`
  - ❌ InitVar
  - ✅ ClassVar
  - ✅ `kw_only=False` in `@dataclass`
  - ✅ `kw_only=MISSING` in `dataclasses.field`
  - ✅ dataclasses.KW_ONLY
- [x] Repr: generates `__repr__`
  - ✅ flag `repr=True` in `@dataclass`
  - ✅ flag `repr=True` in `dataclasses.field`

- [x] Copy: Generate copy interface
  - ✅ `def __copy__(self)`
  - ✅ `def __deepcopy__(self, memo)`
  - ✅ `def __replace__(self, **changes)` and dataclasses.replace

- [x] Equality
  - ✅ `eq=True` in `@dataclass`: generates `__eq__` and `__ne__`

- [x] Comparison
  - ✅ `order=False` in `@dataclass`: generates `__lt__`, `__le__`, `__gt__`, and `__ge__`, 
  - ✅ `compare=True` in `dataclasses.field`

- [x] Mutability and extra slots:
  - ✅ `slots`
  - ❌ `frozen=False`: mutability
  - ❌ `weakref_slot=False`

- [x] Hashing
  - ❌ `unsafe_hash=False`
  - ✅ hash=None

Misc
- [x] dataclasses.replace
- [x] dataclasses.fields
- [x] dataclasses.is_dataclass
- [x] dataclasses.asdict
- [x] dataclasses.astuple
- [x] `match_args=True` in `@dataclass`
- [ ] metadata=None in `dataclasses.field`
- [ ] doc=None in `dataclasses.field`
- [ ] dataclasses.make_dataclass

junrushao:
PR #555 adds a bit more coverage to dataclass APIs

junrushao:
PR #556 adds `asdict`, `astuple` and `match_args`

PR #384, opened 2026-01-06
feat: add kw_only support for dataclass init generation

guan404ming:
## Related Issue

#356

## Why

Python's standard dataclasses support kw_only parameter to make fields keyword-only in __init__. This feature was missing from @c_class decorator.

## How

- Add `KW_ONLY` sentinel class for marking keyword-only fields
- Add `kw_only` parameter to field() function and @c_class decorator
- Update `method_init()` to generate proper signature
- Add tests

### DIFF
--- a/python/tvm_ffi/dataclasses/field.py
+++ b/python/tvm_ffi/dataclasses/field.py
@@ -18,8 +18,17 @@
 
 from __future__ import annotations
 
-from dataclasses import _MISSING_TYPE, KW_ONLY, MISSING  # type: ignore[attr-defined]
+from dataclasses import _MISSING_TYPE, MISSING
 from typing import Any, Callable, TypeVar, cast
+
+try:
+    from dataclasses import KW_ONLY  # type: ignore[attr-defined]
+except ImportError:
+    # Python < 3.10: define our own KW_ONLY sentinel
+    class _KW_ONLY_Sentinel:
+        __slots__ = ()
+
+    KW_ONLY = _KW_ONLY_Sentinel()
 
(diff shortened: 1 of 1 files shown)


python/tvm_ffi/dataclasses/field.py
```diff
@@ -24,6 +24,31 @@
 _FieldValue = TypeVar("_FieldValue")
 
 
+class _KW_ONLY_TYPE:
```
junrushao: No need to redefine a new class - You may want to reuse `dataclasses.KW_ONLY`

tests/python/test_dataclasses_c_class.py
```diff
@@ -94,3 +98,61 @@ def test_cxx_class_init_subset_positional() -> None:
...
     obj.optional_field = 11
     assert obj.optional_field == 11
+
+
+@c_class("testing.TestCxxKwOnly", kw_only=True)
+class _TestKwOnly:
+    x: int
+    y: int
+    z: int
+    w: int = 100
```
junrushao: Move this to `testing.py`

junrushao:
<del>OK it turns out KW_ONLY may not exist in `dataclasses` in Python 3.8. This is silly :)</del>

<del>If import fails, let's do this:</del>

```python
_KW_ONLY_TYPE = type("_KW_ONLY_TYPE", (object, ), {})
KW_ONLY = _KW_ONLY_TYPE()
```

**Update** Looks like your fix is proper. Thanks!

guan404ming:
Thanks!

PR #386, opened 2026-01-06
chore: replace C-style cast with static_cast in String::substr

guan404ming:
## Related

https://github.com/apache/tvm-ffi/actions/runs/20755375337/job/59595882892

## Why

clang-tidy fails with google-readability-casting error due to C-style cast in default parameter.

## How

Replace size_t(-1) with `nops` in String::substr method


junrushao (approved):
LGTM

guan404ming:
Thanks!

PR #387, opened 2026-01-07
fix: out-of-bounds read in `StringViewToDLDataType_` with non-null-terminated strings

khj809:
## Summary
Fixes a critical bug in `StringViewToDataType_` that causes undefined behavior when parsing dtype strings from non-null-terminated `std::string_view`. This bug particularly affects Electron applications due to aggresive memory reuse in Chromium's allocators.

## Problem
The original implementation used `strtoul()` to parse numeric values (bits and lanes) directly from `std::string_view::data()`. However, `strtoul()` requires null-terminated strings and will continue reading beyond the string_view bounds until it finds a null terminator or non-digit character.

## Solution
Replace `strtoul()` with manual digit parsing that respects `string_view` boundaries:
```c++
auto parse_digits = [](const char*& ptr, const char* end) -> uint32_t {
  uint32_t value = 0;
  while (ptr < end && *ptr >= '0' && *ptr <= '9') {
    value = value * 10 + (*ptr - '0');
    ptr++;
  }
  return value;
};
```
- Never reads beyond `str.data() + str.length()`
- Works correctly with non-null-terminated strings
- No memory allocation overhead
- Same performance as `strtoul` but safe

## Testing
New test `TEST(DType, NonNullTerminatedStringView)` validates:
- Parsing with digit garbage after the string (`"float16999888777"` length 7 -> bits=16, not 16999888777)
- Parsing with lane specifications outside bounds (`"int32x4extradata"` length 5 -> lanes=1, not 4)
- Correct parsing when lanes are within bounds (`"bfloat16x2"` length 10 -> lanes=2)
- Non-null-terminated buffers (`"float64AAAAA"` length 7 -> float64)
- Edge cases with various dtype types (int, uint, float, bfloat)

### DIFF
--- a/src/ffi/dtype.cc
+++ b/src/ffi/dtype.cc
@@ -209,20 +209,48 @@ inline DLDataType StringViewToDLDataType_(std::string_view str) {
   const char* str_end = str.data() + str.length();
 
   // Helper lambda to parse decimal digits from a bounded string_view
-  // Returns the parsed value and updates ptr to point past the last digit
-  auto parse_digits = [](const char*& ptr, const char* end) -> uint32_t {
+  // Returns the parsed value and updates *ptr to point past the last digit
+  auto parse_digits = [](const char** ptr, const char* end) -> uint32_t {
     uint64_t value = 0;
-    const char* start_ptr = ptr;
-    while (ptr < end && *ptr >= '0' && *ptr <= '9') {
-      value = value * 10 + (*ptr - '0');
-      ptr++;
+    const char* start_ptr = *ptr;
+    while (*ptr < end && **ptr >= '0' && **ptr <= '9') {
+      value = value * 10 + (**ptr - '0');
+      (*ptr)++;
     }
     if (value > UINT32_MAX) {
       TVM_FFI_THROW(ValueError) << "Integer value in dtype string '"
-                                << std::string_view(start_ptr, ptr - start_ptr)
+                                << std::string_view(start_ptr, *ptr - start_ptr)
                                 << "' is out of range for uint32_t";
     }
     return static_cast<uint32_t>(value);
+  };
+
+  // Helper lambda to parse lanes specification (e.g., "x16" or "xvscalex4")
+  // Returns the parsed lanes value and updates *ptr to point past the lanes specification
+  // Supports scalable vectors with the "xvscale" prefix (represented as negative lanes)
+  auto parse_lanes = [&](const char** ptr, const char* end, const std::string_view& dtype_str,
+                         bool allow_scalable = false) -> uint16_t {
+    int multiplier = 1;
+    // Check for "xvscale" prefix for scalable vectors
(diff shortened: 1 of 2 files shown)


junrushao:
Interesting! @tqchen could you take a look?

src/ffi/dtype.cc
```diff
@@ -206,17 +206,46 @@ inline DLDataType StringViewToDLDataType_(std::string_view str) {
   dtype.bits = 32;
   dtype.lanes = 1;
   const char* scan;
+  const char* str_end = str.data() + str.length();
+
+  // Helper lambda to parse decimal digits from a bounded string_view
+  // Returns the parsed value and updates ptr to point past the last digit
+  auto parse_digits = [](const char*& ptr, const char* end) -> uint32_t {
```
tqchen: minor google c style pass ptr by ptr

junrushao: echo that. please do

```diff
- const char*& ptr
+ const char* ptr
```

khj809: Actually the `*ptr` need to be updated to point to the end of parsed digits, so I tried to update it as `const char** ptr`. Is that okay?

src/ffi/dtype.cc
```diff
@@ -293,19 +322,38 @@ inline DLDataType StringViewToDLDataType_(std::string_view str) {
...
+    if (endpt == digits_start || lanes_val == 0) {
+      TVM_FFI_THROW(ValueError) << "Invalid lanes specification in dtype '" << str
+                                << "'. Lanes must be a positive integer.";
+    }
+    if (lanes_val > UINT16_MAX) {
+      TVM_FFI_THROW(ValueError) << "Lanes value " << lanes_val
+                                << " is out of range for uint16_t in dtype '" << str << "'";
+    }
+    dtype.lanes = static_cast<uint16_t>(scalable_multiplier * lanes_val);
   }
```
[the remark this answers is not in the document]
khj809: Added `parse_lanes` for this.

src/ffi/dtype.cc
junrushao: Oh sorry I misread. Yes in this case you will need `const char** ptr `

src/ffi/dtype.cc
```diff
@@ -206,17 +206,61 @@ inline DLDataType StringViewToDLDataType_(std::string_view str) {
...
+
+  // Helper lambda to parse decimal digits from a bounded string_view
+  // Returns the parsed value and updates *ptr to point past the last digit
+  auto parse_digits = [](const char** ptr, const char* end) -> uint32_t {
+    uint64_t value = 0;
+    const char* start_ptr = *ptr;
+    while (*ptr < end && **ptr >= '0' && **ptr <= '9') {
+      value = value * 10 + (**ptr - '0');
+      (*ptr)++;
+    }
```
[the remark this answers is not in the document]
junrushao: Gemini is getting too much

PR #388, opened 2026-01-07
feat: add starts_with and ends_with to String

guan404ming:
## Why

String pattern matching is essential for validation and parsing in compiler infrastructure. Users currently lack efficient prefix/suffix checking methods.

## How

- Add `starts_with(prefix)` method supporting String, const char*, and raw pointer with length
- Add `ends_with(suffix)` method with same overload variants
- Add tests

### DIFF
--- a/include/tvm/ffi/string.h
+++ b/include/tvm/ffi/string.h
@@ -667,6 +667,15 @@ class String {
    * \param prefix The prefix to check for
    * \return true if the string starts with prefix, false otherwise
    */
+  bool starts_with(std::string_view prefix) const {
+    return starts_with(prefix.data(), prefix.size());
+  }
+
+  /*!
+   * \brief Check if the string starts with a prefix
+   * \param prefix The prefix to check for
+   * \return true if the string starts with prefix, false otherwise
+   */
   bool starts_with(const char* prefix) const { return starts_with(prefix, std::strlen(prefix)); }
 
   /*!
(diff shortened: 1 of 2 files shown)


include/tvm/ffi/string.h
```diff
@@ -655,6 +655,60 @@ class String {
     return String(data() + pos, rcount);
   }
 
+  /*!
+   * \brief Check if the string starts with a prefix
+   * \param prefix The prefix to check for
+   * \return true if the string starts with prefix, false otherwise
+   */
+  bool startswith(const String& prefix) const { return startswith(prefix.data(), prefix.size()); }
```
junrushao: maybe prefer `std::string_view`? It doesn't copy the string and seems a standardized interface to work with

junrushao (changes requested):

tqchen:
We try to intentionally minimize the implementation, if the function is not directly related to functions std::string, i think we can leave it to downstream

junrushao:
@tqchen My rule of thumb is that if it's header-only, it's likely fine

guan404ming:
Thanks for the discussion. I think I understand tq’s point about keeping the core FFI surface minimal and leaving string algorithms to downstream usage, and I also see the motivation behind the std::string_view / header-only argument.
I’ll keep this PR open for now and am happy to proceed based on the maintainers’ preference.

tqchen:
it is worhwhile cross checking stdcpp document. i did find this 

https://en.cppreference.com/w/cpp/string/basic_string/starts_with for c++20, so if we have a consistent name we can bring it in

guan404ming:
Thanks for the guidence. I've updated with correct api name align to cpp doc and added string_view overload

PR #389, opened 2026-01-07
ci: add clang-tidy to ci_test

guan404ming:
## Why

PR CI passed but main CI failed after merge because clang-tidy only ran on main branch (in ci_mainline_only.yml), not on PRs.

## How

- Add `clang-tidy` hook to `ci_test`

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -51,7 +51,7 @@ jobs:
         id: cpp_files
         run: |
           FILES=$(git diff --name-only --diff-filter=ACMR origin/${{ github.base_ref }}...HEAD -- \
-            '*.c' '*.cc' '*.cpp' '*.cxx' '*.h' '*.hpp' '*.hxx' | tr '\n' ' ')
+            '*.c' '*.cc' '*.cpp' '*.cxx' | tr '\n' ' ')
           echo "files=$FILES" >> $GITHUB_OUTPUT
           [ -n "$FILES" ] && echo "changed=true" >> $GITHUB_OUTPUT || echo "changed=false" >> $GITHUB_OUTPUT
 


guan404ming:
I tested this by making random modifications to a C++ file and confirmed that it works.

Reference: https://github.com/apache/tvm-ffi/actions/runs/20777613039/job/59667160177

tqchen:
we deliberately deferred it for now to reduce immediate feedback overhead, while we need some posthoc fixes, it works ok for our cases so far

junrushao:
Thanks for bringing up this PR! We used to consider clang-tidy in PRs, but turns out it runs too long, and thus moved it to mainline only. Therefore, we may not want to do so unfortunately

guan404ming:
Thanks for the context. Just to add some context from my side: I think main CI is slow mainly because it runs clang-tidy on all files, while here we only check the modified ones. Curious to hear your thoughts on whether this trade-off would be acceptable. Thanks!

image
(run for 1 file modified)

junrushao:
Seems like a good idea!

.github/workflows/ci_test.yml
```diff
@@ -45,6 +47,13 @@ jobs:
           github_event_name: ${{ github.event_name }}
           pr_base_ref: ${{ github.event.pull_request.base.ref || '' }}
           pr_head_sha: ${{ github.event.pull_request.head.sha || '' }}
+      - name: Get changed C++ files
+        id: cpp_files
+        run: |
+          FILES=$(git diff --name-only --diff-filter=ACMR origin/${{ github.base_ref }}...HEAD -- \
+            '*.c' '*.cc' '*.cpp' '*.cxx' '*.h' '*.hpp' '*.hxx' | tr '\n' ' ')
```
junrushao: We should probably skip header files, because there’s no compilation database generated for headers

guan404ming: Sure, I've removed it. Thanks!

junrushao (approved):
I don't fully understand some of the logics but it seems right. Very happy to get it in and worry about it later if we got into problems :D

guan404ming:
> I don't fully understand some of the logics but it seems right. Very happy to get it in and worry about it later if we got into problems :D

Thanks. Let's see what happen lol, I could handle it in the future if needed.

PR #393, opened 2026-01-09
fix: TypeStr support for Any type in reflection Init

guan404ming:
## Why
- Using refl::init<Any>() to define a constructor that takes Any fails to compile
- The reflection system converts Any to Any&& internally
- No Type2Str<Any&&> specialization existed to handle this case

## How
- Added Type2Str<Any&&> specialization in any.h
- Added forward declaration of Any in type_traits.h

### DIFF
--- a/tests/cpp/test_reflection.cc
+++ b/tests/cpp/test_reflection.cc
@@ -307,6 +307,20 @@ TVM_FFI_STATIC_INIT_BLOCK() {
   refl::ObjectDef<TestObjWithAny>().def(refl::init<Any>()).def_ro("value", &TestObjWithAny::value);
 }
 
+struct TestObjWithAnyView : public Object {
+  Any value;
+  explicit TestObjWithAnyView(AnyView value) : value(value) {}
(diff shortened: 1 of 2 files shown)


tests/cpp/test_reflection.cc
```diff
@@ -294,4 +294,31 @@ TEST(Reflection, AccessPath) {
   auto root_parent = root->GetParent();
   EXPECT_FALSE(root_parent.has_value());
 }
+
+struct TestObjWithAny : public Object {
+  Any value;
+  explicit TestObjWithAny(Any value) : value(std::move(value)) {}
```
junrushao, line 300: could you also try 

```suggestion
  explicit TestObjWithAny(AnyView value) : value(value) {}
```

and see if it compiles?

guan404ming: The code fails to compile with the same missing `TypeStr` error when using `refl::init<AnyView>()` so I also add `Type2Str<AnyView&&>` specialization and related tests.

guan404ming:
Thanks!

PR #394, opened 2026-01-10
doc: Function, Exception and Module

junrushao:


docs/concepts/func_module.rst
```diff
@@ -0,0 +1,584 @@
...
+
+    RAII:
+     // clean up owned resources
+     return err_code;
+   }
+
+Handling Non-C++ Errors
+~~~~~~~~~~~~~~~~~~~~~~~
+
+TVM-FFI preserves frontend errors (e.g., Python exceptions) through a dedicated ``-2`` return path.
```
tqchen: This is not accurate.

For common path, TVM-FFI catches translate python errors to ffi::Error with stacktrace. This allows python stack trace and c++ stack trace to be preserved across multiple cross function calls. 

For rare cases where the python error is triggered by signal handlersg, such error cannot be caught by the c++ frontend, in such case, we raise an ErrorAlreadySet and translate to -2 in the case. see also https://github.com/apache/tvm-ffi/blob/main/include/tvm/ffi/error.h#L73

The code there is the only use-case of -2 return values atm

docs/concepts/func_module.rst
```diff
@@ -0,0 +1,584 @@
...
+     TVMFFISafeCallType safe_call;
+     void* cpp_call;
+   } TVMFFIFunctionCell;
+
+``safe_call`` is used for cross-ABI function calling, which intercepts and stores exceptions in TLS,
+while ``cpp_call`` is used inside the same DSO, and exceptions are thrown directly for performance.
+
+.. important::
+
+  Do not directly use ``safe_call`` or ``cpp_call`` to call functions. Instead, a C API :cpp:func:`TVMFFIFunctionCall`
```
tqchen: It is OK to use `safe_call` directly in low-level abi manipulations, `TVMFFIFunctionCall` was mainly meant as a way to expose the call in cases where function ptr manipulation is not accessible.

docs/concepts/func_module.rst
```diff
@@ -0,0 +1,584 @@
...
+   }
+
+The example above shows how to throw an error in C++. :c:macro:`TVM_FFI_THROW` captures
+the current file name, line number, stacktrace, and error message, and then constructs a
+:cpp:class:`tvm::ffi::Error` object. At ABI boundary, this :cpp:class:`tvm::ffi::Error`
+object is stored in the TLS slot, and the function returns ``-1`` as error code in the C ABI
+:cpp:type:`TVMFFISafeCallType` signature.
+
+.. hint::
+  A detailed implementation of such graceful handling behavior can be found
```
tqchen, line 406: blank line needed in rst

docs/concepts/func_module.rst
```diff
@@ -0,0 +1,584 @@
...
+   result = mod.add_two(40)
+
+Essentially, :py:func:`tvm_ffi.cpp.load` is a shortcut that combines JIT-building the library,
+and then loading the library into a :py:class:`tvm_ffi.Module`.
+
+System Library
+~~~~~~~~~~~~~~
+
+System library modules contain symbols that are statically linked to the current executable.
+This is useful for AOT (ahead-of-time) compilation where functions are bundled with the binary.
```
tqchen: This technique is useful when we would like to preserve module dynamic module loading behavior but do not have dlopen available in the system in a static bundle. Instead, we register the symbols in the system so they can be looked up via tvm_ffi.system_lib

docs/concepts/func_module.rst
junrushao: It's needed for codeblock but not hint block

docs/concepts/func_module.rst
junrushao: Updated

docs/concepts/func_module.rst
```diff
@@ -0,0 +1,564 @@
...
+
+.. code-block:: cpp
+
+   int outer_function(...) {
+     int err_code = 0;
+
+     err_code = inner_function(...);
+     if (err_code != 0) goto RAII;  // Propagate error; TLS has the details
+
+    RAII:
```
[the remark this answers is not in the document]
junrushao: Sorry it's **my** style

junrushao:
@tqchen I made some updates accordingly. Could you please re-review?

PR #395, opened 2026-01-10
feat: Add `tvm_ffi.Function.__init__`

junrushao:
This change enables us to convert Python functions directly to `tvm_ffi.Function` using its constructor. Example:

```python
def add(a: int, b: int) -> int:
    return a + b

func = tvm_ffi.Function(add)
assert isinstance(func, tvm_ffi.Function)
assert func(1, 2) == 3
```


junrushao:
A more general question: if an `tvm_ffi.Object`'s constructor is not defined, say its name is `Foo`, what would `Foo()` do?

guan404ming:
> A more general question: if an tvm_ffi.Object's constructor is not defined, say its name is Foo, what would Foo() do?

Based on my understanding, I think

1. Object container class (FooObj : public Object): FooObj() calls the compiler-generated default constructor, which invokes Object()'s constructor and default-initializes member fields. However, I think this might not be the prper way to create objects. Should we use `make_object<FooObj>()` instead?
2. ObjectRef handle class (Foo : public ObjectRef):
    - If using `TVM_FFI_DEFINE_OBJECT_REF_METHODS_NULLABLE`: Foo() creates a null/undefined reference
    - If using `TVM_FFI_DEFINE_OBJECT_REF_METHODS_NOTNULLABLE`: Compilation error

PR #396, opened 2026-01-10
[RFC][Error][ABI] Update Error to enable future compact to cause chaining

tqchen:
This PR brings a backward compatible update to error ABI to enable possible future support of cause chaining. Specifically, we add two fields:

- cause_chain is an optional field for chaining errors
- extra_context can be used to optionally attach opaque object (e.g. python error) if needed.

The change is backward compatible as we only append to the error field. Most of the existing usages will simply ignore the two fields and use a single error.


tqchen:
mainly for review for now, do not yet merge.

PR #397, opened 2026-01-11
[CMAKE] Allow threads to be not found for cross compilation

tqchen:
This PR updates the threads linkage to be optional so the library can be build for cross-compilation targets where threads may not be available.


tqchen:
Needed for hexagon build

junrushao (approved):

junrushao:
@tqchen Could you adopt Gemini’s recommendation in a follow up PR?

Issue #234, opened 2025-11-07
[Feature Request] Introduce `ffi::Expected<>`


junrushao:
**Goal** Enable an exception-free C++ API by introducing an `ffi::Expected<T>` type (mirroring `std::expected` / Rust `Result`) and wiring it through `ffi::Function` invocation and registration.

**Problem** Current `ffi::Function` calls always return `Any`; failures rely on exceptions. Consumers needing hard real-time or exception-free environments lack a native way to detect errors without exceptions, and we can’t guanratee callbacks are all exception-free.

**Proposal**
- Add `ffi::Expected<T>` where `T` is any type TVM-FFI supports but not `tvm::ffi::Error`; Err is always `ffi::Error`.
- Extend `ffi::Function` invocation helpers so users can request wrappers of type `(Args...) -> ffi::Expected<T>`; calls that produce an `ffi::Error` fill the error branch instead of throwing.
- Allow registering functions that implement `(Args...) -> ffi::Expected<T>`; the dispatcher should convert returned `Expected` objects into the existing error propagation path so current exception-based APIs keep working.
- Document usage patterns, including examples of calling and registering `Expected` functions, and guidance on when to "use exception" vs. `Expected` flows.

tqchen:
Some notes:

- `ffi::Expected<T>` can be effectively implemented like ffi::Variant
   - We forbid T to be ffi::Error  
- We can add ffi::Function::CallExpected (naming of CallExpected can be debated), it simply redirect to the TVMFFISafeCall
```c++
func : ffi::Function = ...;
ffi::Expected<Any> res = func.CallExpected(a0, a1, a2);
```
- Add ffi::Function::FromTypedExpected (similar to FromTyped, naming can be debated)  that takes `(Args...) -> ffi::Expected<T>` and convery into ffi function


At high-level, a `ffi::Function` can be called with normal mode(throws exception) or expected mode(always go through safe call and converts into Expected). The currenr runtime mechanism should enable such interop quite nicely

junrushao:
CC @DarkSharpness if you are interested

Ubospica:
Good topic! FYI, we had an implementation for expected (named Result) in xgrammar: https://github.com/mlc-ai/xgrammar/blob/main/cpp/support/utils.h#L251. Below are several of its design choices, which could be helpful for the design here:
* It forces moving its content in or out of the result object, to eliminate the overhead of copying content
[...]

PR #399, opened 2026-01-11
feat: add ffi::Expected<T> for exception-free error handling

guan404ming:
## Why

 Enable exception-free C++ API similar to Rust's Result or C++23's std::expected, as requested in #234.

 ## How

 - Add `Expected<T>` class holding either success value T or Error
 - Add `Function::CallExpected<T>()` for exception-free function calls
 - Add `is_expected` type trait and Expected handling in `unpack_call`
 - Support registering functions that return `Expected<T>`"

### DIFF
--- a/include/tvm/ffi/function_details.h
+++ b/include/tvm/ffi/function_details.h
@@ -72,16 +72,6 @@ static constexpr bool ArgSupported =
       std::is_same_v<std::remove_const_t<std::remove_reference_t<T>>, AnyView> ||
       TypeTraitsNoCR<T>::convert_enabled));
 
-template <typename T>
-struct is_expected : std::false_type {
-  using value_type = void;
-};
-
-template <typename T>
-struct is_expected<Expected<T>> : std::true_type {
-  using value_type = T;
-};
-
 // NOTE: return type can only support non-reference managed returns
 template <typename T>
 static constexpr bool RetSupported =


tqchen:
Thanks a lot for the contribution, this would benefit from careful reviews, @junrushao @DarkSharpness @Ubospica can you help

Kathryn-cat:
@guan404ming Thanks for the contribution! I will take a closer look sometime this week to provide a round of review.

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,256 @@
...
+  TVM_FFI_INLINE bool has_value() const { return is_ok(); }
+
+  /*!
+   * \brief Access the success value.
+   * \return The success value.
+   * \throws RuntimeError if the Expected contains an error.
+   */
+  TVM_FFI_INLINE T value() const {
+    if (is_err()) {
+      TVM_FFI_THROW(RuntimeError) << "Bad expected access: contains error";
```
Kathryn-cat: If `value()` failed, it appears that it throws an error here but doesn't preserve the original error. Is it possible to retain the original error information?

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return data_.cast<T>();
```
Kathryn-cat, line 138: For large data, we can use move instead of copy, so I agree with Gemini that we can add an overload function here:
```
TVM_FFI_INLINE T value() && {
  if (is_err()) { throw error(); }
  return std::move(data_).cast<T>();
}
```

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,256 @@
...
+   * \brief Check if the Expected contains a success value.
+   * \return True if contains success value, false if contains error.
+   */
+  TVM_FFI_INLINE bool is_ok() const { return data_.as<T>().has_value(); }
+
+  /*!
+   * \brief Check if the Expected contains an error.
+   * \return True if contains error, false if contains success value.
+   */
+  TVM_FFI_INLINE bool is_err() const { return data_.as<Error>().has_value(); }
```
Kathryn-cat: It is likely that `ObjectRef` as a base class of `Error` can have both `is_ok()=True` and `is_err()=True`, since `CheckAnyStrict` relies on inheritance checking, not type matching. I would suggest returning `!is_ok()` here.

Kathryn-cat: Also it would be helpful to enhance the test cases on this.

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,256 @@
...
+    if (is_err()) {
+      TVM_FFI_THROW(RuntimeError) << "Bad expected access: contains error";
+    }
+    return data_.cast<T>();
+  }
+
+  /*!
+   * \brief Access the error value.
+   * \return The error value.
+   * \note Behavior is undefined if the Expected contains a success value.
```
Kathryn-cat: Let's instead of saying "behavior is undefined", say it throws a RuntimeError on bad access.

Perhaps change the function body into 
```
  if (!is_err()) {
    TVM_FFI_THROW(RuntimeError) << "Bad expected access: contains value, not error";
  }
```
to mimic `value()` impl.

Kathryn-cat (changes requested):

include/tvm/ffi/expected.h
guan404ming: Done. Changed value() to throw the contained error directly instead of a generic RuntimeError:                                                               
```
if (is_err()) {                                                                                                                                              
  throw data_.cast<Error>();                                                                                                                                 
}  
```

include/tvm/ffi/expected.h
guan404ming: Agree as well. I've added both const& and && qualified overloads for value():

include/tvm/ffi/expected.h
guan404ming: Changed is_err() to return !is_ok() and also fixed is_ok() to check for Error first.

include/tvm/ffi/expected.h
guan404ming: Sure, I've updated the docstring and changed from TVM_FFI_ICHECK to TVM_FFI_THROW(RuntimeError)

include/tvm/ffi/expected.h
Kathryn-cat: Actually we might not need both here, @tqchen would like to hear your opinion if we wanna keep one of them

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+ */
+template <typename T>
+TVM_FFI_INLINE Expected<T> ExpectedOk(T value) {
+  return Expected<T>::Ok(std::move(value));
+}
+
+/*!
+ * \brief Helper function to create Expected::Err.
+ * \param error The error value.
+ * \return Expected<Any> containing the error.
```
Kathryn-cat: Let's fix the comment here, should return `Expected<T>` same as the ExpectedOk case.

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+ * \note Returns Expected<Any> to allow usage in contexts where T is inferred.
+ */
+template <typename T = Any>
```
Kathryn-cat: Perhaps `template <typename T>` is cleaner? Do we have to default to `Any` here

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+    return std::move(data_).template cast<Error>();
+  }
+
+  /*!
+   * \brief Get the success value or a default value.
+   * \param default_value The value to return if Expected contains an error.
+   * \return The success value if present, otherwise the default value.
+   */
+  template <typename U = std::remove_cv_t<T>>
+  TVM_FFI_INLINE T value_or(U&& default_value) const {
```
Kathryn-cat, line 163: I think this function might be rarely used, because the philosophy of `Expected<T>` is handling error explicitly. Let's leave it for discussion here.

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+// TypeTraits specialization for Expected<T>
+template <typename T>
+inline constexpr bool use_default_type_traits_v<Expected<T>> = false;
+
+template <typename T>
+struct TypeTraits<Expected<T>> : public TypeTraitsBase {
+  TVM_FFI_INLINE static void CopyToAnyView(const Expected<T>& src, TVMFFIAny* result) {
+    const TVMFFIAny* src_any = reinterpret_cast<const TVMFFIAny*>(&src.data_);
+    if (TypeTraits<T>::CheckAnyStrict(src_any)) {
+      TypeTraits<T>::MoveToAny(TypeTraits<T>::CopyFromAnyViewAfterCheck(src_any), result);
```
Kathryn-cat: We possibly don't need to create a copy of T then move it.

Possible simplification:
```
  if (src.is_err()) {
    TypeTraits<Error>::CopyToAnyView(src.error(), result);
  } else {
    TypeTraits<T>::CopyToAnyView(src.value(), result);
  }
```

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+    } else {
+      TypeTraits<Error>::MoveToAny(TypeTraits<Error>::CopyFromAnyViewAfterCheck(src_any), result);
+    }
+  }
+
+  TVM_FFI_INLINE static void MoveToAny(Expected<T> src, TVMFFIAny* result) {
+    TVMFFIAny* src_any = reinterpret_cast<TVMFFIAny*>(&src.data_);
+    if (TypeTraits<T>::CheckAnyStrict(src_any)) {
+      TypeTraits<T>::MoveToAny(TypeTraits<T>::MoveFromAnyAfterCheck(src_any), result);
```
Kathryn-cat: ditto,
```
  if (src.is_err()) {
    TypeTraits<Error>::MoveToAny(std::move(src).error(), result);
  } else {
    TypeTraits<T>::MoveToAny(std::move(src).value(), result);
  }
```

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+  TVM_FFI_INLINE static std::string GetMismatchTypeInfo(const TVMFFIAny* src) {
+    return TypeTraitsBase::GetMismatchTypeInfo(src);
+  }
+
+  TVM_FFI_INLINE static std::string TypeStr() {
+    return "Expected<" + TypeTraits<T>::TypeStr() + ">";
+  }
+
+  TVM_FFI_INLINE static std::string TypeSchema() {
+    return R"({"type":"Expected","args":[)" + details::TypeSchema<T>::v() + "]}";
```
Kathryn-cat: Maybe also mention `{"type":"ffi.Error"}` in type schema?

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,239 @@
...
+    if (auto opt = TypeTraits<T>::TryCastFromAnyView(src)) {
+      return Expected<T>::Ok(*std::move(opt));
+    }
+    if (auto opt_err = TypeTraits<Error>::TryCastFromAnyView(src)) {
+      return Expected<T>::Err(*std::move(opt_err));
+    }
+    return std::nullopt;
+  }
+
+  TVM_FFI_INLINE static std::string GetMismatchTypeInfo(const TVMFFIAny* src) {
```
Kathryn-cat: Let's remove this

include/tvm/ffi/expected.h
guan404ming: Thanks, this is more clearer.

include/tvm/ffi/expected.h
guan404ming: Yes, that would be better. Thanks.

include/tvm/ffi/expected.h
guan404ming: I think maybe we should keep this since that is kind of standard practice std::optional and C++23 std::expected both have value_or.

ref [std::expected (C++23)](https://en.cppreference.com/w/cpp/utility/expected/value_or.html) and [std::optional (C++17)](https://en.cppreference.com/w/cpp/utility/optional/value_or)

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,51 @@ class Function : public ObjectRef {
...
+    Any result;
+    FunctionObj* func_obj = static_cast<FunctionObj*>(data_.get());
+
+    // Use safe_call path to catch exceptions
+    int ret_code = func_obj->safe_call(func_obj, reinterpret_cast<const TVMFFIAny*>(args_pack),
+                                       kNumArgs, reinterpret_cast<TVMFFIAny*>(&result));
+
+    if (ret_code == 0) {
+      // Success - cast result to T and return Ok
+      return Expected<T>::Ok(std::move(result).cast<T>());
```
Kathryn-cat: Here's a tricky part. If the result does not match the type T, `cast<T>()` will throw, and this error will escape CallExpected.

Some proposed change: (feel free to brainstorm better ones)
```
if (ret_code == 0) {
  if (auto val = std::move(result).template as<T>()) {
    return Expected<T>::Ok(std::move(*val));
  } else {
    return Expected<T>::Err(Error("TypeError", 
        "CallExpected: result type mismatch, expected " + TypeTraits<T>::TypeStr(), ""));
  }
}
```

include/tvm/ffi/function_details.h
```diff
@@ -67,10 +72,23 @@ static constexpr bool ArgSupported =
...
+
+template <typename T>
+inline constexpr bool is_expected_v = is_expected<T>::value;
+
 // NOTE: return type can only support non-reference managed returns
 template <typename T>
-static constexpr bool RetSupported =
-    (std::is_same_v<T, Any> || std::is_void_v<T> || TypeTraits<T>::convert_enabled);
+static constexpr bool RetSupported = (std::is_same_v<T, Any> || std::is_void_v<T> ||
+                                      TypeTraits<T>::convert_enabled || is_expected_v<T>);
```
Kathryn-cat: I think `convert_enabled` is already True here, `is_expected_v` might be redundant

include/tvm/ffi/function_details.h
```diff
@@ -219,6 +237,14 @@ TVM_FFI_INLINE void unpack_call(std::index_sequence<Is...>, const std::string* o
   // use index sequence to do recursive-less unpacking
   if constexpr (std::is_same_v<R, void>) {
     f(ArgValueWithContext<std::tuple_element_t<Is, PackedArgs>>{args, Is, optional_name, f_sig}...);
+  } else if constexpr (is_expected_v<R>) {
+    R expected_result = f(ArgValueWithContext<std::tuple_element_t<Is, PackedArgs>>{
+        args, Is, optional_name, f_sig}...);
+    if (expected_result.is_ok()) {
+      *rv = expected_result.value();
+    } else {
+      throw expected_result.error();
```
Kathryn-cat: Perhaps use move?
```
if (expected_result.is_ok()) {
  *rv = std::move(expected_result).value();
} else {
  throw std::move(expected_result).error();
}
```

guan404ming: Sure, updated!

include/tvm/ffi/function_details.h
guan404ming: You're right, let me remove it.

include/tvm/ffi/function.h
guan404ming: Nice catch. The implementation looks good. I think there is a potential edge case that we probably don't want ot try casting if `T = Any`. Here is updated implementation

```c
if (ret_code == 0) {
  if constexpr (std::is_same_v<T, Any>) {
    return Expected<T>::Ok(std::move(result));
  } else {
    if (auto val = std::move(result).template as<T>()) {
      return Expected<T>::Ok(std::move(*val));
    } else {
      return Expected<T>::Err(Error(
          "TypeError",
          "CallExpected: result type mismatch, expected " + TypeTraits<T>::TypeStr(), ""));
    }
  }
}
```

guan404ming:
> Since it introduces new function calling API to the users, @guan404ming could you document the usage?

Sure, I think I could do it as a follow up after this one!

Kathryn-cat:
@tqchen the CI fails because tvm-ffi's DLPack version is 1.0.2, while dlpack exchange api is tested against PyTorch's native impl (DLPack 1.0.3), perhaps we need to upgrade tvm-ffi's DLPack version

Kathryn-cat:
> > Since it introduces new function calling API to the users, @guan404ming could you document the usage?
> 
> Sure, I think I could do it as a follow up after this one!

Thanks @guan404ming , I think the PR is in good state, let's wait for CI to be fixed and get it in!

guan404ming:
Here is the upgrade pr https://github.com/apache/tvm-ffi/pull/420. I've checked there is no any compat issue need to be updated in our src code.

guan404ming:
The ci passed after the upgrade!

include/tvm/ffi/function_details.h
```diff
@@ -219,6 +237,14 @@ TVM_FFI_INLINE void unpack_call(std::index_sequence<Is...>, const std::string* o
   // use index sequence to do recursive-less unpacking
   if constexpr (std::is_same_v<R, void>) {
     f(ArgValueWithContext<std::tuple_element_t<Is, PackedArgs>>{args, Is, optional_name, f_sig}...);
+  } else if constexpr (is_expected_v<R>) {
```
tqchen: This behavior can be a bit confusing. If the ffi.Function is explicitly returning Expected Value (instead of throw using the error handling mechanism, then the function should successully return  instead of implicitly throw when error is found?

Mayb need a regression testcase for this

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,234 @@
...
+};
+
+/*!
+ * \brief Helper function to create Expected::Ok with type deduction.
+ * \tparam T The success type (deduced from argument).
+ * \param value The success value.
+ * \return Expected<T> containing the success value.
+ */
+template <typename T>
+TVM_FFI_INLINE Expected<T> ExpectedOk(T value) {
```
tqchen: I think it is good to do round of API review, can you list all the APIs that are not conforming to std::expected, list their names, and discuss choice(i know some comes from rust API style, but good to be explicit). Would be good to list the APIs in the comment

tqchen:
Thanks @guan404ming , i think we are close, i have one behavior comment. Besides that, given this is new public API, i think it would be helpful to do a round of quick API review, can you provide a list of APIs(including methods), would be good to categorize them as:

- C0: strictly follows c++ standard
- C1: others and list rationale(e.g. we follows rust API)

Would also be good to discuss move and value copy behavior.

include/tvm/ffi/function_details.h
Kathryn-cat: i think the current design unwraps `Expected<T>` at the FFI boundary. `CallExpected` use safe_call to catch the error and return `Expected<T>`; on the contrary, the normal call path would throw an error. This might seem a bit convoluted. Do you have better suggested ideas?

tqchen: There are two ways that Expected<T> get returned

- W0: Expected<T> get "raised" internally by setting the TLS value via TVMFFISetRaised
- W1: Expected<T> get returned as a normal return value

And there are two ways to call a function now
- C0: Normal call
- C1: CallExpected

Would be good to discuss the overall relation in the mix of four cases

- C0+ W0: we need to throw the error to caller
- C0 + W1: no error should be thrown, the Expected<T>  should be returned to the caller
- C1 + W0: we should return Expected<Any> with the error "raised" set to the returned Expected value
- C1 + W1: we need to return the Expected<T>, note that in this case it is harder to distinguish error being returned versus error being raised if the desirable logic is to return an error, but this is more of a rare case.

In any case, it would be good first to make sure behavior of C0 is correct, and the particular context seems to suggest C0+W1, in such case, we should return the value to the caller.

guan404ming: Thanks for the detailed breakdown of the four scenarios (C0/C1 + W0/W1) which really helped clarify the design considerations. Based on my understanding, the key question is about the C0 + W1 scenario (normal Call() on a function returning Expected<T>). Here's how I see the two possible directions: 

Option A: Keep current behavior (auto-unwrap)   
``` c                                                                                                                              
  // Current implementation                                                                                                                                                      
  if (expected_result.is_ok()) {                                                                                                                                                 
      *rv = std::move(expected_result).value();  // unwrap value                                                                                                                 
  } else {                                                                                                                                                                       
      throw std::move(expected_result).error();  // throw error                                                                                                                  
  }  
  ```                                                                                                                                                                            
  - Pros: Caller uses Call<T>() and gets the value directly                                                                                                                      
  - Cons: Cannot distinguish "returned error" from "raised error"                                                                                                                
                                                                                                                                                                                 
  Option B: No unwrap, return Expected directly                                                                                                                                  
  ```c
  // Proposed change                                                                                                                                                             
  *rv = f(...);  // return Expected as-is                                                                                                                                        
  ```
  - Pros: Clear distinction between returned vs raised errors                                                                                                                    
  - Cons: Caller must use Call<Expected<T>>() and handle it explicitly 
  
Personally, I'd slightly lean toward Option B after @tqchen's breakdown. If a function explicitly chooses to return Expected, I think we should respect that and let the caller receive it directly.                                                                                                                                             

I'd like to confirm your preference before updating the implementation. Happy to add regression tests or update implementation for whichever approach we go with.

Comment (tqchen) [reply include/tvm/ffi/function_details.h]: agree option B is better

**Comment by guan404ming** [comment]
Here is the API review

```
| API                | Category | Notes                                                              |                                                                         
|--------------------|----------|--------------------------------------------------------------------|                                                                         
| is_ok()            | C0       | Follows std::expected::has_value() semantics                       |                                                                         
| is_err()           | C1       | Rust-style naming; equivalent to !has_value()                      |                                                                         
| has_value()        | C0       | Alias for is_ok(), follows C++ standard                            |                                                                         
| value()            | C0       | Follows std::expected::value()                                     |                                                                         
| value() &&         | C1       | Move-qualified overload for efficiency (not in C++23 standard)     |                                                                         
| error()            | C0       | Follows std::expected::error()                                     |                                                                         
| error() &&         | C1       | Move-qualified overload for efficiency                             |                                                                         
| value_or(default)  | C0       | Follows std::expected::value_or()                                  |                                                                         
| ExpectedOk(value)  | C1       | Rust-style helper; C++ standard has no equivalent factory function |                                                                         
| ExpectedErr(error) | C1       | Rust-style helper; C++ standard has no equivalent factory function |                                                                         
```

Rationale for C1 APIs:                                                                                                                                                         
- Move-qualified overloads (&&): Added for performance when the Expected is a temporary, avoiding unnecessary copies of large objects.                                         
- ExpectedOk() / ExpectedErr(): Rust-style factory functions for cleaner syntax with type deduction. C++23 std::expected relies on constructors instead.                       
- is_err(): Rust naming convention; provides symmetry with is_ok().


APIs not implemented (compared to C++23 std::expected):                                                                                                                        
- operator bool(): can use has_value() or is_ok() instead                                                                                                                     
- operator*() / operator->(): unchecked access, omitted for safety                                                                                                            
- Monadic operations (and_then, transform, or_else, transform_error): may add in future if needed

File: include/tvm/ffi/function_details.h (hunk 15 above)
Comment (guan404ming) [reply include/tvm/ffi/function_details.h]: Updated with option B. Please take another look, thanks!

**Comment by tqchen** [comment]
thanks @guan404ming , cc @Ubospica @Seven-Streams @junrushao @DarkSharpness for quick check on APIs. the ask

- please comment on the API naming/choice, especially C1 APIs.
- help cross check needs in possible downstream (@Ubospica would be good to confirm xgrammar needs)

**Comment by DarkSharpness** [comment]
> Here is the API review
> 
> ```
> | API                | Category | Notes                                                              |                                                                         
> |--------------------|----------|--------------------------------------------------------------------|                                                                         
> | is_ok()            | C0       | Follows std::expected::has_value() semantics                       |                                                                         
> | is_err()           | C1       | Rust-style naming; equivalent to !has_value()                      |                                                                         
> | has_value()        | C0       | Alias for is_ok(), follows C++ standard                            |                                                                         
> | value()            | C0       | Follows std::expected::value()                                     |                                                                         
> | value() &&         | C1       | Move-qualified overload for efficiency (not in C++23 standard)     |                                                                         
> | error()            | C0       | Follows std::expected::error()                                     |                                                                         
> | error() &&         | C1       | Move-qualified overload for efficiency                             |                                                                         
> | value_or(default)  | C0       | Follows std::expected::value_or()                                  |                                                                         
> | ExpectedOk(value)  | C1       | Rust-style helper; C++ standard has no equivalent factory function |                                                                         
> | ExpectedErr(error) | C1       | Rust-style helper; C++ standard has no equivalent factory function |                                                                         
> ```
> 
> Rationale for C1 APIs:
> 
> * Move-qualified overloads (&&): Added for performance when the Expected is a temporary, avoiding unnecessary copies of large objects.
> * ExpectedOk() / ExpectedErr(): Rust-style factory functions for cleaner syntax with type deduction. C++23 std::expected relies on constructors instead.
> * is_err(): Rust naming convention; provides symmetry with is_ok().
> 
> APIs not implemented (compared to C++23 std::expected):
> 
> * operator bool(): can use has_value() or is_ok() instead
> * operator*() / operator->(): unchecked access, omitted for safety
> * Monadic operations (and_then, transform, or_else, transform_error): may add in future if needed

Thanks for the detailed API review.

One small difference worth calling out for `ExpectedOk` / `ExpectedErr` is a C++-specific type deduction behavior compared to Rust. In Rust, generic parameters can be inferred from the return type, while in C++ template argument deduction only considers function arguments, not the return type. This means that with factory-style helpers like `ExpectedOk`, the deduced type may sometimes differ from the desired return type. For example:

```C++
template <typename T>
Expect <T> ExpectOk(T value);

Expect<long> function(int x) {
  return ExpectOk(x); // <- in many cases, we want [T = long], which deduced from return type.
                      // but in this case T will deduced as int, which may lead to compile error.
}
```

In such cases, C++ cannot deduce `T = long` from the return type, which may lead to a compilation error.

I'm not sure whether this will cause some inconvenience in real usage, but the current design is reasonable for simplicity.

In XGrammar we implement something like `PartialExpected`, which can be implicitly converted to any `Expected<T>` and solves the return type inference problem. However, this actually introduces much complexity https://github.com/mlc-ai/xgrammar/blob/aa44ded202fdaa2cb55f48deb31d17c9822a8a1f/cpp/support/utils.h#L130-L212

guan404ming:
Thanks for bringing up the discussion! I agree with keeping the design simple. While return type deduction is nice, explicit type specification is a fair trade-off to keep the API clean.

I think type deduction works for most cases. For exceptions, users can simply use like `Expected<long>::Ok(x)`. For a foundational library like `tvm-ffi`, the complexity of `PartialExpected` isn't worth the marginal benefit in my opinion. This strikes the right balance between simplicity and usability.

tqchen:
just want to make sure xgrammar can switch to make use of the new impl here. once you think the remaining apis are good please explicitly approve,  @Ubospica @Seven-Streams @junrushao @DarkSharpness

tqchen:
Thinking a bit more about partial deduction, it might be useful to think about the usages. 

The main principle we like to follow to deliberate the API design is mainly because of stability, so we want to have one consistent behavior once introduced(less suprising to users), or not introducing the API so users can do their own wrapping. We want to avoid the case where api causes inconsistency with standard usages, or find need to update the api spec later, that is why aligning with standard is good, and having people to chime in is always better.

Thanks a lot for working through this deliberation process. Specifically,  it is good to think about what are the expected usage pattern of ExpectedOk/ExpectedErr, ideally we want to see real code snippest from examples like xgrammar or simply ask genai to provide some.

The main thing to consider is consistency with `std::optional`, the following code is generally a common pattern:

```c++
std::optional<long> func_optional(bool valid, int x) {
   if (valid) return x;
   return std::nullopt;
}
```

We can find that return value conversion from x to `std::optional<long>` is automatically deduced by the return type. 

So what @DarkSharpness mentioned is that it is desirable to have similar syntax. In which case ExpectedOk and ExpectedErr follows the similar rationale of targetted usage. 

```c++
Expected<long> func_optional_rvalue_deduce(bool valid, int x) {
   if (value) return ExpectedOk(x);
   return ExpectedErr(Error("ValueError", "xx"));
}
```

In the current impl, we need to write it as
```c++
Expected<long> func_optional_param_deduce(bool valid, int x) {
   if (value) return Expected<long>(x);
   return Expected<long>(Error("ValueError", "xx"));
}
```
And for ExpectedErr seems we always have to write ExpectedErr<T> because T cannot be deduced from input type

The main question we want to ask is the common usage scenarios the sugar ExpectedOk/ExpectedErr, e.g. are they mostly used in the case similar to `func_optional` where we want to simplify by deducing from return values, if so, then I think the deferred deduction is wortwhile, alternatively we should left out the  ExpectedOk/ExpectedErr sugar since for most common case we expect people to explicitly write `Expected<T>`

guan404ming:
Thanks for the feedback! Sorry about the late, spending some time to figure out whole things. After reviewing the design options and looking at real-world patterns, I agree that we should support implicit conversion. Here is some of my note to clarify all thing:

### Pros and Cons

| Approach | Success Return | Error Return | Pros | Cons |
|----------|---------------|--------------|------|------|
| Factory functions (current) | `return ExpectedOk(x);` | `return ExpectedErr<T>(err);` | Explicit intent | Asymmetric API - Ok deduces type, Err requires explicit `<T>` |
| Static methods only | `return Expected<T>::Ok(x);` | `return Expected<T>::Err(err);` | No extra functions | Verbose, type repeated |
| Implicit conversion | `return x;` | `return Error(...);` | Concise, symmetric, follows C++ conventions | Slightly less explicit |

### C++ Standard Precedent

From [cppreference std::expected](https://en.cppreference.com/w/cpp/utility/expected):

>`std::expected<T,E>` provides implicit conversion from `T` for the success case, and uses `std::unexpected<E>` for errors.

Similarly, `std::optional<T>` allows implicit conversion from `T`.

### XGrammar Current Pattern

Looking at [XGrammar's exception.h](https://github.com/mlc-ai/xgrammar/blob/main/include/xgrammar/exception.h), they currently use traditional C++ exceptions:

```cpp
// Current XGrammar (exceptions)
Grammar Grammar::FromJSON(const std::string& json) {
    if (!valid) throw InvalidJSONError("...");
    return grammar;
}

// With Expected + implicit conversion
Expected<Grammar> Grammar::FromJSON(const std::string& json) {
    if (!valid) return Error("InvalidJSONError", "...");
    return grammar;
}
```

The implicit conversion approach makes migration from exception-based code more natural.

### Implementation

I think to add implicit converting constructors would be ideal to handle this feat:

```cpp
template <typename T>
class Expected {
 public:
  Expected(T value) : data_(Any(std::move(value))) {}      // implicit from T
  Expected(Error error) : data_(Any(std::move(error))) {}  // implicit from Error
  // ...
};
```

This eliminates the asymmetry issue where `ExpectedOk` deduces type but `ExpectedErr<T>` requires explicit specification. In this way, we then could remove `ExpectedOk()`, `ExpectedErr()`  factory functions and `Expected<T>::Ok()`, `Expected<T>::Err()` static methods accordingly. Does this approach sound reasonable? Happy to discuss if there are any concerns or alternative suggestions.

Ubospica:
Thanks for the thorough discussion! The overall implementation looks good to me. API-wise, I agree it would be great to follow C++ conventions. Adding implicit conversion looks like a reasonable approach.

I also checked XGrammar's implementation of Expected (called Result, at https://github.com/Ubospica/xgrammar/blob/15dc61d02ed24820215ca1d491d6a07ba32fff50/cpp/support/utils.h#L251). It's more like a C++ imitation of the Result in rust. There is one difference in usage: It allows enum-based error.

```
enum class SchemaErrorType : int {
  kInvalidSchema = 0,
  kUnsatisfiableSchema = 1,
};

using SchemaError = TypedError<SchemaErrorType>;

Result<SchemaSpecPtr, SchemaError> Parse(
      const picojson::value& schema, const std::string& rule_name_hint = "root"
);
```

`std::expected` also supports it through customizing E to an enum class in `std::expected<T, E>`.

We currently restrict the error kind to `tvm::ffi::Error`. I think the design is good because it simplifies the error handling. Also, it allows customizing the kind through its `std::string kind` parameter.

In comparison, the enum-based error would be more efficient, but I don't think it's a blocker. To support this, we may consider supporting errno in the Error class in the future.

Another consideration is whether we should expose ffi::expected to Python. I feel it is nice to have in the future. For this PR, it may not be required to simplify the scope.

Seven-Streams:
Currently, in XGrammar's result library, [`ResultOk(...)` and `ResultErr(...)`](https://github.com/Ubospica/xgrammar/blob/15dc61d02ed24820215ca1d491d6a07ba32fff50/cpp/support/utils.h#L124-L212) are used to return success or error values, and I think it is easy to use. In addition, I also agree that implicit conversion is a good choice, since it follows [the examples in cppreference](https://en.cppreference.com/w/cpp/utility/expected.html) well.

guan404ming:
> In comparison, the enum-based error would be more efficient, but I don't think it's a blocker. To support this, we may consider supporting errno in the Error class in the future. Another consideration is whether we should expose ffi::expected to Python. I feel it is nice to have in the future. For this PR, it may not be required to simplify the scope.

Thanks for the detailed review and the XGrammar comparison! The enum-based error and Python exposure are both good ideas worth exploring. I think we can keep them as follow-up items to keep this PR focused on the core Expected<T> with implicit conversion support.

> Currently, in XGrammar's result library, [ResultOk(...) and ResultErr(...)](https://github.com/Ubospica/xgrammar/blob/15dc61d02ed24820215ca1d491d6a07ba32fff50/cpp/support/utils.h#L124-L212) are used to return success or error values, and I think it is easy to use.

The current implementation removes `ExpectedOk/ExpectedErr` since implicit conversion makes them redundant. But given that XGrammar uses a similar pattern (ResultOk/ResultErr), I want to confirm whether we prefer to keep both options available for users who favor the explicit style.

> I think to add implicit converting constructors would be ideal to handle this feat:

Just updated with implicit conversion as planned please take another look thanks!

tqchen:
One minor thing on consistency with c++, i think it might makes sense to have Unexpected(Error()) so we are being explicit that this is an Unexpected value

guan404ming:
> One minor thing on consistency with c++, i think it might makes sense to have Unexpected(Error()) so we are being explicit that this is an Unexpected value

Thanks for the nice catch, just updated with `Unexpected`

include/tvm/ffi/expected.h
Ubospica: I feel it's fine to have both, since their usage would be different: `std::move(expected).value()` (for the move overload) vs `expected.value()` (for the const& overload), and this also follows C++ standard.

Ubospica (approved):
The new changes look great to me. Thank you so much!

guan404ming:
> The new changes look great to me. Thank you so much!

Thanks for your review!

include/tvm/ffi/expected.h
```diff
@@ -0,0 +1,218 @@
...
+
+/*!
+ * \brief Wrapper to explicitly construct an Expected in the error state.
+ */
+struct Unexpected {
+  /*! \brief Construct from an error. */
+  // NOLINTNEXTLINE(google-explicit-constructor,runtime/explicit)
+  Unexpected(Error error) : error_(std::move(error)) {}
+  /*! \brief The wrapped error. */
+  Error error_;
```
tqchen: consider align with `std::unexpected<E>` API, make it a class and keep it an private member. The error can be accessed via error() member function https://en.cppreference.com/w/cpp/utility/expected/unexpected.html

Would be good to see if we can use Unexpected<E> for E as subclass of Error if such auto deduction is possible.

tqchen:
Some final comments, @Kathryn-cat @DarkSharpness @Seven-Streams  please take another look

include/tvm/ffi/expected.h
guan404ming: It will be more consistent in this way, thanks. I just updated with your suggestion.

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,64 @@ class Function : public ObjectRef {
...
+        if (result.template as<Error>().has_value()) {
+          return std::move(result).template cast<Error>();
+        }
+        // Try to extract as T
+        if (auto val = std::move(result).template as<T>()) {
+          return std::move(*val);
+        }
+        return Error("TypeError",
+                     "CallExpected: result type mismatch, expected " + TypeTraits<T>::TypeStr(),
+                     "");
```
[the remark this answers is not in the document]
guan404ming: This one makes sense to me, just updated.

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,65 @@ class Function : public ObjectRef {
...
+      } else {
+        // Check if result is Error (from Expected-returning function that returned Err)
+        if (result.template as<Error>().has_value()) {
+          return std::move(result).template cast<Error>();
+        }
+        // Try to extract as T
+        if (auto val = result.template try_cast<T>()) {
+          return std::move(*val);
+        }
+        return Error("TypeError",
```
tqchen: I know it is a bit more verbose, but i think it is helpful to always use Unexpected(Errror()), and ban implicit conversion from Error

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,65 @@ class Function : public ObjectRef {
...
+      if constexpr (std::is_same_v<T, Any>) {
+        return std::move(result);
+      } else {
+        // Check if result is Error (from Expected-returning function that returned Err)
+        if (result.template as<Error>().has_value()) {
+          return std::move(result).template cast<Error>();
+        }
+        // Try to extract as T
+        if (auto val = result.template try_cast<T>()) {
+          return std::move(*val);
```
tqchen: move optional bvalue should be `*std::move(val)`, would be good to confirm with genai

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,65 @@ class Function : public ObjectRef {
...
+    // Use safe_call path to catch exceptions
+    int ret_code = func_obj->safe_call(func_obj, reinterpret_cast<const TVMFFIAny*>(args_pack),
+                                       kNumArgs, reinterpret_cast<TVMFFIAny*>(&result));
+
+    if (ret_code == 0) {
+      if constexpr (std::is_same_v<T, Any>) {
+        return std::move(result);
+      } else {
+        // Check if result is Error (from Expected-returning function that returned Err)
+        if (result.template as<Error>().has_value()) {
```
tqchen: T should take higher priority than Error as that is the fast path, always hold the value as optional first and move that value, avoid double has value checking

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,65 @@ class Function : public ObjectRef {
...
+                         ", but got " + result.GetTypeKey(),
+                     "");
+      }
+    } else if (ret_code == -2) {
+      // Environment error already set (e.g., Python KeyboardInterrupt)
+      // We still throw this since it's a signal, not a normal error
+      throw ::tvm::ffi::EnvErrorAlreadySet();
+    } else {
+      // Error occurred - retrieve from safe call context and return Err
+      return details::MoveFromSafeCallRaised();
```
tqchen: Explicitly use Unexpected

include/tvm/ffi/function_details.h
```diff
@@ -219,6 +237,9 @@ TVM_FFI_INLINE void unpack_call(std::index_sequence<Is...>, const std::string* o
...
+    *rv = f(ArgValueWithContext<std::tuple_element_t<Is, PackedArgs>>{args, Is, optional_name,
```
tqchen: looks like this is not necessary, please cross check

include/tvm/ffi/function_details.h
```diff
@@ -67,6 +72,19 @@ static constexpr bool ArgSupported =
...
+  using value_type = void;
+};
+
+template <typename T>
+struct is_expected<Expected<T>> : std::true_type {
+  using value_type = T;
+};
+
+template <typename T>
+inline constexpr bool is_expected_v = is_expected<T>::value;
```
tqchen: not sure if this is needed anymore given the new convention

include/tvm/ffi/function.h
```diff
@@ -643,6 +643,65 @@ class Function : public ObjectRef {
...
+          return std::move(*val);
+        }
+        return Error("TypeError",
+                     "CallExpected: result type mismatch, expected " + TypeTraits<T>::TypeStr() +
+                         ", but got " + result.GetTypeKey(),
+                     "");
+      }
+    } else if (ret_code == -2) {
+      // Environment error already set (e.g., Python KeyboardInterrupt)
+      // We still throw this since it's a signal, not a normal error
```
tqchen: i think explicitly throw here is not as good since it breaks the assumption of no exception. Actually we find it is better to further unify things a bit, i made https://github.com/apache/tvm-ffi/pull/425 so we don't have to handle -2

include/tvm/ffi/function_details.h
guan404ming: It did, just removed.

include/tvm/ffi/function.h
guan404ming: All error returns now use explicit `Unexpected()` wrapper instead of implicit Error conversion, which I think would be more helpful!

include/tvm/ffi/function.h
guan404ming: Verified and changed to `*std::move(val)`

include/tvm/ffi/function.h
guan404ming: Reordered to check T first (fast path), then Error. Now uses try_cast for both to avoid double checking.

include/tvm/ffi/function.h
guan404ming: Just removed, thanks for the pr!

include/tvm/ffi/function_details.h
guan404ming: I think the trait is still used in test assertions (test_expected.cc:213) to verify Expected type detection, so I keeping it for now.

tqchen: i see, it is worthwhile minimizing the code mainly because is_expected_v may be a future std type, the test uscase seems to be a duplication, so still good to remove here

tests/cpp/test_expected.cc
```diff
@@ -0,0 +1,347 @@
...
+      return Error("ValueError", "Division by zero", "");
+    }
+    return a / b;
+  };
+
+  // Verify the FunctionInfo extracts Expected<int> as return type
+  using FuncInfo = tvm::ffi::details::FunctionInfo<decltype(safe_divide)>;
+  static_assert(std::is_same_v<FuncInfo::RetType, Expected<int>>,
+                "Return type should be Expected<int>");
+  static_assert(tvm::ffi::details::is_expected_v<FuncInfo::RetType>,
```
tqchen: this duplicates with the is_same_v assertion on top

include/tvm/ffi/function_details.h
guan404ming: Make sense to me. Just removed!

include/tvm/ffi/function_details.h
```diff
@@ -67,6 +72,16 @@ static constexpr bool ArgSupported =
       std::is_same_v<std::remove_const_t<std::remove_reference_t<T>>, AnyView> ||
       TypeTraitsNoCR<T>::convert_enabled));
 
+template <typename T>
+struct is_expected : std::false_type {
```
tqchen: same goes for is_expected, let us remove this, i think we are good to go

guan404ming: Thanks, just removed!

guan404ming:
Thanks for all reviews and input!

PR #410, opened 2026-01-13
[TEST] Introduce release extra tests

tqchen:
This PR introduces release extra tests to help run some integration of downstream packages, making it easy to run quick checks during release.


junrushao:
I do think it's a good idea to test out those unittests per commit on mainline. Just need some test machines

PR #411, opened 2026-01-14
feat: Add __repr__ generation support for @c_class dataclasses

Dayuxiaoshui:
- Add `repr` parameter to `c_class()` decorator (default: True)
- Add `repr` parameter to `field()` function (default: True)
- Implement `method_repr()` to generate __repr__ methods
- Generated repr format: ClassName(field1=value1, field2=value2, ...)
- Fields with repr=False are excluded from the representation

This implements part of #356 dataclass feature parity.

### DIFF
--- a/python/tvm_ffi/dataclasses/_utils.py
+++ b/python/tvm_ffi/dataclasses/_utils.py
@@ -131,13 +131,15 @@ def method_repr(type_cls: type, type_info: TypeInfo) -> Callable[..., str]:
         body_lines = [f"return f'{type_cls.__name__}()'"]
     else:
         # Build field representations
-        fields_str = ", ".join(f"{field_name}={{self.{field_name}!r}}" for field_name in repr_fields)
+        fields_str = ", ".join(
+            f"{field_name}={{self.{field_name}!r}}" for field_name in repr_fields
+        )
         body_lines = [f"return f'{type_cls.__name__}({fields_str})'"]
 
     source_lines = ["def __repr__(self) -> str:"]
     source_lines.extend(f"    {line}" for line in body_lines)
     source = "\n".join(source_lines)
-    
+
     # Note: Code generation in this case is guaranteed to be safe,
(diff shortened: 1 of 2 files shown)


Dayuxiaoshui:
@tqchen cc

junrushao:
Please add some unittests

Dayuxiaoshui:
@junrushao I am ready, please review it again.

python/tvm_ffi/dataclasses/_utils.py
```diff
@@ -58,14 +58,13 @@ def type_info_to_cls(
...
-        if name in attrs:  # already defined
-            return
+        # Allow overriding methods (including from base classes like Object.__repr__)
+        # by always adding to attrs, which will be used when creating the new class
         func.__module__ = cls.__module__
         func.__name__ = name
         func.__qualname__ = f"{cls.__qualname__}.{name}"
         func.__doc__ = f"Method `{name}` of class `{cls.__qualname__}`"
         attrs[name] = func
-        setattr(cls, name, func)
```
junrushao, line 68: curious why we removed this line?

Dayuxiaoshui: The setattr(cls, name, func) line was removed because methods are already added to the attrs dictionary (line 67) and the new class is created via type(cls.__name__, cls_bases, attrs) (line 76), which automatically includes all attributes from attrs. Setting methods on cls (the input template class) is unnecessary since cls is not the final class being created - the final class gets all methods from attrs when created by type().

Dayuxiaoshui: We haven't explicitly tested nested c_class cases yet. The current __repr__ implementation uses {self.{field_name}!r} which calls repr() on each field value, so nested c_class objects should automatically use their own __repr__ methods. I can add a test case to verify this behavior works correctly with nested structures.

PR #418, opened 2026-01-20
Update the logic of checking `#embed`

oraluben:
This PR slightly update the example to use cubin launcher with c++'s `#embed`, make it fail earlier if compiler do not support the feature.


examples/cubin_launcher/embedded_cubin/cpp_embed/CMakeLists.txt
```diff
@@ -20,6 +20,16 @@ project(embedded_cubin_example LANGUAGES CXX CUDA)
 
 set(CMAKE_CXX_STANDARD 26)
```
junrushao, line 21: Looks like a pretty new compiler feature tbh :)

oraluben: clang implement it as part of its C++23 feature, but yes, this is a pretty new one.

Issue #413, opened 2026-01-16
UndefinedBehaviorSanitizer errors in string.h and object.h


henrylhtsang:
Hello experts! I am testing tvm-ffi, but then see some sanitizer errors. I am trying to see if they are fixable from upstream.

# Repro
```
// Compile with: clang++ -std=c++17 -fsanitize=undefined -I./include -I./3rdparty/dlpack/include repro.cpp -o repro
// run ./repro

#include <tvm/ffi/string.h>
#include <tvm/ffi/object.h>
#include <iostream>

// A simple subclass to trigger GetObjectOffsetToSubclass
class TestObj : public tvm::ffi::Object {
 public:
  static constexpr const char* _type_key = "test.TestObj";
};

int main() {
  std::cout << "Testing TVM FFI sanitizer issues...\n";

  // Issue 1: Null pointer to memcpy in String::InitData
  // This is triggered when passing a TVMFFIByteArray with null data
  std::cout << "Test 1: Creating string from null TVMFFIByteArray...\n";
  TVMFFIByteArray null_bytes;
  null_bytes.data = nullptr;
  null_bytes.size = 0;
  tvm::ffi::String str_from_null(null_bytes);
  std::cout << "String from null bytes size: " << str_from_null.size() << "\n";

  // Issue 2: Null pointer dereference in GetObjectOffsetToSubclass
  // This is typically triggered during module loading/type registration
  std::cout << "Test 2: Computing object offset...\n";
  int64_t offset = tvm::ffi::details::ObjectUnsafe::GetObjectOffsetToSubclass<TestObj>();
  std::cout << "Offset: " << offset << "\n";

  std::cout << "Done.\n";
  return 0;
}
```

# errors
```
Test 1: Creating string from null TVMFFIByteArray...
include/tvm/ffi/string.h:757:28: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:44:28: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/tvm/ffi/string.h:757:28 
String from null bytes size: 0
Test 2: Computing object offset...
include/tvm/ffi/object.h:1097:71: runtime error: member access within null pointer of type 'tvm::ffi::Object'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/tvm/ffi/object.h:1097:71 
include/tvm/ffi/object.h:1098:72: runtime error: member access within null pointer of type 'Object'
[...]

PR #419, opened 2026-01-20
Fix two issues that undefined behavior sanitizer caught

henrylhtsang:
Fixes https://github.com/apache/tvm-ffi/issues/413

### Changes                                                                                                                                         
  1. **string.h**: Guard `std::memcpy` call with size check to avoid passing null source pointer with size 0                                                                                                                                                                     
  2. **object.h**: Replace null pointer member access with standard `offsetof` macro for computing object header offsets             

I tested manually for clang and gcc. Didn't test locally for msvc since I don't have access to windows. CI should test that.

script:
```
#include <tvm/ffi/string.h>
#include <tvm/ffi/object.h>
#include <iostream>

class TestObj : public tvm::ffi::Object {
  public:
  static constexpr const char* _type_key = "test.TestObj";
};

int main() {
  std::cout << "Test 1: Creating string from null TVMFFIByteArray...\n";
  TVMFFIByteArray null_bytes{nullptr, 0};
  tvm::ffi::String str_from_null(null_bytes);
  std::cout << "String size: " << str_from_null.size() << "\n";

  std::cout << "Test 2: Computing object offset...\n";
  int64_t offset = tvm::ffi::details::ObjectUnsafe::GetObjectOffsetToSubclass<TestObj>();
  std::cout << "Offset: " << offset << "\n";

  std::cout << "Done - no UBSan errors!\n";
  return 0;
}
```

commands:
```
cd ~/tvm-ffi
git submodule update --init --recursive

clang++ -std=c++17 -fsanitize=undefined -I./include -I./3rdparty/dlpack/include -x c++  repro.cpp  -o repro

g++ -std=c++17 -fsanitize=undefined -I./include -I./3rdparty/dlpack/include repro.cpp -o repro
```

### DIFF
--- a/include/tvm/ffi/object.h
+++ b/include/tvm/ffi/object.h
@@ -1092,11 +1092,20 @@ struct ObjectUnsafe {
     return const_cast<TVMFFIObject*>(&(src->header_));
   }
 
+// Suppress -Winvalid-offsetof: we intentionally use offsetof on non-standard-layout types
+// to avoid undefined behavior from null pointer arithmetic that sanitizers flag.
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Winvalid-offsetof"
+#endif
   template <typename Class>
   TVM_FFI_INLINE static int64_t GetObjectOffsetToSubclass() {
     return static_cast<int64_t>(__builtin_offsetof(Class, header_)) -
            static_cast<int64_t>(__builtin_offsetof(Object, header_));
   }
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
 
   template <typename T>
   TVM_FFI_INLINE static T ObjectRefFromObjectPtr(const ObjectPtr<Object>& ptr) {


henrylhtsang:
cc @junrushao

include/tvm/ffi/object.h
```diff
@@ -1094,8 +1095,8 @@ struct ObjectUnsafe {
 
   template <typename Class>
   TVM_FFI_INLINE static int64_t GetObjectOffsetToSubclass() {
-    return (reinterpret_cast<int64_t>(&(static_cast<Class*>(nullptr)->header_)) -
-            reinterpret_cast<int64_t>(&(static_cast<Object*>(nullptr)->header_)));
+    return static_cast<int64_t>(offsetof(Class, header_)) -
```
tqchen: it is helpful to likely use __builtin_offsetof instead, mainly because it is more permissive per standard

henrylhtsang:
Updated per feedbacks. Re-tested manually for clang++ and g++.

junrushao:
Looks like the compiler doesn't like `__builtin_offsetof` :(

henrylhtsang:
@junrushao I can add a macro to suppress invalid-offsetof if need. Thoughts?

junrushao:
@henrylhtsang I like this idea!

tqchen:
agree that supressing warning would be better in this case

henrylhtsang:
updated. need to rerun ci

tqchen:
Thanks @henrylhtsang !

PR #420, opened 2026-01-24
[DLPack] bump to latest version

guan404ming:
This PR bumps DLPack to latest version. I've checked there is no any compat issue need to be updated in our src code.


guan404ming:
I have changed this pr to update the dlpack version instead of relaxing the check.

PR #423, opened 2026-02-02
refactor: Get `compiled_kernel` from Triton Call Directly

sgjzfzzf:
Triton call will return the `CompiledKernel` directly to get the cubin. The previous code searches the cubin by cache, which is unnecessary and not robust.

This PR refactors that by taking compiled_kernel directly from the Triton kernel call.


tqchen:
cc @yaoyaoding do u mind take a look

yaoyaoding:
I run the updated example and it works. 

LGTM, thanks @sgjzfzzf !

PR #435, opened 2026-02-07
Reliability update

### LINKED ISSUE #433: [Feature Request] Provide `cp314t` for `torch-c-dlpack-ext`
As title.

Related issue: https://github.com/tile-ai/tilelang/pull/1805

Also, since `torch-c-dlpack-ext` is only useful when installed as prebuilt binary, I propose to upload an empty sdist as a fallback. This would help when it's not necessary for torch (e.g. >= 2.10) or other cases (e.g. new python versions such as this issue)

oraluben:
Closes #433 in a way that do not bailout when compiling in environment without torch.

Also fixes that dlpack ext's symbols does not match with torch (e.g. in ngc container)

### DIFF
--- a/addons/torch_c_dlpack_ext/build_backend.py
+++ b/addons/torch_c_dlpack_ext/build_backend.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import os
 import subprocess
 import sys
+from contextlib import suppress
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


junrushao (approved):

junrushao:
Good catch!

PR #438, opened 2026-02-09
feat: add `__copy__`, `__deepcopy__`, and `__replace__` for FFI objects

junrushao:
## Summary
This PR adds first-class copy support to TVM-FFI objects, including Python `copy.copy`, `copy.deepcopy`, and dataclass-style `__replace__`.

## Changes
- Adds reflection-based deep-copy runtime support in C++:
  - new `ffi.DeepCopy` entrypoint (`src/ffi/extra/deep_copy.cc`), which
    does memoized graph copy that preserves shared references and
    cycles.
    - recursive copy of `ffi.Array` / `ffi.Map`
    - treats `Str`/`Bytes` as immutable terminal values
    - resolves fields by runtime value so `Any`/`ObjectRef` fields containing containers/objects are copied correctly
- Auto-registers shallow-copy support for copy-constructible reflected object types:
    - introduces `__ffi_shallow_copy__` registration in `ObjectDef`
    - exposes shallow-copy method via type attributes for generic lookup
- Wires Python class registration to install:
    - `__copy__`, `__deepcopy__`, and `__replace__` for supported types
    - clear `TypeError` for non-copyable types
    - deepcopy support for container roots (`ffi.Array`, `ffi.Map`) via `ffi.DeepCopy`
- Updates dataclass/c_class method attachment to use reflected callables consistently and hook copy setup.
- Adds extensive test coverage in `tests/python/test_copy.py`, including:
    - shallow/deep copy behavior
    - shared-reference and cycle preservation
    - long-string deep copy edge cases
    - `Any` and `ObjectRef` deep copy edge cases
    - non-copyable type failures (`TestNonCopyable`)

### DIFF
--- a/include/tvm/ffi/reflection/registry.h
+++ b/include/tvm/ffi/reflection/registry.h
@@ -460,6 +460,12 @@ struct init {
   }
 };
 
+/*! \brief Well-known type attribute names used by the reflection system. */
+namespace type_attr {
+inline constexpr const char* kInit = "__ffi_init__";
+inline constexpr const char* kShallowCopy = "__ffi_shallow_copy__";
+}  // namespace type_attr
+
 /*!
  * \brief Helper to register Object's reflection metadata.
  * \tparam Class The class type.
@@ -481,6 +487,7 @@ class ObjectDef : public ReflectionDefBase {
   explicit ObjectDef(ExtraArgs&&... extra_args)
       : type_index_(Class::_GetOrAllocRuntimeTypeIndex()), type_key_(Class::_type_key) {
     RegisterExtraInfo(std::forward<ExtraArgs>(extra_args)...);
+    AutoRegisterCopy();
   }
 
   /*!
@@ -591,6 +598,25 @@ class ObjectDef : public ReflectionDefBase {
   template <typename T>
   friend class OverloadObjectDef;
 
+  /*! \brief Shallow-copy \p self via the C++ copy constructor. */
+  static ObjectRef ShallowCopy(const Class* self) {
+    return ObjectRef(ffi::make_object<Class>(*self));
+  }
+
+  void AutoRegisterCopy() {
+    if constexpr (std::is_copy_constructible_v<Class>) {
+      // Register __ffi_shallow_copy__ as an instance method
+      RegisterMethod(type_attr::kShallowCopy, false, &ObjectDef::ShallowCopy);
+      // Also register as a type attribute for generic deep copy lookup
+      Function copy_fn = GetMethod(std::string(type_key_) + "." + type_attr::kShallowCopy,
+                                   &ObjectDef::ShallowCopy);
+      TVMFFIByteArray attr_name = {type_attr::kShallowCopy,
+                                   std::char_traits<char>::length(type_attr::kShallowCopy)};
+      TVMFFIAny attr_value = AnyView(copy_fn).CopyToTVMFFIAny();
+      TVM_FFI_CHECK_SAFE_CALL(TVMFFITypeRegisterAttr(type_index_, &attr_name, &attr_value));
+    }
+  }
+
   template <typename... ExtraArgs>
   void RegisterExtraInfo(ExtraArgs&&... extra_args) {
     TVMFFITypeMetadata info;
@@ -663,7 +689,7 @@ class ObjectDef : public ReflectionDefBase {
 
   int32_t type_index_;
   const char* type_key_;
-  static constexpr const char* kInitMethodName = "__ffi_init__";
+  static constexpr const char* kInitMethodName = type_attr::kInit;
 };
 
 /*!
--- a/python/tvm_ffi/registry.py
+++ b/python/tvm_ffi/registry.py
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import functools
 import json
 import sys
 from typing import Any, Callable, Literal, Sequence, TypeVar, overload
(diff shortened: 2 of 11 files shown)


python/tvm_ffi/registry.py
```diff
@@ -335,25 +335,97 @@ def _add_class_attrs(type_cls: type, type_info: TypeInfo) -> type:
...
+
+
+def _get_deep_copy_func() -> core.Function:
+    global _deep_copy_func_cache  # noqa: PLW0603
+    if _deep_copy_func_cache is None:
+        _deep_copy_func_cache = get_global_func("ffi.DeepCopy")
+    return _deep_copy_func_cache
+
+
+_deep_copy_func_cache: core.Function | None = None
```
junrushao: probably don't need this - just call `get_global_func` every time

include/tvm/ffi/extra/copy.h
```diff
@@ -0,0 +1,48 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*!
+ * \file tvm/ffi/extra/copy.h
+ * \brief Reflection-based object copy utilities
+ */
+#ifndef TVM_FFI_EXTRA_COPY_H_
+#define TVM_FFI_EXTRA_COPY_H_
```
tqchen: deep_copy.h?

include/tvm/ffi/extra/copy.h
```diff
@@ -0,0 +1,48 @@
...
+ * Recursively copies the value and all reachable objects in its object graph.
+ * Objects must have copy support enabled via `refl::enable_copy()` to be deep copied.
+ * Primitive types, strings, and bytes are returned as-is (they are immutable).
+ * Arrays and Maps are recursively deep copied.
+ * Objects without copy support are kept as shared references.
+ *
+ * \param value The value to deep copy.
+ * \return The deep copied value.
+ */
+TVM_FFI_EXTRA_CXX_API Any DeepCopyValue(const Any& value);
```
tqchen: I feel DeepCopy as an API name is fine here

include/tvm/ffi/reflection/registry.h
```diff
@@ -460,6 +460,49 @@ struct init {
...
+ *
+ * \code{.cpp}
+ * refl::ObjectDef<MyObject>()
+ *     .def(refl::enable_copy())
+ *     .def(refl::init<int64_t, std::string>());
+ * \endcode
+ *
+ * \note The object type must have a copy constructor.
+ */
+struct enable_copy {
```
tqchen: should it be enable_deep_copy to be explicit?

src/ffi/extra/copy.cc
```diff
@@ -0,0 +1,268 @@
...
+
+namespace tvm {
+namespace ffi {
+
+/*!
+ * \brief Deep copier that recursively copies objects using shallow copy + field replacement.
+ *
+ * Algorithm:
+ * Phase 1: Walk the entire object graph (objects, arrays, maps) using BFS,
+ *          collecting all reachable nodes in a vector. An unordered_set
```
tqchen: Likely this algorithm can be simplified to a single pass algorithm with memo_ that maps previously mapped Object into a new one.

The RunCopy will internally look up the memo and return copied objects if found. If not, it will then attempt to run the deep copy recursively and store the result into the memo. There is no need to record things like node order, visited will be part of the memo_ map that dedup double visits

include/tvm/ffi/reflection/registry.h
junrushao: No it's for shallow copy

tqchen: yes, i understand, seems the confusion part is that it enables a shallow copy API that was required for deep copy?

junrushao:
Still - implication of cyclic reference and CoW containers (Array, Map) makes it pretty challenging to have an absolutely correct implementation of `__deepcopy__`. Converting it to draft.

include/tvm/ffi/reflection/registry.h
junrushao: alternatively, let's enable copy by default for all object types which are copy_constructible

include/tvm/ffi/extra/copy.h
junrushao: good idea!

include/tvm/ffi/extra/copy.h
junrushao: updated

junrushao:
Updated per discussion with TQ. @tqchen please re-review

src/ffi/extra/deep_copy.cc
```diff
@@ -0,0 +1,162 @@
...
+    if (value.type_index() < TypeIndex::kTVMFFIStaticObjectBegin) {
+      return value;
+    }
+    const Object* obj = value.as<Object>();
+    if (auto it = copy_map_.find(obj); it != copy_map_.end()) {
+      return it->second;
+    }
+    int32_t ti = obj->type_index();
+    // Strings and bytes are immutable — return as-is.
+    if (ti == TypeIndex::kTVMFFIStr || ti == TypeIndex::kTVMFFIBytes) {
```
tqchen: this also applies to kTVMFFIShape which is immutable and can return as it is

include/tvm/ffi/reflection/registry.h
```diff
@@ -460,6 +460,9 @@ struct init {
   }
 };
 
+/*! \brief Method name used for shallow copy registration. */
+inline constexpr const char* kShallowCopyMethodName = "__ffi_shallow_copy__";
```
tqchen: kInitMethodName is member of ObjectDef, for consistency maybe make it also member? 

Alterntaively, consider group these constants into struct or namespace like `struct StaticTypeKey {`. Maybe TypeAttrName or `reflection::type_attr::`

Issue #430, opened 2026-02-06
[Feature] Support specifying CUDA version


DarkSharpness:
Related issue: https://github.com/sgl-project/sglang/issues/18108

Root cause: `CUDA Driver Version: 470.82.01` (too old), which leads to failure in version detectection in line 148

https://github.com/apache/tvm-ffi/blob/437323c9aecce445409820527abcf237c1c66dc1/python/tvm_ffi/cpp/extension.py#L135-L158

I wonder if users can manually specify the cuda major/minor version when building jit modules.

tqchen:
@DarkSharpness would `TVM_FFI_CUDA_ARCH_LIST` fit this need? or are you suggesting we pass them as arguments in the API. 

Both should be fine approaches, and i think we would be happy to take a pr

DarkSharpness:
I just realized  `TVM_FFI_CUDA_ARCH_LIST` can fit this need by 1. overiding the env var 2. build JIT kernel 3. reset the env var. This brings a little inconvenience for user side, but can fit in every case. I'm not sure whether we really need a special argument to override the version.

tqchen:
sounds good, maybe we can suggest `TVM_FFI_CUDA_ARCH_LIST`  first then update the api if needed, whatever the path we think convenient for the user side.

junrushao:
A related issue: https://github.com/apache/tvm-ffi/pull/440.

Could you all suggest actionable plan?

mvanhorn:
Implemented in https://github.com/apache/tvm-ffi/pull/605.

tqchen:
closing for now since there is alternative support in sglang

PR #440, opened 2026-02-11
Add Support for NVIDIA Ampere GPUs in _get_cuda_target

guoyuhong:
I'm using SGLang, which relies on TVM-FFI, on a machine equipped with an NVIDIA A10 GPU. We encountered the following error:
image

This issue is commonly observed on NVIDIA Ampere-generation GPUs (e.g., A10, A100) — see related discussion: https://github.com/sgl-project/sglang/issues/18108 , https://github.com/sgl-project/sglang/pull/18496 , https://github.com/apache/tvm-ffi/issues/430.

The root cause is that older NVIDIA drivers (commonly deployed on Ampere systems) do not support the compute_cap query field in nvidia-smi. As a result, _get_cuda_target fails when trying to auto-detect the CUDA compute capability.

image
To address this, we fall back to querying the GPU name via:

```bash
nvidia-smi --query-gpu=name --format=csv,noheader
```

image

and then map known Ampere GPU names (e.g., "NVIDIA A10") to their corresponding compute capabilities (e.g., 8.6).

This change enables robust GPU detection on Ampere devices with legacy drivers while maintaining backward compatibility.

### DIFF
--- a/python/tvm_ffi/cpp/extension.py
+++ b/python/tvm_ffi/cpp/extension.py
@@ -171,11 +171,10 @@ def _get_cuda_target() -> str:
                     if gpu_key in gpu_name:
                         return f"-gencode=arch=compute_{major}{minor},code=sm_{major}{minor}"
             except (subprocess.CalledProcessError, FileNotFoundError):
-                # If nvidia-smi fails or the GPU is not in our map,
-                # we proceed to the default fallback.
                 pass
-            # fallback to a reasonable default
-            return "-gencode=arch=compute_70,code=sm_70"
+            raise RuntimeError(
+                "Could not detect CUDA compute_cap automatically. Please set TVM_FFI_CUDA_ARCH_LIST environment variable."
+            )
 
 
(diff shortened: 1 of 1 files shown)


python/tvm_ffi/cpp/extension.py
```diff
@@ -154,6 +154,26 @@ def _get_cuda_target() -> str:
             major, minor = compute_cap.split(".")
             return f"-gencode=arch=compute_{major}{minor},code=sm_{major}{minor}"
         except Exception:
+            try:
```
junrushao, line 157: I don't particularly like the hard coding here. Can we instead emit a warning, saying "please set env variable `TVM_FFI_CUDA_ARCH_LIST`"

guoyuhong: @junrushao I've changed the default return statement to raise an exception instead. Is this the correct approach?

junrushao (approved):
Thanks for the patch!

PR #442, opened 2026-02-12
fix: handle UTF-8 bytes correctly in JSON parser control character check

MasterJH5574:
Cast `*cur_` to `unsigned char` before comparing against space (0x20). On platforms where `char` is signed, bytes >= 0x80 (valid UTF-8 continuation/lead bytes) were misinterpreted as negative values, causing all non-ASCII strings to be rejected as "Invalid control character".

### DIFF
--- a/src/ffi/extra/json_parser.cc
+++ b/src/ffi/extra/json_parser.cc
@@ -131,7 +131,9 @@ class JSONParserContext {
         ++cur_;
         return true;
       }
-      if (*cur_ < ' ' || *cur_ == '\\') {
+      // Use uint8_t: on platforms where char is signed, raw UTF-8 bytes (>= 0x80)
+      // would be negative and incorrectly treated as control characters.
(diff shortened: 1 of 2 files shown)


tqchen:
good to followup with a comment

junrushao:
Good catch! BTW, could you take a look at general UTF-8 handling in strings in mlc-python: https://github.com/mlc-ai/mlc-python/blob/main/include/mlc/core/str.h#L293-L443, and figure out what is missing in tvm-ffi?

junrushao:
also this one might be interesting (not necessarily relevant): https://github.com/sgl-project/mini-sglang/blob/main/python/minisgl/tokenizer/detokenize.py#L14-L36

MasterJH5574:
@junrushao Thanks for the comments. To be standard JSON compliance, it looks like UTF-8 handling in tvm-ffi is already complete and no extra changes are needed.

PR #443, opened 2026-02-12
feat: Introduce List as Mutable Sequence

junrushao:
## Summary

- Introduces `List<T>`, a mutable sequence container alongside the existing immutable `Array<T>`, with full C++ and Python support
- Extracts `SeqBaseObj` as a shared base class for `ArrayObj` and `ListObj`, consolidating type-erased sequence operations (iteration, element access, reverse, etc.)
- Refactors `ArrayObj` to inherit from `SeqBaseObj`, significantly reducing code duplication while preserving copy-on-write semantics
- Adds cycle detection in serialization, structural hash, structural equal, and JSON writer for `List` (which can form reference cycles unlike `Array`)
- Updates `stl.h` so `TypeTraits<std::vector<T>>` accepts both `kTVMFFIArray` and `kTVMFFIList`

## Test plan

- [x] C++ unit tests for `List` operations (`test_list.cc`: construction, push_back, pop_back, insert, erase, resize, Set, clear, iterator, reverse, COW-free mutation)
- [x] C++ tests for `List` structural equal/hash (`test_structural_equal_hash.cc`: List equality, List-vs-Array type mismatch)
- [x] C++ tests for `List` serialization (`test_serialization.cc`: empty/single-element list, round-trip, cycle detection)
- [x] Python tests for `List` (`test_container.py`: construction, indexing, slicing, append, insert, pop, extend, `__delitem__`, `__setitem__`, `__contains__`, iteration, reverse, `len`, pickle round-trip)

### DIFF
--- a/include/tvm/ffi/container/list.h
+++ b/include/tvm/ffi/container/list.h
@@ -70,6 +70,10 @@ class ListObj : public SeqBaseObj {
   /*!
    * \brief Ensure the list has at least n slots.
    * \param n The lower bound of required capacity.
+   *
+   * \note Leak-safety: Any's move constructor is noexcept, so the
+   *       move loop below cannot throw and leave the buffer in a
+   *       partially-constructed state.
    */
   void Reserve(int64_t n) {
     if (n <= TVMFFISeqCell::capacity) {
--- a/include/tvm/ffi/container/seq_base.h
+++ b/include/tvm/ffi/container/seq_base.h
@@ -266,6 +266,14 @@ class SeqBaseObj : public Object, protected TVMFFISeqCell {
   /// \endcond
 };
 
+/*!
+ * \brief CRTP base for sequence type-traits (Array, List).
+ *
+ * \tparam Derived Must expose:
+ *   - `static constexpr int32_t kPrimaryTypeIndex` — the canonical FFI type index
+ *   - `static constexpr int32_t kOtherTypeIndex`   — an alternative accepted type index
+ *   - `static constexpr const char* kTypeName`      — human-readable name for diagnostics
+ */
 template <typename Derived, typename SeqRef, typename T>
 struct SeqTypeTraitsBase : public ObjectRefTypeTraitsBase<SeqRef> {
   using Base = ObjectRefTypeTraitsBase<SeqRef>;
--- a/src/ffi/extra/structural_equal.cc
+++ b/src/ffi/extra/structural_equal.cc
@@ -32,7 +32,6 @@
 
 #include <cmath>
 #include <unordered_map>
-#include <unordered_set>
 #include <utility>
 
 namespace tvm {
@@ -320,22 +319,6 @@ class StructEqualHandler {
   template <typename SeqType>
   // NOLINTNEXTLINE(performance-unnecessary-value-param)
   bool CompareSequence(SeqType lhs, SeqType rhs) {
-    const Object* lhs_ptr = lhs.get();
-    const Object* rhs_ptr = rhs.get();
-    auto pair = std::make_pair(lhs_ptr, rhs_ptr);
-    if (active_sequence_pairs_.count(pair)) {
-      TVM_FFI_THROW(ValueError) << "Cycle detected during StructuralEqual: a "
-                                << lhs_ptr->GetTypeKey() << " contains itself";
-    }
-    active_sequence_pairs_.insert(pair);
-    bool result = CompareSequenceImpl(std::move(lhs), std::move(rhs));
-    active_sequence_pairs_.erase(pair);
-    return result;
-  }
-
-  template <typename SeqType>
-  // NOLINTNEXTLINE(performance-unnecessary-value-param)
-  bool CompareSequenceImpl(SeqType lhs, SeqType rhs) {
     if (lhs.size() != rhs.size()) {
       // fast path, size mismatch, and there is no path tracing
       // return false since we don't need informative error message
@@ -437,17 +420,6 @@ class StructEqualHandler {
   std::unordered_map<ObjectRef, ObjectRef, ObjectPtrHash, ObjectPtrEqual> equal_map_lhs_;
   // map from rhs to lhs
   std::unordered_map<ObjectRef, ObjectRef, ObjectPtrHash, ObjectPtrEqual> equal_map_rhs_;
-
-  // track active sequence (lhs, rhs) pairs for cycle detection
-  struct PointerPairHash {
-    size_t operator()(const std::pair<const Object*, const Object*>& p) const {
-      auto h1 = std::hash<const void*>()(p.first);
-      auto h2 = std::hash<const void*>()(p.second);
-      return h1 ^ (h2 * 0x9e3779b97f4a7c15ULL + 0x9e3779b9 + (h1 << 6) + (h1 >> 2));
-    }
-  };
-  std::unordered_set<std::pair<const Object*, const Object*>, PointerPairHash>
-      active_sequence_pairs_;
 };
 
 bool StructuralEqual::Equal(const Any& lhs, const Any& rhs, bool map_free_vars,
--- a/src/ffi/extra/structural_hash.cc
+++ b/src/ffi/extra/structural_hash.cc
@@ -34,7 +34,6 @@
 #include <cmath>
 #include <limits>
 #include <unordered_map>
-#include <unordered_set>
 #include <utility>
 
 namespace tvm {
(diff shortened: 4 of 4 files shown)


include/tvm/ffi/container/list.h
```diff
@@ -0,0 +1,784 @@
...
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace tvm {
+namespace ffi {
+
+/*! \brief List node content in list */
+class ListObj : public Object {
```
tqchen: consider share the same Object layout as ArrayObj but only differ in type index, likely can re-use code and also brings possible future compact if we want to turn List into Array via freeze

junrushao: @tqchen are you recommending me to move the layout out to `c_api.h`?

src/ffi/extra/serialization.cc
```diff
@@ -58,92 +62,121 @@ class ObjectGraphSerializer {
...
-      }
-      default: {
-        if (value.type_index() >= TypeIndex::kTVMFFIStaticObjectBegin) {
-          // serialize type key since type index is runtime dependent
-          node.Set("type", value.GetTypeKey());
-          node.Set("data", CreateObjectData(value));
-        } else {
-          TVM_FFI_THROW(RuntimeError) << "Cannot serialize type `" << value.GetTypeKey() << "`";
-          TVM_FFI_UNREACHABLE();
+    try {
```
tqchen: use RAII scope instead of try catch here

src/ffi/extra/serialization.cc
```diff
@@ -245,8 +280,20 @@ class ObjectGraphDeserializer {
     if (decoded_nodes_[node_index] != nullptr) {
       return decoded_nodes_[node_index];
     }
+    if (active_decode_nodes_.find(node_index) != active_decode_nodes_.end()) {
+      TVM_FFI_THROW(RuntimeError) << "Cyclic references are only supported for ffi.List in "
+                                  << "FromJSONGraph";
+    }
+    active_decode_nodes_.insert(node_index);
```
tqchen: use RAII guard

include/tvm/ffi/c_api.h
```diff
@@ -356,7 +359,35 @@ typedef struct {
...
+ */
+#ifdef __cplusplus
+struct TVMFFISeqCell {
+#else
+typedef struct {
+#endif
+  /*! \brief Data pointer to the first element of the sequence. */
+  void* data;
+  /*! \brief Number of elements used. */
+  int64_t length;
```
junrushao: rename to `size`

include/tvm/ffi/c_api.h
```diff
@@ -356,7 +359,35 @@ typedef struct {
...
+  /*! \brief Number of elements allocated. */
+  int64_t capacity;
+  /*!
+   * \brief Optional data deleter when data is allocated separately
+   *        and its deletion is not managed by object deleter.
+   */
+  void (*data_deleter)(void*);
```
junrushao, line 390: clarify what if it's nullptr

include/tvm/ffi/container/array.h
```diff
@@ -42,20 +42,22 @@ namespace tvm {
 namespace ffi {
 
 /*! \brief Array node content in array */
-class ArrayObj : public Object, public details::InplaceArrayBase<ArrayObj, TVMFFIAny> {
+class ArrayObj : public Object,
```
junrushao: unify `ArrayObj` and `ListObj` on non-resizing operations

include/tvm/ffi/container/list.h
```diff
@@ -0,0 +1,810 @@
...
+struct TypeTraits<List<T>> : public ObjectRefTypeTraitsBase<List<T>> {
+  static constexpr int32_t field_static_type_index = TypeIndex::kTVMFFIList;
+  using ObjectRefTypeTraitsBase<List<T>>::CopyFromAnyViewAfterCheck;
+
+  TVM_FFI_INLINE static std::string GetMismatchTypeInfo(const TVMFFIAny* src) {
+    if (src->type_index != TypeIndex::kTVMFFIList && src->type_index != TypeIndex::kTVMFFIArray) {
+      return TypeTraitsBase::GetMismatchTypeInfo(src);
+    }
+    if constexpr (!std::is_same_v<T, Any>) {
+      auto for_each = [&](auto* n) -> std::optional<std::string> {
```
tqchen: move as member template function

junrushao: readability upgrade

src/ffi/extra/structural_equal.cc
```diff
@@ -31,6 +32,8 @@
 
 #include <cmath>
 #include <unordered_map>
+#include <unordered_set>
```
junrushao: TODO: we dont allow cycles

include/tvm/ffi/container/seq_base.h
```diff
@@ -0,0 +1,357 @@
...
+
+  void MoveElementsRight(int64_t dst, int64_t src_begin, int64_t src_end) {
+    Any* begin = MutableBegin();
+    std::move_backward(begin + src_begin, begin + src_end, begin + dst + (src_end - src_begin));
+  }
+  /// \endcond
+};
+
+template <typename Derived, typename SeqRef, typename T>
+struct SeqTypeTraitsBase : public ObjectRefTypeTraitsBase<SeqRef> {
```
tqchen, line 278: document the expected derived properties

include/tvm/ffi/container/list.h
```diff
@@ -0,0 +1,523 @@
...
+   * \brief Ensure the list has at least n slots.
+   * \param n The lower bound of required capacity.
+   */
+  void Reserve(int64_t n) {
+    if (n <= TVMFFISeqCell::capacity) {
+      return;
+    }
+    Any* old_data = MutableBegin();
+    Any* new_data = static_cast<Any*>(::operator new(sizeof(Any) * static_cast<size_t>(n)));
+    for (int64_t i = 0; i < TVMFFISeqCell::size; ++i) {
```
tqchen, line 84: This is not leaksafe when constructor of AnyMove causes exception, might worth cross check and ensure any construct of move from any is no except

src/ffi/extra/structural_equal.cc
```diff
@@ -302,6 +309,33 @@ class StructEqualHandler {
...
+    return CompareSequence(std::move(lhs), std::move(rhs));
+  }
+
+  template <typename SeqType>
+  // NOLINTNEXTLINE(performance-unnecessary-value-param)
+  bool CompareSequence(SeqType lhs, SeqType rhs) {
+    const Object* lhs_ptr = lhs.get();
+    const Object* rhs_ptr = rhs.get();
+    auto pair = std::make_pair(lhs_ptr, rhs_ptr);
+    if (active_sequence_pairs_.count(pair)) {
```
tqchen: likely we don't need cycle detection in structequal to keep things simpleZ?

src/ffi/extra/structural_hash.cc
```diff
@@ -185,11 +190,25 @@ class StructuralHashHandler {
...
-    for (const auto& elem : arr) {
+  uint64_t HashArray(Array<Any> arr) { return HashSequence(std::move(arr)); }
+
+  // NOLINTNEXTLINE(performance-unnecessary-value-param)
+  uint64_t HashList(List<Any> list) { return HashSequence(std::move(list)); }
+
+  template <typename SeqType>
+  // NOLINTNEXTLINE(performance-unnecessary-value-param)
+  uint64_t HashSequence(SeqType seq) {
+    const Object* obj_ptr = seq.get();
```
tqchen: consider assuming DAG in structural hash to simplify things

include/tvm/ffi/container/list.h
junrushao: Wait - just want to double check - is `Any`'s move constructor leak-safe? I think they are, right?

src/ffi/extra/structural_equal.cc
junrushao: Still a minimal cycle detection should be okay just in case there's recursion 🤔

include/tvm/ffi/container/seq_base.h
junrushao: done

include/tvm/ffi/container/list.h
tqchen: i think so, we should document via comment

src/ffi/extra/structural_equal.cc
tqchen: main  reason is for sequal/hash there can be more kind of cycles, so only doing it for list is a bit limited, and usually the usecase is for non-cycle, so keeping things simple here should be good. we can use cycle detection in json dumps

junrushao: removed

src/ffi/extra/structural_hash.cc
junrushao: removed

include/tvm/ffi/container/list.h
junrushao: updated

tqchen (approved):

PR #446, opened 2026-02-14
feat: add `DefaultFactory` support to field reflection

junrushao:
`DefaultValue` stores a single static default shared across all instances created via reflection.  For mutable defaults (Array, Map, etc.) this causes aliasing: every object receives the same underlying container.  `DefaultFactory` fixes this by storing a callable `() -> Any` that is invoked each time a default is needed, producing a fresh value per instance—mirroring Python dataclass `default_factory`.

Concrete changes:

- Rename `TVMFFIFieldInfo::default_value` → `default_value_or_factory` to reflect that the slot now holds either a value or a factory.
- Add `kTVMFFIFieldFlagBitMaskDefaultFromFactory` (1 << 5) to `TVMFFIFieldFlagBitMask`.
- Add `reflection::DefaultFactory` trait (registry.h), symmetric to `DefaultValue`.
- Add `reflection::SetFieldToDefault` helper (accessor.h) that resolves the default—calling the factory when the flag is set—so the three consumption sites (creator.h, reflection_extra.cc, serialization.cc) share one implementation.
- Propagate the rename through Rust (`c_api.rs`) and Cython (`base.pxi`) bindings.
- Add `TestObjWithFactory` + three tests exercising flag inspection, per-instance freshness, and explicit-value bypass.

### DIFF
--- a/include/tvm/ffi/c_api.h
+++ b/include/tvm/ffi/c_api.h
@@ -861,6 +861,15 @@ typedef enum {
    * This is an optional meta-data for structural eq/hash.
    */
   kTVMFFIFieldFlagBitMaskSEqHashDef = 1 << 4,
+  /*!
+   * \brief The default_value_or_factory is a callable factory function () -> Any.
+   *
+   * When this flag is set along with kTVMFFIFieldFlagBitMaskHasDefault,
+   * the default_value_or_factory field contains a Function that should be
+   * called with no arguments to produce the default value, rather than
+   * being used directly as the default value.
+   */
+  kTVMFFIFieldFlagBitMaskDefaultFromFactory = 1 << 5,
 #ifdef __cplusplus
 };
 #else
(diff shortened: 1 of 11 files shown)


include/tvm/ffi/c_api.h
```diff
@@ -861,6 +861,15 @@ typedef enum {
...
   kTVMFFIFieldFlagBitMaskSEqHashDef = 1 << 4,
+  /*!
+   * \brief The default_value_or_factory is a callable factory function () -> Any.
+   *
+   * When this flag is set along with kTVMFFIFieldFlagBitMaskHasDefault,
+   * the default_value_or_factory field contains a Function that should be
+   * called with no arguments to produce the default value, rather than
+   * being used directly as the default value.
+   */
+  kTVMFFIFieldFlagBitMaskDefaultIsFactory = 1 << 5,
```
tqchen: kTVMFFIFieldFlagBitMaskDefaultFromFactory? Is is a bit ambiguious because default value is not factory

PR #447, opened 2026-02-14
feat: add `ffi.GetInvalidObject` global function for MISSING singleton

junrushao:
C++ changes:
- Register `ffi.GetInvalidObject` global function in container.cc
- Remove `ffi.MapGetMissingObject` in favor of the new name

Python changes:
- Initialize `MISSING` in core.pyx via `_get_global_func` after `Function` is registered
- Re-export `MISSING` in container.py from core
- Add `MISSING: Object` to core.pyi type stub
- Update _ffi_api.py stubs and __all__

Tests:
- Add `test_missing_object` verifying singleton identity across imports and Map.get() integration


include/tvm/ffi/c_api.h
```diff
@@ -551,6 +551,18 @@ TVM_FFI_DLL int TVMFFIObjectIncRef(TVMFFIObjectHandle obj);
...
+ * \brief Get the global invalid/missing object as an owned TVMFFIAny.
+ *
+ * Returns a singleton Object that represents an "invalid" or "missing" state.
+ * The returned TVMFFIAny holds an owned reference to the object (i.e., the
+ * reference count is incremented). The caller is responsible for releasing
+ * it via the normal Any cleanup path.
+ *
+ * \param out The output owned TVMFFIAny.
+ */
+TVM_FFI_DLL void TVMFFIAnyGetInvalidObject(TVMFFIAny* out);
```
tqchen: does it have to be a C API, or can we have it as a packed function API? I think it is better to have it as a packed function API for now

PR #448, opened 2026-02-14
feat: Support direct construction of `tvm_ffi.Function` from callables

junrushao:
**blocked by cuteDSL usage of this private API**

Add `__init__` to the Cython `Function` class so users can write:

    fadd = tvm_ffi.Function(add)

instead of going through `tvm_ffi.convert_to_tvm_func`. The constructor:
- Validates that the argument is callable
- Copies the handle (with incref) if the input is already a Function
- Otherwise converts via `_convert_to_ffi_func_handle`

Also adds the corresponding type stub in `core.pyi` and a test in `test_function.py`.

Re-lands the original change reverted in #406.


tqchen:
we need to hold this after a few cutedsl releases

python/tvm_ffi/cython/function.pxi
```diff
@@ -889,6 +889,27 @@ cdef class Function(Object):
...
+
+        This constructor allows creating a `tvm_ffi.Function` directly
+        from a Python function or another `tvm_ffi.Function` instance.
+
+        Parameters
+        ----------
+        func : Callable[..., Any]
+            The Python callable to wrap.
+        """
+        cdef TVMFFIObjectHandle chandle = NULL
```
tqchen, line 903: we need to hold the change for a few cutedsl releases before cutedsl get updated

PR #449, opened 2026-02-15
fix: prevent silent NULL-handle creation in Object.__init__

junrushao:
**blocked by cuteDSL usage of this private API**

Previously, calling `Object()` or constructing an unregistered subclass without `__init__` would silently succeed and produce an object with a NULL handle. Add `Object.__init__` that raises TypeError to catch this misuse early. Remove the now-redundant `super().__init__()` call in AccessPath.__init__.

### DIFF
--- a/python/tvm_ffi/access_path.py
+++ b/python/tvm_ffi/access_path.py
@@ -108,7 +108,6 @@ class AccessPath(Object):
 
     def __init__(self) -> None:
         """Disallow direct construction; use `AccessPath.root()` instead."""
-        super().__init__()
         raise ValueError(
             "AccessPath can't be initialized directly. "
             "Use AccessPath.root() to create a path to the root object"
--- a/python/tvm_ffi/cython/object.pxi
+++ b/python/tvm_ffi/cython/object.pxi
@@ -106,6 +106,12 @@ cdef class CObject:
         # case of error before chandle is set
         self.chandle = NULL
 
+    def __init__(self):
+        raise TypeError(
+            f"Cannot directly create {type(self).__name__} instance. "
+            f"Use FFI constructors or register the class with @register_object."
+        )
+
     def __dealloc__(self):
         if self.chandle != NULL:
             CHECK_CALL(TVMFFIObjectDecRef(self.chandle))
--- a/tests/python/test_object.py
+++ b/tests/python/test_object.py
@@ -136,6 +136,33 @@ def test_opaque_type_error() -> None:
     )
 
 
+def test_object_direct_init_disabled() -> None:
+    """Object() and unregistered subclasses must not silently create NULL-handle objects."""
+    # Base Object class
+    with pytest.raises(TypeError, match="Cannot directly create Object instance"):
+        tvm_ffi.Object()
+
+    # Unregistered subclass without __init__
+    class UnregisteredObj(tvm_ffi.Object):
+        pass
+
+    with pytest.raises(TypeError, match="Cannot directly create UnregisteredObj instance"):
+        UnregisteredObj()
+
+    # Registered class without __c_ffi_init__ should still raise
+    with pytest.raises(RuntimeError, match="__init__ method of this class is not implemented"):
+        tvm_ffi.testing.TestObjectBase()
+
+    # Registered class with __c_ffi_init__ should work fine
+    pair = tvm_ffi.testing.TestIntPair(3, 4)  # ty: ignore[too-many-positional-arguments]
+    assert pair.a == 3 and pair.b == 4
+
+    # FFI-returned objects should work fine
+    obj = tvm_ffi.testing.create_object("testing.TestObjectBase", v_i64=7)
+    assert obj.__chandle__() != 0
+    assert obj.v_i64 == 7  # ty: ignore[unresolved-attribute]
+
+
 def test_object_protocol() -> None:
     class CompactObject:
         def __init__(self, backend_obj: Any) -> None:


python/tvm_ffi/access_path.py
```diff
@@ -105,7 +105,6 @@ def _path_equal(self, _1: AccessPath, /) -> bool: ...
 
     def __init__(self) -> None:
         """Disallow direct construction; use `AccessPath.root()` instead."""
-        super().__init__()
         raise ValueError(
```
junrushao, line 111: **[high | API Consistency | Consensus]** `AccessPath.__init__` raises `ValueError` while the base `Object.__init__` now raises `TypeError`. Three different exception types are used for "cannot construct":
- `Object.__init__` → `TypeError`
- `__init__invalid` (registry) → `RuntimeError`
- `AccessPath.__init__` → `ValueError`

`TypeError` is the most Pythonic choice for "this class cannot be instantiated this way" (matches abstract classes and `__new__` guards). Consider unifying:

```python
def __init__(self) -> None:
    """Disallow direct construction; use `AccessPath.root()` instead."""
    raise TypeError(
        "AccessPath cannot be initialized directly. "
        "Use AccessPath.root() to create a path to the root object"
    )
```

_Source: Consensus (Claude + Codex)_

python/tvm_ffi/cython/object.pxi
```diff
@@ -116,6 +116,12 @@ cdef class Object:
         # case of error before chandle is set
         self.chandle = NULL
 
+    def __init__(self):
```
junrushao, line 109: **[low | Maintainability | Codex]** The error message hardcodes "FFI constructors" and "@register_object" as the solutions. If new mechanisms for object creation are added, this message won't mention them. Minor suggestion — consider making it slightly more generic, e.g. adding "factory methods" as an option.

_Source: Codex_

tests/python/test_object.py
```diff
@@ -136,6 +136,33 @@ def test_opaque_type_error() -> None:
...
+
+    # Unregistered subclass without __init__
+    class UnregisteredObj(tvm_ffi.Object):
+        pass
+
+    with pytest.raises(TypeError, match="Cannot directly create UnregisteredObj instance"):
+        UnregisteredObj()
+
+    # Registered class without __c_ffi_init__ should still raise
+    with pytest.raises(RuntimeError, match="__init__ method of this class is not implemented"):
```
junrushao, line 153: **[medium | Correctness | Claude]** The match string `"__init__ method of this class is not implemented"` is a fragile substring — the actual message in `registry.py:387` is `"The __init__ method of this class is not implemented."` (starts with "The"). While `re.search` makes this work, a reader may think the full message starts with `"__init__"`.

**[medium | Test Clarity | Codex]** It's also unclear from the test whether the `RuntimeError` comes from the missing `__c_ffi_init__` path or from the new `__init__` guard. A brief comment clarifying the error path would help.

Suggested fix:
```python
# Registered class without __c_ffi_init__: the registry replaces __init__
# with __init__invalid, which raises RuntimeError (not the base TypeError)
with pytest.raises(RuntimeError, match="The __init__ method of this class is not implemented"):
    tvm_ffi.testing.TestObjectBase()
```

_Source: Claude (match string) + Codex (clarity)_

tests/python/test_object.py
```diff
@@ -136,6 +136,33 @@ def test_opaque_type_error() -> None:
     )
 
 
+def test_object_direct_init_disabled() -> None:
```
junrushao, line 139: **[medium | Test Coverage | Claude]** Missing test for `AccessPath.__init__` after removing `super().__init__()`. The removal changes exception ordering — previously the base `Object.__init__` (no-op) ran first, then `ValueError` fired. Now only `ValueError` fires directly. If someone accidentally removes the `raise ValueError`, the base `TypeError` would kick in with different semantics.

Suggested addition:
```python
def test_access_path_direct_init_raises():
    """AccessPath() must raise ValueError directing users to root()."""
    from tvm_ffi.access_path import AccessPath
    with pytest.raises(ValueError, match="AccessPath can.t be initialized directly"):
        AccessPath()
```

_Source: Claude_

tests/python/test_object.py
```diff
@@ -136,6 +136,33 @@ def test_opaque_type_error() -> None:
...
+    with pytest.raises(TypeError, match="Cannot directly create UnregisteredObj instance"):
+        UnregisteredObj()
+
+    # Registered class without __c_ffi_init__ should still raise
+    with pytest.raises(RuntimeError, match="__init__ method of this class is not implemented"):
+        tvm_ffi.testing.TestObjectBase()
+
+    # Registered class with __c_ffi_init__ should work fine
+    pair = tvm_ffi.testing.TestIntPair(3, 4)  # ty: ignore[too-many-positional-arguments]
+    assert pair.a == 3 and pair.b == 4
```
junrushao, line 158: **[nit | Readability | Claude]** If `pair.a == 3` passes but `pair.b == 4` fails, pytest will only report the combined expression failed without showing which half. Consider splitting:

```python
assert pair.a == 3
assert pair.b == 4
```

_Source: Claude_

junrushao (commented):
## Synthesized Code Review Report

**Scope**: PR #449 (all commits since diverging from main) | **Files changed**: 3 | **Reviewers**: Claude Code, Codex

**Summary**: 2 high, 3 medium, 5 low/nit findings. Core mechanism is sound and well-motivated — `__cinit__` still initializes `chandle = NULL`, `__init__` now raises early, and all legitimate construction paths go through `__new__` + handle setup.

### General Findings (not mapped to specific diff lines)

**[high | API Design | Consensus]** **Breaking change**: Any code calling `Object()` or `super().__init__()` on unregistered subclasses will now fail with `TypeError`. This is intentional but should be documented in release notes.

**[low | Maintainability | Claude]** `registry.py:387` — `__init__invalid` message says `"The __init__ method of this class is not implemented."` without naming the class. The new `Object.__init__` demonstrates how to include `type(self).__name__`; consider aligning for consistency in a follow-up.

**[low | Test Coverage | Codex]** Missing test for a registered subclass that naively overrides `__init__` — would help document the expected behavior.

**[nit | Test | Claude]** `# ty: ignore[too-many-positional-arguments]` appears in multiple test locations; may indicate a stub issue for `TestIntPair.__init__`.

### Questions for Author

1. **Pickle path**: `__reduce__` returns `(_new_object, (cls,), self.__getstate__())` which uses `cls.__new__(cls)`, bypassing `__init__`. Have you verified pickle round-trips still work?
2. **Cython MRO**: If a Python subclass doesn't define `__init__`, the Cython base `Object.__init__` fires. Registry replaces `__init__` via `setattr` on the Python type's `__dict__`. Confirmed for all registered types?

---
_Generated by multi-model code review (Claude Code + Codex)_

python/tvm_ffi/cython/object.pxi
tqchen, line 109: there might be a cases that depends on class construction and handle move, so we need to cross check downstream for this change

junrushao:
blocked by cuteDSL usage of this private API

junrushao:
Closing - will open a new PR once unblocked.

PR #454, opened 2026-02-16
feat: add DFS-based ffi.ReprPrint for unified object repr

junrushao:
## Summary

- Single C++ `ffi.ReprPrint` function produces human-readable repr for any TVM FFI value
- DFS with 3-state tracking (NotVisited/InProgress/Done):
  - **DAGs**: memoized repr returned in full on every re-encounter
  - **Cycles**: detected via InProgress state, shown as `...`
- Addresses hidden by default; set `TVM_FFI_REPR_WITH_ADDR=1` to show
- Per-field `Repr(false)` InfoTrait to exclude fields from repr output
- Built-in repr for String, Bytes, Tensor, Shape, Array, List, Map
- All Python `__repr__` methods delegate to this function

## Format Examples

```
42                                    # int
"hello"                               # String
(1, 2, 3)                             # Array
[1, 2, 3]                             # List
{"key": "value"}                      # Map
testing.MyObj(x=1, y="hi")            # User object
...                                   # Cycle marker
float32[3, 4]@cpu:0@0x1234            # Tensor
```

## Test plan

- [x] 55 Python tests covering primitives, containers, user objects, DAGs, cycles, and `TVM_FFI_REPR_WITH_ADDR`
- [x] All pre-commit hooks pass (ruff, ty, clang-format, markdownlint, etc.)
- [x] Container tests (`test_container.py`) pass with updated Array format


python/tvm_ffi/cython/object.pxi
```diff
@@ -26,8 +26,21 @@ def _set_class_object(cls):
...
+    """Object repr function using ffi.ReprPrint when available."""
+    global _REPR_PRINT, _REPR_PRINT_LOADED
+    if not _REPR_PRINT_LOADED:
+        _REPR_PRINT_LOADED = True
+        _REPR_PRINT = _get_global_func("ffi.ReprPrint", False)
+    if _REPR_PRINT is not None:
+        try:
+            return str(_REPR_PRINT(obj))
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
junrushao: Done. Added a comment explaining that `__repr__` must never raise, so the silent fallback is intentional.

src/ffi/extra/repr_print.cc
```diff
@@ -0,0 +1,531 @@
...
+        reflection::ForEachFieldInfoWithEarlyStop(type_info, [&](const TVMFFIFieldInfo* finfo) {
+          std::string_view field_name(finfo->name.data, finfo->name.size);
+          if (field_name == name) {
+            reflection::FieldGetter getter(finfo);
+            CollectAny(getter(obj));
+            return true;
+          }
+          return false;
+        });
+      }
```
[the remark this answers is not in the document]
junrushao: Done. Refactored to use a per-field `kTVMFFIFieldFlagBitMaskReprOff` bit flag instead of `repr_fields` name lookup. Both `CollectFieldChildren` and `GenericRepr` now iterate fields once with `ForEachFieldInfo`, checking the flag per field — O(M) instead of O(N*M).

src/ffi/extra/repr_print.cc
```diff
@@ -0,0 +1,531 @@
...
+        unsigned char c = static_cast<unsigned char>(b.data()[i]);
+        if (c >= 32 && c < 127 && c != '\"' && c != '\\') {
+          os << static_cast<char>(c);
+        } else {
+          os << "\\x" << std::hex << std::setw(2) << std::setfill('0') << static_cast<int>(c);
+        }
+      }
+      os << "\"";
+      return os.str();
+    }
```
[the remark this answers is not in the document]
junrushao: Done. Extracted a `FormatBytes(const char*, size_t)` helper used by both `ReprOfAny` (SmallBytes) and `ReprBytes`.

src/ffi/extra/repr_print.cc
```diff
@@ -0,0 +1,531 @@
...
+  bool first = true;
+  for (const Any& elem : *obj) {
+    if (!first) os << ", ";
+    first = false;
+    String s = fn_repr(elem).cast<String>();
+    os << std::string_view(s.data(), s.size());
+  }
+  os << "]";
+  return String(os.str());
+}
```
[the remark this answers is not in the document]
junrushao: Done. Extracted a `ReprSequence<TContainer>` template. `ReprArray` and `ReprList` now delegate to it.

tests/python/test_repr.py
```diff
@@ -0,0 +1,163 @@
...
+
+def test_repr_duplicate_reference() -> None:
+    """Test that duplicate object references use short form."""
+    inner = tvm_ffi.testing.create_object("testing.TestIntPair", a=1, b=2)
+    arr = tvm_ffi.Array([inner, inner])
+    result = _ffi_api.ReprPrint(arr)
+    # The array should show two elements; one full, one short form
+    # The full form includes (a=1, b=2), the short form is just type@addr
+    assert "a=1" in result
+    assert "b=2" in result
```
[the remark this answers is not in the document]
junrushao: Not applicable — the BFS dedup prevents infinite recursion during collection, but both occurrences intentionally render the full form. The test correctly reflects this behavior.

src/ffi/extra/repr_print.cc
```diff
@@ -0,0 +1,478 @@
...
+  const TVMFFITypeInfo* type_info = TVMFFIGetTypeInfo(obj->type_index());
+  return std::string(type_info->type_key.data, type_info->type_key.size);
+}
+
+/*!
+ * \brief Lazily initialize and return the __ffi_repr__ TypeAttrColumn.
+ *
+ * Returns nullptr if the column does not exist (e.g., before registration).
+ */
+const TVMFFITypeAttrColumn* GetReprColumn() {
```
tqchen: use refl::TypeAttrColumn class we don't need inline C ABI impl as in any because we are now at extra

tqchen:
Looks like great improvement, would be good to discuss a bit how are we thinking in terms of repr printing and behavior against python repr. The repr do not exactly need round trippable(serialization perhaps is a better choice there). So there is a question whether we want to have the duplicated value printing (or as an option), and whether it should be default.

The default behavior of python atm is simply expand. Expansion also could make sense for cases like immutable data structure. Say a shape value get reference in multiple places beause of the way we copy the data structure

```python
x = (1,2,3)
y = (1,4)

print([y, y])
> [(1,4), (1,4)]

# circle case
x = [12]
x.append(x)
> [12, [...]]
``` 

This being said, there can be value in cases where we might want duplicated value printing. Perhaps we can do it under a flag.

src/ffi/extra/repr_print.cc
```diff
@@ -0,0 +1,478 @@
...
+  std::ostringstream os;
+  os << "\"" << std::string_view(obj->data, obj->size) << "\"";
+  return String(os.str());
+}
+
+String ReprBytes(const details::BytesObj* obj, const Function& fn_repr) {
+  return String(FormatBytes(obj->data, obj->size));
+}
+
+String ReprTensor(const TensorObj* obj, const Function& fn_repr) {
```
tqchen, line 286: although this one is concise, personally i think it is good to be explicit here

```python
Tensor(shape=(1, 2), dtype="float32", device="cuda:0")
```

junrushao: I think it's a matter of personal taste, but also please consider the output length and readability to users. In that case:

```
float32[10, 20]@gpu:0
```

seems an overall win.

junrushao:
Updated the text format and duplication handling

src/ffi/extra/repr_print.cc
tqchen: see https://github.com/apache/tvm-ffi/blob/main/src/ffi/extra/structural_equal.cc#L169

src/ffi/extra/repr_print.cc
tqchen: i agree it is a close call. My original thinking is ideally align with existing ones and not inventing new syntax (that users needs to learn from). Indeed the output synax would be longer, but being explicit allows user not having to learn what `float32[10, 20]` means. 

Just to note some of the nit comments when i read the new syntax:

- when looking at it, i would have questions like "is it maps to an on stack raw array?" (in which case it is not, and it maps to a tensor). 
    - `Tensor(shape=(1, 2), dtype="float32", device="cuda:0")` avoids that confusion at the cost of slightly longer  
- Another minor nit is that `@gpu:0` syntax conflicts a bit with `@{addr}` although it is really nit.

junrushao:
> The default behavior of python atm is simply expand. Expansion also could make sense for cases like immutable data structure. Say a shape value get reference in multiple places beause of the way we copy the data structure
> 
> ```python
> x = (1,2,3)
> y = (1,4)
> 
> print([y, y])
> > [(1,4), (1,4)]
> 
> # circle case
> x = [12]
> x.append(x)
> > [12, [...]]
> ```
> 
> This being said, there can be value in cases where we might want duplicated value printing. Perhaps we can do it under a flag.

Behavior from the latest commit:

```
import tvm_ffi as ffi
from tvm_ffi._ffi_api import ReprPrint

x = ffi.List([12])
x.append(x)
print(ReprPrint(x))
# gives: [12, ...]

y = ffi.Array((1,4))
y = ffi.Array((y, y))
print(ReprPrint(y))
# gives: ((1, 4), (1, 4))
```

@tqchen LMK if it looks good to you

src/ffi/extra/repr_print.cc
junrushao: updated

src/ffi/extra/repr_print.cc
junrushao: I updated the syntax. `@0x{addr}` shows up only when `TVM_FFI_REPR_WITH_ADDR` is set to `1`. It means there's no point of confusion

PR #455, opened 2026-02-16
[CORE] Add TypeConverter system for runtime type validation/conversion

junrushao:
## Summary

- Add a `TypeConverter` system to the Cython FFI module that takes a `TypeSchema` and returns a `tvm_ffi.Function` wrapping a C++ conversion function `(Any) -> Any`
- The converter validates/converts input values to match the target type, with fast-path (zero-copy) for containers when elements already match and slow-path (element-by-element conversion) otherwise
- Detailed `TypeError` messages are provided on type mismatch (e.g., `"expected int, got float"`, `"expected list[int], but element at index 2 has type float"`)

## New Files

- `python/tvm_ffi/cython/tvm_ffi_type_converter.h` — C++ header implementing `TVMFFIPyTypeConverterNode` tree with `CheckStrict` (zero-copy eligibility) and `Convert` methods for all supported kinds
- `python/tvm_ffi/cython/type_converter.pxi` — Cython bridge mapping `TypeSchema` origins to C++ converter nodes, exposes `create_type_converter(schema)`
- `tests/python/test_type_converter.py` — 34 comprehensive tests covering POD types, strings/bytes, special types, Optional, Union, container fast/slow paths, tuples, nested containers, object hierarchy, and error messages

## Supported Type Kinds

| Kind | CheckStrict | Convert (also accepts) |
|------|------------|----------------------|
| int | kTVMFFIInt | + bool (coerce) |
| float | kTVMFFIFloat | + int, bool (coerce) |
| bool | kTVMFFIBool | strict only |
| str | kTVMFFIStr/SmallStr | same |
| bytes | kTVMFFIBytes/SmallBytes | same |
| Tensor | kTVMFFITensor | + DLTensorPtr |
| Object | type hierarchy check | same |
| Optional | None or child | same |
| Union | any child strict | try each child |
| list | Array + all elems strict | slow-path: new Array |
| dict | Map + all k/v strict | slow-path: new Map |
| tuple | Array + size + per-elem | slow-path: new Array |

## Test plan

- [x] `uv run pytest tests/python/test_type_converter.py -v` — 34/34 pass
- [x] `uv run pytest tests/python/ -v` — 425 passed, 0 failures (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


junrushao:
Addressed Gemini's review comments:

1. **Arity validation in constructor** (high): Added argument count validation in the `TypeChecker` constructor for generic kinds:
   - `Optional`: requires exactly 1 type argument
   - `Array`/`List`: requires 0 or 1 type arguments
   - `Map`: requires 0 or 2 type arguments
   
   Malformed schemas now fail early with a clear error message via `TVM_FFI_ICHECK`.

2. **Simplified DLTensorPtr logic** (medium): Removed the redundant inner `if (ti == TypeIndex::kTVMFFITensor)` branch since both paths returned `Any(src)`. The simplified code just checks `src.try_cast<DLTensor*>()` and returns `Any(src)` on success.

PR #456, opened 2026-02-16
doc: Properly document `field_static_type_index` and add more tests

junrushao:
## Summary

Restore full usage of `field_static_type_index` in TVMFFIFieldInfo with
thorough documentation explaining its compile-time static type semantics:
the field reflects the declared type at compile time, which may differ
from the runtime type (e.g., `Any` vs `int`, `Array<Any>` vs `Array<int>`).

The serializer uses this field to inline POD values (None, Bool, Int,
Float, DataType) directly, avoiding unnecessary node-graph overhead,
while routing other types through the standard node graph.

Changes:
- Add comprehensive documentation to `field_static_type_index` in
  TVMFFIFieldInfo (C API) explaining all possible values and semantics
- Restore `TypeToFieldStaticTypeIndex` and `TypeToRuntimeTypeIndex`
  templates in type_traits.h
- Restore `field_static_type_index` in all TypeTraits specializations
- Restore POD field inlining in serialization/deserialization
- Add comprehensive C++ and Python serialization tests (40 + 58 tests)

## Test plan

- [x] All 302 C++ tests pass (including 40 serialization + 21 reflection tests)
- [x] All 58 Python serialization tests pass
- [x] Verified `field_static_type_index` is populated correctly in reflection via test object field info

### DIFF
--- a/include/tvm/ffi/c_api.h
+++ b/include/tvm/ffi/c_api.h
@@ -979,23 +979,28 @@ typedef struct {
    */
   TVMFFIAny default_value_or_factory;
   /*!
-   * \brief Records the static type kind of the field.
+   * \brief The compile-time static type index of the field.
    *
-   * Possible values:
+   * This reflects the type declared at compile time, which is only trustworthy
+   * for statically-inferrable types. It does NOT necessarily match the runtime
+   * type. For example, a field declared as `Any` will have
+   * `field_static_type_index == kTVMFFIAny` even if it holds an `int` at runtime,
+   * and `Array<Any>` will report `kTVMFFIArray` even though the elements may be
+   * `Array<int>` at runtime.
    *
-   * - TVMFFITypeIndex::kTVMFFIObject for general objects.
-   *   The value is nullable when kTVMFFIObject is chosen.
-   * - Static object type kinds such as Map, Dict, String
-   * - POD type index, note it does not give information about storage size of the field.
-   * - TVMFFITypeIndex::kTVMFFIAny if we don't have specialized info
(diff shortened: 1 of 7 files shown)


junrushao:
Thanks for the review @gemini-code-assist! No actionable changes were requested — the review confirmed the approach is sound. The PR is ready for human review.

src/ffi/extra/serialization.cc
```diff
@@ -364,36 +333,12 @@ class ObjectGraphDeserializer {
...
-      }
-    };
-
     json::Object data_object = data.cast<json::Object>();
     reflection::ForEachFieldInfo(type_info, [&](const TVMFFIFieldInfo* field_info) {
       String field_name(field_info->name);
       void* field_addr = reinterpret_cast<char*>(ptr.get()) + field_info->offset;
       if (data_object.count(field_name) != 0) {
-        Any field_value = decode_field_value(field_info, data_object[field_name]);
+        Any field_value = GetOrDecodeNode(data_object[field_name].cast<int64_t>());
```
tqchen: would be good to deliberate this choice a bit further. Since one motivation was to simplify the serialization of a known schema, so serialization of fields like int/bool/string, do not need to go through second level indirection.

tqchen:
> Incorrect for Any/Optional fields: compile-time types don't reliably match runtime values

This claim was incorrect as for Any field, the static type index will be marked as any and the system will default to normal serialization path like object

junrushao:
@tqchen I updated the PR - now the only change is a strongly worded docstring and a few new unittests

PR #462, opened 2026-02-18
[REACTOR] Cleanup and improve map array containers

tqchen:
This PR runs the following refactor of the map and array containers.

- Remove InplaceArrayBase since it is not as useful
- Extract MapObj into map_base.h so it can be used for other different subclass and variant of map if needed.
- Add extra helper methods such as clean

We also reorganized SmallBaseMapObj after DenseMapObj as SmallBaseMapObj is a specialization, which can help possible future refactors to make SmallBaseMapObj be aware of DenseMapObj.

The split out of map_base.h is roughly manual copy and a few extra manual edits to reduce unexpected surprises.


tqchen:
To simplify the review, the PR is structured in two commits

- The first commit splits out the related files into map_base.h and cleans up the sequence
   - Copy of MapObj implementation is just copy with no editing
- The second commit contains the related edits of map_base.h to allow subclass

PR #467, opened 2026-02-20
[COMPACT][TEMP] Array compact temp fix 

tqchen:
In the 0.1.19 patch, we changed the operator[] to return T(from const T). This is the right change. However, flashinfer depends on an sutble implicit conversion rule that requires const T behavior (which is also not desirable). There is a patch in flashinfer to fx this issue.

This PR is a temp workaround so 0.1.19 release won't block flashinfer. We can reland this in next release as flashinfer's change gets into a release.

### DIFF
--- a/include/tvm/ffi/container/array.h
+++ b/include/tvm/ffi/container/array.h
@@ -376,7 +376,7 @@ class Array : public ObjectRef {
    * \param i The index
    * \return the i-th element.
    */
-  T operator[](int64_t i) const {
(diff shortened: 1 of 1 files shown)


tqchen:
cc @junrushao @yongwww

tqchen:
cross ref https://github.com/flashinfer-ai/flashinfer/pull/2602, we can reland after this PR goes into flashinfer release

PR #475, opened 2026-02-25
[FIX] Prevent integer overflow in GetDataSize on 32-bit platforms

gnguralnick:
On 32-bit platforms (e.g. WASM), the intermediate multiplication `numel * dtype.bits * dtype.lanes` in `GetDataSize()` can overflow `size_t` for large tensors, even when the final byte count (after dividing by 8) fits within 32 bits.

This PR casts the intermediate computation to `uint64_t` to avoid the overflow, then casts the result back to `size_t`.

I was experiencing this error when trying to run a model with webllm using a runtime compiled by mlc-llm - the operation was overflowing, causing the allocation of 0-byte buffers.


junrushao:
CC @tqchen

PR #479, opened 2026-02-27
[FEAT] Add CMake namespace aliases for source dependency builds

ctuu:
## Summary
Add namespace-qualified ALIAS targets (tvm_ffi::header, tvm_ffi::shared, tvm_ffi::static) that point to the in-tree build targets (tvm_ffi_header, tvm_ffi_shared, tvm_ffi_static).

## Problem
When tvm-ffi is used as a source dependency, `tvm_ffi_configure_target` expects `tvm_ffi::header` / `tvm_ffi::shared` targets to exist. However, these are only created by the package config file (`cmake/tvm_ffi-config.cmake`), which is not included during source builds. Downstream projects using source dependencies cannot use these targets.

## Solution
Create ALIAS targets in both:
- Top-level `CMakeLists.txt` for `tvm_ffi::header`
- `cmake/Utils/Library.cmake` for `tvm_ffi::static` and `tvm_ffi::shared` (created by `tvm_ffi_add_target_from_obj`)

This ensures the namespace-qualified names are available in both source and package builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/cmake/Utils/Library.cmake)
 # they can be used in cases where user do not want to link into the library in cases like deferred
 # linking
 add_library(tvm_ffi_header INTERFACE)
+add_library(tvm_ffi::header ALIAS tvm_ffi_header)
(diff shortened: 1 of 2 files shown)


tqchen:
cc @junrushao

PR #480, opened 2026-02-27
feat(python)!: enforce __slots__=() for Object subclasses via _ObjectSlotsMeta

junrushao:
## Summary

- Split Cython `Object` into `CObject` (extension type) + `Object` (Python class with `_ObjectSlotsMeta` metaclass) to enforce `__slots__=()` across the entire Object hierarchy.
- Remove the `slots: bool = True` keyword argument from `_ObjectSlotsMeta`; subclasses that need a `__dict__` declare `__slots__ = ("__dict__",)` explicitly.
- Simplify `Module.__getattr__` to store looked-up functions via `setattr` (into `__dict__` via `__slots__ = ("__dict__",)`), letting normal attribute lookup serve as cache.
- Replace module-global `_REPR_PRINT` cache with local `_ffi_api` import in `object_repr()`.

## Breaking Changes

- All `Object` subclasses now enforce `__slots__=()` by default. Code that sets arbitrary instance attributes will raise `AttributeError`. Migrate by declaring needed slots explicitly (e.g., `__slots__ = ("__dict__",)`).
- `_ObjectSlotsMeta` no longer accepts `slots` keyword argument.
- `DLDeviceType.kDLTrn` renumbered from 17→18; `kDLMAIA=17` added.

## Test plan

- [x] `uv run pytest -vvs tests/python` — 532 passed, 23 skipped, 1 xfailed
- [x] All pre-commit hooks pass (ruff, ty, cython-lint, etc.)
- [ ] C++ tests (no C++ changes in this PR)
- [ ] Rust tests (no Rust changes in this PR)


tests/python/test_object.py
```diff
@@ -136,6 +136,29 @@ def test_opaque_type_error() -> None:
     )
 
 
+def test_object_direct_init_disabled() -> None:
```
junrushao: @tqchen please take a look at this breaking change - this is orthogonal to `__slots__=()` and I'm happy to take it back if needed.

tqchen: would be good to keep it for a bit and we reland once the downstream cutedsl releases catchs up

Issue #487, opened 2026-02-28
Heavy Python Initialization During Build-Time Path Discovery


ctuu:
## 1. Problem Statement

The `tvm-ffi` package currently uses a Python-based configuration tool (`tvm_ffi.config`) to provide metadata (headers, library paths) to external build systems. However, importing this tool triggers a full package initialization.

When **PyTorch+CUDA** is present in the environment, the overhead is catastrophic. The Python interpreter must load massive binary frameworks and initialize the GPU driver/context just to retrieve a simple string path.

---

### Scenario A: CMake `find_package()`

When using `find_package(tvm_ffi CONFIG REQUIRED)`, CMake executes the Python interpreter multiple times to resolve variables like `includedir` and `libfiles`.

https://github.com/apache/tvm-ffi/blob/b1abaeac7103606a458d2bb91438652030d5ae88/cmake/tvm_ffi-config.cmake#L25-L41

* **The Cost:** On a system with PyTorch+CUDA, this adds **7–10 seconds** to every configuration run.
* **The Risk:** It creates a hard dependency on Python environment during the C++ build phase, violating principles of hermetic builds.

### Scenario B: Python CLI & Other Build Systems

Even when querying paths directly via the CLI (used by Meson, Bazel, or manual scripts), the latency is unacceptably high.

```bash
# In an environment with PyTorch + CUDA
$ time python -m tvm_ffi.config --includedir
/path/to/include

real    0m2.3s 

```

* **The Root Cause:** `tvm_ffi.config` is tightly coupled to `__init__.py`. Importing it triggers a "Heavy Initialization Sequence":
1. **Framework Loading:** Loads GBs of PyTorch/CUDA shared objects.
2. **Hardware Handshake:** GPU driver context initialization.
3. **Symbol Resolution:** Thousands of symbols resolved across heavy binaries.

---

## 3. Recommended Solution (Scenario A)

I recommend moving to a **Native CMake Target Export** model. This is the industry-standard "clean way" to handle C++ library discovery.

* **Implementation:** Use `install(EXPORT)` to generate a static `.cmake` file during the installation of `tvm-ffi`.
* **Why it works:** It shifts the work from **runtime discovery** (calling Python) to **install-time generation**.
* **Result:** `find_package()` becomes a near-instant ($<0.01\text{s}$) file read. It works offline and without Python installed.

---

## 4. Discussion (Scenario B)

[...]

PR #489, opened 2026-02-28
[FEAT] Skip imports in config mode

tqchen:
This PR detects and skips import in config mode,
reducing the overall time and dependency needed invoking tvm_ffi.config


tqchen:
address #487

junrushao:
image

looks like my agent is not happy with this change

PR #492, opened 2026-03-01
feat(python): add Function.__init__ constructor for direct callable wrapping

junrushao:
## Summary

- Add `Function.__init__` constructor to the Cython `Function` class, enabling direct construction via `tvm_ffi.Function(callable)` instead of requiring `tvm_ffi.convert(callable)`
- Support three input modes: wrap a plain Python callable, copy-construct from an existing `Function` (with ref-count increment), or construct with no arguments for backward compatibility (null handle)
- Reject non-callable inputs with `TypeError` and moved-from `Function` objects with `ValueError`

## Architecture

- Adds `__init__` to the Cython `Function` extension type in `function.pxi`. For plain callables, delegates to the existing `_convert_to_ffi_func_handle` helper. For `Function`-to-`Function` copies, increments the C-level ref count via `TVMFFIObjectIncRef` and shares the handle.
- `func` parameter defaults to `None` for backward compatibility: when omitted the object remains in its default null-handle state (matching the prior behavior where `__cinit__` left `chandle = NULL`).
- No changes to the C ABI, C++ layer, or global registry.

## Public Interfaces

- **New**: `tvm_ffi.Function(func: Callable[..., Any] | None = None) -> Function` constructor.
- Type stub updated in `core.pyi`.

## Behavioral Changes

- Previously, `tvm_ffi.Function` could only be obtained via `tvm_ffi.convert()` or the global function registry. Now it can be constructed directly.
- Copy-constructing from a `Function` produces a second ref to the **same** underlying packed function (verified via `same_as`).
- Calling `Function()` with no arguments preserves backward compatibility by returning a null-handle instance identical to the prior default construction.

## Tests

- `test_pyfunc_init` covers: wrapping a Python callable, copy-constructing from `Function`, `ValueError` on moved-from source, no-arg null-handle backward compat, `TypeError` on non-callable input.
- Moved-from test uses `__new__` to avoid leaking a handle that would be immediately overwritten by `__move_handle_from__`.

## Untested Edge Cases

- Thread-safety of concurrent `__init__` calls (covered by GIL in CPython).
- Interaction with `release_gil` property on the newly constructed function (inherits default from `__cinit__`).


python/tvm_ffi/cython/function.pxi
```diff
@@ -889,6 +889,32 @@ cdef class Function(CObject):
     def __cinit__(self) -> None:
         self.c_release_gil = _RELEASE_GIL_BY_DEFAULT
 
+    def __init__(self, func: Optional[Callable[..., Any]] = None) -> None:
```
tqchen, line 892: let us wait until cutedsl release cycles fixes the compact issue, the land this one

tqchen (changes requested):

PR #496, opened 2026-03-03
[FIX] Fix 3 deny-level clippy lints in Rust bindings

Aphoh:
## Summary

Fix 3 deny-level clippy lints in the Rust bindings that block `cargo clippy` from passing cleanly.

### 1. `data_as_slice_mut(&self)` → `data_as_slice_mut(&mut self)`

**Lint:** `clippy::mut_from_ref`

The original signature returned `&mut [T]` from `&self`. This allows creating aliased mutable references through multiple shared borrows — which is unsound. Changing to `&mut self` lets the borrow checker enforce exclusivity.

Per the book's ["Creating a Safe Abstraction over Unsafe Code"](https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html) section: a function that wraps unsafe code does not need to be marked `unsafe fn` if it validates its own preconditions. Here the signature (`&mut self`) enforces the key invariant, and the body validates dtype, device, and contiguity before calling `from_raw_parts_mut`.

### 2. `with_stream` → `unsafe fn with_stream`

**Lint:** `clippy::not_unsafe_ptr_arg_deref`

Takes a `TVMFFIStreamHandle` (`*mut c_void`) and passes it to the C function `TVMFFIEnvSetStream`. The book says: "just because a function contains unsafe code doesn't mean we need to mark the entire function as unsafe." But that only applies when the function can validate its own preconditions. Here it cannot — there is no way to verify the stream handle is valid (CUDA stream handles are opaque with no validation API). The caller must uphold that invariant, so the function must be `unsafe fn`.

### 3. `from_extern_c` → `unsafe fn from_extern_c`

**Lint:** `clippy::not_unsafe_ptr_arg_deref`

Takes a `*mut c_void` handle, a C function pointer (`safe_call`), and an optional `deleter`, all forwarded to `TVMFFIFunctionCreate`. The function cannot verify that the handle is valid, that `safe_call` is compatible with it, or that `deleter` will correctly free it. These are invariants the caller must uphold — same reasoning as `with_stream`.

### DIFF
--- a/rust/tvm-ffi/src/collections/tensor.rs
+++ b/rust/tvm-ffi/src/collections/tensor.rs
@@ -148,21 +148,22 @@ impl Tensor {
             ))
         }
     }
-    /// Get the data as a mutable slice
-    ///
-    /// Note that we do allow mutable data access to copies of the Tensor,
-    /// as in the case of low-level deep learning frameworks.
-    ///
-    /// # Arguments
-    /// * `T` - The type of the data
-    ///
-    /// # Returns
-    /// * `Result<&mut [T]>` - The data as a mutable slice
     /// Returns the tensor data as a mutable slice.
     ///
-    /// Note: if the `Tensor` has been cloned (via `ObjectArc`), the caller
-    /// must ensure no other clone is concurrently reading the data.
-    pub fn data_as_slice_mut<T: AsDLDataType>(&mut self) -> Result<&mut [T]> {
+    /// This method takes `&self` rather than `&mut self` by design: like
+    /// `std::fs::File::write`, the *metadata* of a Tensor (shape, dtype,
+    /// device) is governed by Rust's ownership rules, but writing to the
+    /// underlying data buffer (CPU memory or a GPU pointer) is a side-effect
+    /// outside Rust's aliasing model.  Most C/CUDA kernel APIs accept a
+    /// non-mut Tensor and mutate its data content, so requiring `&mut self`
+    /// here would force artificial mutability annotations throughout the
+    /// deep-learning stack with no real safety benefit.
+    ///
+    /// # Safety contract (caller responsibility)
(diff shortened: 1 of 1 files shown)


tqchen:
thanks @Aphoh ! the later two makes sense and likely it means we need a Stream wrapper for users so they don't have to always deal with unsafe with stream. 

Would be good to look into fixing the tests. 

The mut Tensor part also makes sense, but likely that could make structuring kernels slightly harder since we need unique mut Tensor for intermediates as most are read writes, we could try to first get it in.  As most common pattern in low-level kernels are destination passing. But most underlying kernels exposed by tvm-ffi still do not follow the mut rule (which indeed it more unsafe due to concurrent write)

Aphoh:
> Would be good to look into fixing the tests.

😂 sorry, should be fixed now!

> The mut Tensor part also makes sense, but likely that could make structuring kernels slightly harder since we need unique mut Tensor for intermediates as most are read writes, we could try to first get it in. As most common pattern in low-level kernels are destination passing. But most underlying kernels exposed by tvm-ffi still do not follow the mut rule (which indeed it more unsafe due to concurrent write)

Hmm agreed that's tricky indeed. I'm fine with either putting a clippy ignore for now or bringing it in as is. Let me know which you'd prefer.

tqchen:
On the Tensor part, i think it is mainly a matter of consistency and our level of abstraction. Just that most of the underlying kernels likely still allows passing non-mut Tensor in and allow users to mut the data content

I think this is somewhat like how File works  in rust, e.g. the metadata of the File is protected by the rust safety rules, but writing to the file do not need the file to be mut, so the implicitly the side effect of writing to content is not part of the deal. So the case now is that the data content on GPU is not yet part of the safety part. This being said, I haven't seen a lot of rust based deep learning code for to reason about whether or not such relaxation is a good or bad thing. 

For consistency i feel we could first do clippy ignore and document the rationale.

PR #503, opened 2026-03-10
feat(ffi)!: add __ffi_convert__ type attribute and Python-defined type support

junrushao:
## Summary

- Introduce `__ffi_convert__` type attribute: a per-type `Function` that performs typed `AnyView -> TObjectRef` conversion via `TypeTraits::TryCastFromAnyView`, registered through `RegisterConvertTypeAttr<T>()` and the new `ObjectDef<T>::ref<TObjectRef>()` builder method
- Add Python-defined type infrastructure in `dataclass.cc`: `PyClassDeleter`, `MakeFFINew` (calloc-based factory), field getter/setter dispatch, and `__ffi_shallow_copy__` for deep-copy support
- Register `__ffi_convert__` for all built-in static types (Object, String, Bytes, Error, Function, Shape, Tensor, Array, Map, List, Dict)
- Expose `ffi.FunctionFromExternC` global function for Python-side construction of `Function` objects from raw C function pointers

## Test plan

- [x] Existing C++ tests pass (reflection, object creation, serialization)
- [x] Existing Python tests pass (c_class / py_class decorator integration)
- [x] CI lint (clang-format, clang-tidy) passes on changed files
- [x] Integration testing with Python-side py_class decorator

**BREAKING CHANGE:** Object base type registration now explicitly sets `structural_eq_hash_kind` to `kTVMFFISEqHashKindUnsupported`, which may affect types that previously inherited an uninitialized default.

### DIFF
--- a/include/tvm/ffi/reflection/registry.h
+++ b/include/tvm/ffi/reflection/registry.h
@@ -693,6 +693,8 @@ inline constexpr const char* kHash = "__ffi_hash__";
 inline constexpr const char* kEq = "__ffi_eq__";
 /*! \brief Type attribute for custom recursive three-way comparison. */
 inline constexpr const char* kCompare = "__ffi_compare__";
+/*! \brief Type attribute for converting AnyView to a specific reflected object type. */
+inline constexpr const char* kConvert = "__ffi_convert__";
 }  // namespace type_attr
 
 /*!
--- a/include/tvm/ffi/reflection/init.h
+++ b/include/tvm/ffi/reflection/init.h
@@ -43,6 +43,27 @@ namespace tvm {
 namespace ffi {
 namespace reflection {
 
+namespace details {
+
+/*!
+ * \brief Convert an AnyView to a specific reflected object type.
+ *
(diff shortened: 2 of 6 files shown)


include/tvm/ffi/reflection/registry.h
```diff
@@ -815,6 +833,21 @@ class ObjectDef : public ReflectionDefBase {
...
   }
 
+  /*!
+   * \brief Register ``__ffi_convert__`` for the object reference wrapper.
+   *
+   * This enables typed ``AnyView -> TObjectRef`` conversion through
+   * ``TypeTraits<TObjectRef>::TryCastFromAnyView``.
+   */
+  template <typename TObjectRef>
+  TVM_FFI_INLINE ObjectDef& ref() {
```
junrushao: @tqchen Not sure if the best naming. Perhaps `.traits<TObjectRef>` can be better?

junrushao: I removed this method as it's not really needed for now

PR #530, opened 2026-04-09
chore(skill): Add skills to better develop with TVM-FFI

junrushao:
I found Claude Code isn't very good at using TVM-FFI's APIs, and hope this skill could help...

This PR is not complete yet as I haven't validated and haven't thought of a good name and trigger, but worth a good starting point. Anyone has better ideas please feel free to start from here. Ideally if we can have TVM-FFI speciifc skills available in npx:

```
npx skills add <owner/repo>
```

that would be awesome


junrushao:
CC: @spectrometerHBH @Kathryn-cat

PR #552, opened 2026-04-15
perf: improve field access overhead

junrushao:
That's the HEAD commit — our current work. Here's what it does:

---

### `bfa81be` — `perf: improve field access overhead`

**Goal**: Make `@py_class` and `@c_class` field reads match `@dataclass(slots=True)` speed (~5ns per access via CPython's `LOAD_ATTR_SLOT` specialization).

**Mechanism — three layers working together:**

1. **`_ObjectSlotsMeta.__new__`** (metaclass, `object.pxi`): Intercepts class creation to inject `__slots__` from annotations *before* `type.__new__` runs. This gives every field a `PyMemberDescrObject` descriptor (the only kind that triggers `LOAD_ATTR_SLOT`). Defaults are saved to `_tvm_ffi_saved_defaults` and cleaned up after field registration. No class recreation needed.

2. **Type-specialized C++ field getters** (`tvm_ffi_python_helpers.h` + `field_access.pxi`): Nine inline C functions (`TVMFFIPyFieldGetInt64`, `Float64`, `Bool`, `ObjectRef`, `Any`, etc.) that read field values directly from object memory via pointer arithmetic — bypassing the generic getter path (no try/catch, no `Any` construction, no `make_ret` dispatch). Cython `cdef class` wrappers (`FieldGetterInt64`, etc.) call these, selected by `_make_specialized_getter` based on field type and size.

3. **Shadow-slot populate + sync helpers** (`field_access.pxi`): When a C++ object is returned to Python (`make_ret_object`), a per-type populate function copies all field values from C++ memory into the shadow `__slots__`, using the specialized getters. Custom `__setattr__` keeps both copies in sync on writes. `__getattr__` is a fallback for unpopulated slots. `_force_set_field` is the escape hatch for frozen fields.

**Key design decisions:**
- `threading.local()` for the re-entrancy guard in `_populate` (free-threading safe)
- `register_object(init=False)` does *not* install shadow helpers — `c_class` handles its own path after dunder installation
- `__init__` wrapping skipped when `init_fn is object.__init__` to avoid breaking `init=False` types
- Separate `__init__` wrapper logic for classes with/without `__post_init__` to avoid double-populate

**Files:**
- **New**: `field_access.pxi` (consolidates all field-access infrastructure — getter/setter classes, specialized getter factory, shadow-slot registry, populate/sync helpers)
- **New**: `benchmark_field_access.py` (microbenchmark confirming `LOAD_ATTR_SLOT` parity)
- **Modified**: `base.pxi` (extern declarations for C++ getters), `tvm_ffi_python_helpers.h` (9 inline C getter functions), `object.pxi` (metaclass, populate hook in `make_ret_object`), `type_info.pxi` (extracted getters to `field_access.pxi`), `py_class.py` / `registry.py` / `c_class.py` (callers updated), `field.py` (docstring), test files (updated for `_force_set_field` API and shadow-slot descriptors)


python/tvm_ffi/cython/field_access.pxi
```diff
@@ -0,0 +1,497 @@
...
+    cdef Py_ssize_t off
+    for off in range(0, 2048, sizeof(void*)):
+        if (<void**>(base + off))[0] == target:
+            _TP_GETATTRO_OFFSET = off
+            break
+    TVMFFIPyFieldTableSetup(
+        <void*>_FT_KEY_STR,
+        _make_ret_object_cb,
+        _make_ret_cb,
+    )
```
junrushao, line 389: HACK: compute the offset of `PyObject_GenericGetAttr` wrt `PyObject*` so that we access it

PR #566, opened 2026-04-23
doc: clarify structural_eq semantics and py_class eq/hash interaction

yaoyaoding:
Rework the const-tree section in docs/concepts/structural_eq_hash.rst: the "Why not use it everywhere?" example was bad — the drawn trees had different pointers for their `+` nodes, so the const-tree pointer shortcut would never have fired. Replaced it with a shared-subtree scenario under `map_free_vars=True` that actually exercises the bug, and expanded "When is this safe?" with a third condition (interning payoff) plus a rule of thumb for choosing between `tree` and `const-tree`.

Expand the `"var"` section to document that non-ignored fields on a var type (e.g. a type annotation) are compared on first encounter but subsequent occurrences short-circuit through the bijective mapping — i.e. the mapping is sticky.

Clarify the py_class docstring so readers understand that `structural_eq` is independent of `eq` / `unsafe_hash`: the default `__eq__` / `__hash__` remain pointer-based (inherited from `Object`), while `structural_equal` / `structural_hash` are driven separately by the kind metadata registered in C++.

### DIFF
--- a/docs/concepts/structural_eq_hash.rst
+++ b/docs/concepts/structural_eq_hash.rst
@@ -396,7 +396,7 @@ The problem
 
    fun x → x + 1       should equal       fun y → y + 1
 
-Variables are not defined by their content like name. They
+Variables are not defined by their content, such as their name. They
 are defined by **where they are introduced** and **how they are used**.
 ``x`` and ``y`` above are interchangeable because they occupy the same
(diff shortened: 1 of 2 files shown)


junrushao (approved):
This is huge improvement - Thanks Yaoyao!

PR #570, opened 2026-04-25
[REFACTOR] libinfo: add extra_lib_paths parameter for foreign-caller support

tqchen:
## Summary

`tvm_ffi.libinfo._find_library_by_basename(package, target_name)`,
`load_lib_ctypes(...)`, and `load_lib_module(...)` now accept an optional
`extra_lib_paths: list[Path] | None` parameter. Caller-supplied dirs take
precedence over the existing `_rel_top_directory()` / `_dev_top_directory()`
fallback. Self-call behavior unchanged.

On miss, the `RuntimeError` now lists every directory searched so debugging
is one error message away.

## API

```python
def load_lib_ctypes(
    package: str,
    target_name: str,
    mode: str,
    extra_lib_paths: list[Path] | None = None,
) -> ctypes.CDLL: ...

def load_lib_module(
    package: str,
    target_name: str,
    keep_module_alive: bool = True,
    extra_lib_paths: list[Path] | None = None,
) -> Module: ...
```

`list[Path]` over `Sequence[Path | str]` — callers naturally produce `Path`
objects from `__file__`; single canonical type, no internal normalization,
future widening is non-breaking.

### DIFF
--- a/python/tvm_ffi/libinfo.py
+++ b/python/tvm_ffi/libinfo.py
@@ -145,7 +145,12 @@ def include_paths() -> list[str]:
     )
 
 
-def load_lib_ctypes(package: str, target_name: str, mode: str) -> ctypes.CDLL:
+def load_lib_ctypes(
+    package: str,
+    target_name: str,
+    mode: str,
+    extra_lib_paths: list[Path] | None = None,
+) -> ctypes.CDLL:
     """Load the tvm_ffi shared library by searching likely paths.
 
     Parameters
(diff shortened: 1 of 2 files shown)


junrushao (approved):

junrushao:
Curious - in which case would the existing path finding approach fail that requires us to explicitly supply this extra paths?

tqchen:
currently the existing approach searches on tvm_ffi dev package root and may not cover the related extra search path of downstream

junrushao:
Are you saying there's a case where libtvm_ffi is distributed by a downstream package instead of inside `apache-tvm-ffi`?

tqchen:
when downstream needs to load libdownstream.so in a dev mode or some extra libraries using the utility

junrushao:
I'm confused. Are you saying there's a case that downstream libraries can't find `libdownstream.so` using:

```
im.distribution(package)
```
?

tqchen:
when downstream develop packages in devmode and not yet packaged like in the fallback path we had in tvm ffi

junrushao:
By “dev mode” you are referring to the case we didn’t install via “pip install -e”, and instead use PYTHONPATH?

tqchen:
that is correct

junrushao:
I see. It’s a rare case nowadays and this patch makes sense

tqchen:
there is also a main compatibility in case downstream expects lib path provided by some environment variable DOWNSTREAM_LIB_PATH

junrushao:
Anything can happen in dev mode if we are not doing “pip install -e .”

tqchen:
no in the case of env var, i was refering to a case where the library is not expected to be shipped by the package itself but instead shipped through system(e.g. things like `JAVA_HOME`), where the path is specified through env variable

junrushao:
indeed i can totally see a dev mode case where we don't ship software by package, don't install them in standard search paths or LD_LIBRARY_PATH. definitely makes sense

PR #584, opened 2026-05-11
[Minor] log ninja result

### LINKED ISSUE #565: [Feature] Add env var to print compiler output in JIT compilation
The output of nvcc/gcc is swallowed by this:

https://github.com/apache/tvm-ffi/blob/d70c905d16d16a93fedd7417e1500aaf11dfb530/python/tvm_ffi/cpp/extension.py#L583-L601

Could we add some env flags to enable some debug output here? Each time I want to inspect the gcc warnings or ptxa output, I have to manually add  these lines. I'm not sure about the design here, but I think there should be some way to access the output of the compiler.

```python
    encoding = "oem" if IS_WINDOWS else "utf-8" 
    print(status.stdout.decode(encoding))
    print(status.stderrr.decode(encoding))
```

DarkSharpness:
should fix #565

### DIFF
--- a/python/tvm_ffi/cpp/extension.py
+++ b/python/tvm_ffi/cpp/extension.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import functools
 import hashlib
+import logging
 import os
 import shutil
 import subprocess
@@ -35,6 +36,8 @@ from tvm_ffi.utils import FileLock
 
 IS_WINDOWS = sys.platform == "win32"
 BACKEND_STR = Literal["cuda", "hip"]
+
+logger = logging.getLogger(__name__)
 
 
 @functools.lru_cache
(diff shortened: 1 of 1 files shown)


tqchen (approved):

python/tvm_ffi/cpp/extension.py
```diff
@@ -600,6 +603,12 @@ def build_ninja(build_dir: str) -> None:
 
         raise RuntimeError("\n".join(msg))
 
+    if os.environ.get("TVM_FFI_LOG_BUILD") in ("1", "stdout"):
```
tqchen: given it is specific to CPP_EXTENSION, would be useful to do TVM_FFI_CPP_EXTENSION_LOG_BUILD

tqchen (commented):
minor nit

tqchen:
Thanks @DarkSharpness lgtm modulo minor nit on naming

tqchen (changes requested):

tqchen (approved):

PR #585, opened 2026-05-12
[BugFix][ROCm] Prefer upstream PyTorch DLPack API in torch extension loader

zihaomu:
Motivation:
  - If PyTorch already provides `__dlpack_c_exchange_api__`, tvm-ffi uses it directly on all backends.
  - The optional `torch-c-dlpack-ext` path is only used as a fallback for older PyTorch builds that do not provide the API.
  - The fallback extension library name is selected with backend-aware detection for CPU, CUDA, and ROCm.

  Tests are added for backend detection, ROCm short-circuit behavior when PyTorch already provides the API, and GPU tensor metadata through DLPack.

Related PR: https://github.com/tile-ai/tilelang/pull/2179, I have finished the A/B test locally.

### DIFF
--- a/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
+++ b/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
@@ -39,10 +39,14 @@ def _create_dlpack_exchange_api_capsule(ptr_as_int: int) -> Any:
     return capsule
 
 
-def _torch_extension_suffix() -> str:
-    """Return the backend suffix used by the prebuilt extension library."""
-    if torch.cuda.is_available():
-        return "rocm" if torch.version.hip is not None else "cuda"
+def _torch_extension_device(torch_module: Any) -> str:
+    """Return the torch backend name used in the optional extension library name."""
+    if torch_module.cuda.is_available():
+        if getattr(torch_module.version, "cuda", None) is not None:
+            return "cuda"
+        if getattr(torch_module.version, "hip", None) is not None:
+            return "rocm"
+        raise ValueError("Cannot determine whether to build with CUDA or ROCm.")
     return "cpu"
 
 
@@ -59,10 +63,10 @@ def load_torch_c_dlpack_extension() -> None:
         extension = "dylib"
     else:
         extension = "so"
-    suffix = _torch_extension_suffix()
+    device = _torch_extension_device(torch)
     lib_path = (
         Path(__file__).parent
-        / f"libtorch_c_dlpack_addon_torch{version.major}{version.minor}-{suffix}.{extension}"
+        / f"libtorch_c_dlpack_addon_torch{version.major}{version.minor}-{device}.{extension}"
     )
     if not lib_path.exists() or not lib_path.is_file():
         raise ImportError("No matching prebuilt torch c dlpack extension")
--- a/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
+++ b/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
@@ -42,6 +42,8 @@ def _create_dlpack_exchange_api_capsule(ptr_as_int: int) -> Any:
 def _torch_extension_device(torch_module: Any) -> str:
     """Return the torch backend name used in the optional extension library name."""
     if torch_module.cuda.is_available():
+        if getattr(torch_module.version, "cuda", None) is not None:
+            return "cuda"
         if getattr(torch_module.version, "hip", None) is not None:
             return "rocm"
         return "cuda"
--- a/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
+++ b/addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
@@ -39,6 +39,13 @@ def _create_dlpack_exchange_api_capsule(ptr_as_int: int) -> Any:
     return capsule
 
 
+def _torch_extension_suffix() -> str:
+    """Return the backend suffix used by the prebuilt extension library."""
+    if torch.cuda.is_available():
+        return "rocm" if torch.version.hip is not None else "cuda"
+    return "cpu"
+
+
 def load_torch_c_dlpack_extension() -> None:
     """Load the torch c dlpack extension based on torch version."""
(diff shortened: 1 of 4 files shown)


addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
```diff
@@ -39,6 +39,13 @@ def _create_dlpack_exchange_api_capsule(ptr_as_int: int) -> Any:
     return capsule
 
 
+def _torch_extension_suffix() -> str:
+    """Return the backend suffix used by the prebuilt extension library."""
+    if torch.cuda.is_available():
+        return "rocm" if torch.version.hip is not None else "cuda"
```
tqchen: use the same code as 
```python
def _torch_extension_device(torch_module: Any) -> str:
    """Return the torch backend name used in the optional extension library name."""
    if torch_module.cuda.is_available():
        if torch_module.version.cuda is not None:
            return "cuda"
        if torch_module.version.hip is not None:
            return "rocm"
        raise ValueError("Cannot determine whether to build with CUDA or ROCm.")
    return "cpu" 
```

tqchen (changes requested):

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -116,23 +127,18 @@ def load_torch_c_dlpack_extension() -> Any:  # noqa: PLR0912, PLR0915
         # `ctypes.CDLL` will raise an `AttributeError`.
         # Keep trying JIT
         pass
+    except OSError:
+        # A prebuilt torch-c-dlpack-ext wheel can be present but linked
```
tqchen: assuming we detect correctly, is this still necessary?

tqchen: Mainly want to know if it is something as a temp measure that we can delete after a few version cycles, if so, please add ad comment

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -94,7 +105,7 @@ def load_torch_c_dlpack_extension() -> Any:  # noqa: PLR0912, PLR0915
         import torch  # noqa: PLC0415
         import torch.version  # noqa: PLC0415
 
-        prefer_rocm_override = bool(torch.cuda.is_available() and torch.version.hip is not None)
+        prefer_rocm_override = _torch_extension_device(torch) == "rocm"
```
tqchen: would be great to confirm if upstream hipify already fix the issue https://github.com/pytorch/pytorch/pull/175432#issuecomment-3936901860 this special case was forcing local override but in principle upstream pytorch should have this version

tqchen (changes requested):

python/tvm_ffi/_optional_torch_c_dlpack.py
zihaomu: That’s a good point. I’ll double-check whether this has already been fixed upstream.

zihaomu:
Hi @tqchen. The scope of this PR has shifted from fixing ROCm-specific Torch C DLPack extension loading to making tvm-ffi avoid loading the optional out-of-tree `torch-c-dlpack-ext` package when upstream PyTorch already provides the C DLPack exchange API.

If newer PyTorch ROCm builds already expose a working `__dlpack_c_exchange_api__`, in that case, tvm-ffi should trust PyTorch and return early instead of forcing a local extension override. The local extension remains only as a compatibility fallback for older PyTorch versions

python/tvm_ffi/_optional_torch_c_dlpack.py
zihaomu: Hi @tqchen. I checked [pytorch/pytorch#175432](https://github.com/pytorch/pytorch/pull/175432#issuecomment-3936901860). It was closed because the author confirmed that the new hipify tool should handle this automatically. I also verified with torch==2.13.0.dev20260428+rocm7.2 that
  PyTorch's built-in DLPack exchange API returns the expected current HIP stream.

  So this PR no longer tries to solve the upstream PyTorch hipify issue. Instead, it removes the ROCm-specific local override and lets tvm-ffi use PyTorch's built-in DLPack API when available.

addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
```diff
@@ -39,6 +39,15 @@ def _create_dlpack_exchange_api_capsule(ptr_as_int: int) -> Any:
     return capsule
 
 
+def _torch_extension_device(torch_module: Any) -> str:
+    """Return the torch backend name used in the optional extension library name."""
+    if torch_module.cuda.is_available():
+        if getattr(torch_module.version, "hip", None) is not None:
```
tqchen, line 47: would be good to use the following pattern, mainly also detect cuda version to be robust.

```python
 if torch_module.cuda.is_available():
    if getattr(torch_module.version, "cuda", None) is not None:
       return "cuda"
    if getattr(torch_module.version, "hip", None) is not None:
       return "rocm"
  return "cuda"
return "cpu"
```

python/tvm_ffi/_optional_torch_c_dlpack.py
```diff
@@ -44,6 +44,15 @@
 logger = logging.getLogger(__name__)
 
 
+def _torch_extension_device(torch_module: Any) -> str:
+    """Return the torch backend name used in the optional extension library name."""
```
tqchen, line 48:
```python
 if torch_module.cuda.is_available():
    if getattr(torch_module.version, "cuda", None) is not None:
       return "cuda"
    if getattr(torch_module.version, "hip", None) is not None:
       return "rocm"
  return "cuda"
return "cpu"
```

tqchen (commented):
thanks, looking good after fixing some minor nits

python/tvm_ffi/_optional_torch_c_dlpack.py
zihaomu: Done

addons/torch_c_dlpack_ext/torch_c_dlpack_ext/core.py
zihaomu: Done

tqchen (approved):

PR #589, opened 2026-05-13
[BUILD] TVM_FFI_COLD_CODE / TVM_FFI_PREDICT_FALSE macros and cold-marking of error paths

tqchen:
## Summary

Adds three header-only macros (`TVM_FFI_COLD_CODE`, `TVM_FFI_PREDICT_FALSE`, `TVM_FFI_PREDICT_TRUE`) in `tvm/ffi/base_details.h` and applies them to a small audited set of error-only helpers. No CMake changes.

## What

`TVM_FFI_COLD_CODE`: `[[gnu::cold]]` on GCC/Clang, no-op on MSVC.
`TVM_FFI_PREDICT_FALSE` / `TVM_FFI_PREDICT_TRUE`: `__builtin_expect` on GCC/Clang, no-op on MSVC.

Cold-marked (error / setup / teardown only):

- `details::ErrorBuilder` ctors and the `[[noreturn]]` destructor
- `TVMFFISegFaultHandler`
- `TVMFFIInstallSignalHandler`
- `ForwardPyErrorToFFI`

C ABI exports stay hot per cross-DSO surface hygiene — `TVMFFIError*` family, `TVMFFIBacktrace`, and `SafeCallContext` setters all remain ordinary entry points. Deleters are not cold-marked: they run on every callback destruction during normal program operation, not on an error path.

`TVM_FFI_PREDICT_FALSE` is applied to `TVM_FFI_CHECK_SAFE_CALL`, `TVM_FFI_CHECK`, and ~17 error-check branches in the Python→FFI dispatchers in `tvm_ffi_python_helpers.h`.

## Mechanism

GCC and Clang emit cold-marked functions into per-TU `.text.unlikely`. The default GNU linker script's `*(.text.unlikely .text.*_unlikely .text.unlikely.*)` rule gathers them into a contiguous slot inside `.text`. No `-ffunction-sections` flag required.

Detailed measurements (binary-size matrix, isolation study, perf table, cold-cluster bounds) posted as a follow-up comment.

### DIFF
--- a/include/tvm/ffi/base_details.h
+++ b/include/tvm/ffi/base_details.h
@@ -76,6 +76,44 @@
 #define TVM_FFI_UNREACHABLE() __builtin_unreachable()
 #endif
 
+/*!
+ * \brief Mark a function as cold so the toolchain places it in a
+ *        separate cold region of `.text`. Apply to functions that only
+ *        run on error / setup / teardown paths.
+ *
+ * On GCC and Clang, expands to `[[gnu::cold]]`, which emits the
+ * function into a per-TU `.text.unlikely` section. The default GNU
+ * linker script gathers `.text.unlikely.*` into a contiguous slot
+ * inside `.text`, so cold-marked functions cluster away from hot
+ * code without any additional CMake flags. On MSVC the attribute
+ * does not exist and the macro is a no-op.
+ */
+#if defined(__GNUC__) || defined(__clang__)
+#define TVM_FFI_COLD_CODE [[gnu::cold]]
+#else
+#define TVM_FFI_COLD_CODE
+#endif
+
+/*!
+ * \brief Branch-prediction / layout hint that the condition is unlikely
+ *        to be true. Use on error-checking branches to keep the hot
+ *        fall-through contiguous and push the error-handling block to
+ *        the function tail.
+ *
+ *   if (TVM_FFI_PREDICT_FALSE(rc != 0)) { ...error... }
+ *
+ * On GCC/Clang, expands to `__builtin_expect((cond), 0)`. On MSVC,
+ * expands to `(cond)` (no equivalent builtin; modern MSVC does its own
+ * profile-driven block reordering).
+ */
+#if defined(__GNUC__) || defined(__clang__)
+#define TVM_FFI_PREDICT_FALSE(cond) (__builtin_expect(static_cast<bool>(cond), 0))
+#define TVM_FFI_PREDICT_TRUE(cond) (__builtin_expect(static_cast<bool>(cond), 1))
+#else
+#define TVM_FFI_PREDICT_FALSE(cond) (cond)
+#define TVM_FFI_PREDICT_TRUE(cond) (cond)
+#endif
+
 #define TVM_FFI_STR_CONCAT_(__x, __y) __x##__y
 #define TVM_FFI_STR_CONCAT(__x, __y) TVM_FFI_STR_CONCAT_(__x, __y)
 
--- a/include/tvm/ffi/error.h
+++ b/include/tvm/ffi/error.h
@@ -338,11 +338,13 @@ TVM_FFI_INLINE void SetSafeCallRaised(const Error& error) {
 
 class ErrorBuilder {
  public:
+  TVM_FFI_COLD_CODE
   explicit ErrorBuilder(std::string kind, std::string backtrace, bool log_before_throw)
       : kind_(std::move(kind)),
         backtrace_(std::move(backtrace)),
         log_before_throw_(log_before_throw) {}
 
+  TVM_FFI_COLD_CODE
   explicit ErrorBuilder(std::string kind, const TVMFFIByteArray* backtrace, bool log_before_throw)
       : ErrorBuilder(std::move(kind), std::string(backtrace->data, backtrace->size),
                      log_before_throw) {}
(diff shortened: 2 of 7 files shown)


tqchen:
## Detailed measurements

### Binary size — libtvm_ffi.so (stripped, Release / GCC 11.4 / ld.bfd 2.38 / x86_64 Linux)

Isolation study: 4 builds from identical source state, toggling the macros via local redefinitions to isolate which macro contributes what.

| Build                              | Stripped size | Delta       |
|------------------------------------|--------------:|------------:|
| baseline (both macros no-op)       |     1,887,800 |          —  |
| `TVM_FFI_COLD_CODE` only           |     1,834,568 |  -53,232 B  |
| `TVM_FFI_PREDICT_FALSE/TRUE` only  |     1,908,280 |  +20,480 B  |
| both (this PR)                     |     1,842,728 |  -45,072 B  |

Cold attribute alone (`current − no_cold`): -65,528 B (-3.4%).
PREDICT macros alone (`current − no_predict`): +8,184 B (+0.4%).

The cold attribute does most of the work for size. PREDICT macros are layout-only and slightly grow the binary — branch-prediction-driven basic-block reordering occasionally inserts extra jumps or duplicates code in the rearranged layout. The benefit of PREDICT is layout (hot fall-through stays contiguous), not size.

### Where the cold-attribute savings come from

Assembly diff on `structural_equal.cc.o`:

- `ErrorBuilder` ctor body: 1,045 B → 484 B per instance (-561 B per TU × 18 TUs ≈ ~10 KB at object level).
- `ErrorBuilder` dtor body: essentially unchanged (727 vs 717 bytes).
- Number of dtor call sites: identical in both builds (18 each).

The dominant mechanism is GCC's "optimized for size rather than speed" codegen on cold function bodies — `-Os`-style codegen replaces loop unrolling, vectorization, branchless tricks — not inlining suppression as I had initially conjectured. The dtor was `[[noreturn]]` and already out-of-line, so cold-marking didn't change its call-site count.

Across all `tvm_ffi_objs` `.cc.o` files: cold adds ~100 KB to `.text.unlikely` and shrinks regular `.text` by ~180 KB, for a net ~52 KB pre-link savings (which matches the stripped-binary delta).

### Cold cluster bounds — libtvm_ffi.so

`.text` total: 1.45 MiB at `0x8430..0x16aee4`.

Cold cluster: `0x8430..~0x22700` = approximately **103 KiB at the head of `.text`** (7.3% of `.text`). Contents:

- ErrorBuilder ctors and dtor
- `TVMFFISegFaultHandler`, `TVMFFIInstallSignalHandler`
- Many compiler-emitted `.cold` / `.part.N` / `.isra.N` split bodies from hot functions whose callees became transitively cold
- libbacktrace symbol-related helpers that get split-cold

Public C ABI exports verified to remain in the hot region — `TVMFFIError*` family at `0x6b820+`, `TVMFFIBacktrace` at `0x47140`, `TVMFFIObjectIncRef`/`DecRef` deeper in.

### Cython extension `core.abi3.so` (stripped)

| Build         | Stripped | `.text`     | `.eh_frame` |
|---------------|---------:|------------:|------------:|
| baseline      |  788,200 |    602,875  |     39,628  |
| with markers  |  788,200 |    603,308  |     39,732  |
| Δ             |        0 |   +433 (+0.07%) |    +104  |

Stripped on-disk size is unchanged — ELF page-alignment padding absorbs the sub-page `.text` delta. `.text` grows by 433 B (eh_frame metadata for cold-marked function epilogues + `__builtin_expect` bookkeeping).

Cold cluster on the Cython side: ~2.7 KiB (`ForwardPyErrorToFFI` plus ~10 GCC auto-cold-split `.cold` thunks from large Pyx wrappers). Much smaller than `libtvm_ffi.so`'s 103 KiB because `core.cpp` is dominated by one giant `__pyx_pymod_exec_core` (~110 KiB) — per-TU `.text.unlikely` content gets absorbed into the dominant function during link-time comdat merging.

### Performance — benchmark_dlpack.py (CPU-only subset, two trials each, median)

| Scenario                            | Baseline | With markers | Delta   |
|-------------------------------------|---------:|-------------:|--------:|
| `tvm_ffi.nop(tvm_tensor x3)`        | 112.8 ns |    112.2 ns  | -0.49%  |
| `tvm_ffi.nop.autodlpack(torch[cpu])`| 308.6 ns |    303.4 ns  | -1.69%  |
| `tvm_ffi.nop.autodlpack(numpy)`     | 939.5 ns |    926.6 ns  | -1.37%  |
| `tvm_ffi.nop+from_dlpack(torch)`    | 791.5 ns |    787.0 ns  | -0.56%  |
| `tvm_ffi.nop(int x3)`               | 133.4 ns |    133.3 ns  | -0.04%  |
| `tvm_ffi.nop()`                     |  90.4 ns |     89.2 ns  | -1.33%  |
| `tvm.__dlpack__()`                  |  84.8 ns |     84.4 ns  | -0.41%  |

All within ±2% run-to-run noise; no regression. The slight negative trend (faster) is within noise and not claimed as a real win.

### Notes on the investigation

- `-ffunction-sections` is not required for cold separation. An earlier draft of this work enabled the flag and saw an extra ~4 KB packing gain; verified that the cold cluster appears with the markers alone via the default linker script's `.text.unlikely.*` grouping rule. Dropped the flag flip from this PR to keep the change header-only.
- `--gc-sections` is intentionally not enabled. tvm-ffi has runtime registration patterns (`TVM_FFI_STATIC_INIT_BLOCK`) where symbols are referenced via runtime registries; aggressive section GC would need a separate audit of `__attribute__((used))` placement to avoid silently stripping registered globals.
- The C ABI audit revision (keeping `TVMFFIError*` and `TVMFFIBacktrace` hot) was applied after measuring an initial draft that cold-marked them. Cross-DSO callers expect public exports to be ordinary hot-tier entry points, and once an error path enters the TLS setter it should be fast — not size-optimized.

junrushao (approved):

PR #592, opened 2026-05-19
[POC][FIX] Zero cause_chain and extra_context in SafeCallContext to prevent SIGBUS on macOS arm64

oraluben:
## Problem

On macOS arm64 with tvm-ffi >= 0.1.8, when a C++ exception (e.g. InternalError from LOG(FATAL)) is thrown inside a PackedFunc and forwarded through SafeCallContext::SetRaised / SetRaisedByCstr, the process crashes with SIGBUS during cleanup.

## Root Cause

SafeCallContext::SetRaised stores an ErrorObj via ObjectPtrFromUnowned. The cause_chain and extra_context fields (added in PR #396 / v0.1.8) may contain stale non-null bytes from reused allocator memory on macOS arm64. When ~ErrorObj() runs, it calls DecRefObjectHandle on these garbage pointer values, causing SIGBUS (EXC_ARM_DA_ALIGN).

Crash address 0x6c616e7265746e49 = ASCII "Internal" (first 8 bytes of "InternalError") being dereferenced as a pointer.

Crash trace:
  SimpleObjAllocator::Handler<ErrorObjFromStd>::Deleter_
  TVMFFIObjectDecRef
  dict_dealloc → BaseException_dealloc → subtype_dealloc

## Fix

Explicitly zero cause_chain and extra_context before storing the ErrorObj in TLS, ensuring ~ErrorObj() correctly sees nullptr and short-circuits the DecRefObjectHandle calls.

## Tested

macOS arm64, tilelang 0.1.9 release
- Before: RC=-10 (SIGBUS)
- After: RC=0

### DIFF
--- a/src/ffi/error.cc
+++ b/src/ffi/error.cc
@@ -31,12 +31,25 @@ namespace ffi {
 class SafeCallContext {
  public:
   void SetRaised(TVMFFIObjectHandle error) {
+    // Explicitly zero cause_chain and extra_context before storing in TLS.
+    // On macOS arm64 the allocator may reuse memory that contains stale
+    // non-null bytes, which would cause ~ErrorObj() to call DecRefObjectHandle
+    // on a garbage pointer → SIGBUS (alignment fault).
+    auto* cell = TVMFFIErrorGetCellPtr(error);
+    cell->cause_chain = nullptr;
+    cell->extra_context = nullptr;
     last_error_ =
         details::ObjectUnsafe::ObjectPtrFromUnowned<ErrorObj>(static_cast<TVMFFIObject*>(error));
   }
 
   void SetRaisedByCstr(const char* kind, const char* message, const TVMFFIByteArray* backtrace) {
     Error error(kind, message, backtrace);
+    // Zero the newly created Error's cause_chain / extra_context as defensive
+    // measure (see comment in SetRaised above).
+    auto* cell = TVMFFIErrorGetCellPtr(
+        details::ObjectUnsafe::TVMFFIObjectPtrFromObjectRef(error));
+    cell->cause_chain = nullptr;
+    cell->extra_context = nullptr;
     last_error_ = details::ObjectUnsafe::ObjectPtrFromObjectRef<ErrorObj>(std::move(error));
   }
 


oraluben:
The current fix is not ideal for two reasons:

1. **Unconditionally clears** cause_chain and extra_context — this would incorrectly discard legitimately set values from `Error(kind, msg, bt, cause_chain, extra_context)` (the 5-arg constructor).
2. **Scattered across two call sites** (SetRaised + SetRaisedByCstr) rather than a single centralized point.

A more proper fix could be one of:
- Zero the memory in the allocator: `std::memset(data, 0, sizeof(T))` in `SimpleObjAllocator::Handler::New` after allocation and before placement new. This guarantees all POD fields are zeroed regardless of constructor, and is the most systematic approach.
- Or use member initializer lists in `ErrorObj` to avoid relying on constructor body execution.

@tqchen @junrushao would appreciate your thoughts on the right approach here.

tqchen:
i think the latest code https://github.com/apache/tvm-ffi/blob/main/include/tvm/ffi/error.h#L69 already ensures the two fields are set to zero. so updating tvm-ffi to latest should resolve the issue

oraluben:
## Updated findings — this is actually an ABI mismatch, not a missing constructor

After extensive testing, the root cause is **not** that the ErrorObj constructor fails to initialize
the fields, but rather an ABI mismatch between compile-time and runtime tvm-ffi.

### Test matrix

| tilelang build (vendored tvm-ffi) | runtime tvm-ffi | Result |
|---|---|---|
| release 0.1.9 (v0.1.3) | 0.1.7 | RC=0 ✓ |
| release 0.1.9 (v0.1.3) | 0.1.8 | SIGBUS |
| release 0.1.9 (v0.1.3) | 0.1.10 | SIGBUS |
| release 0.1.9 (v0.1.3) | 0.1.11 | SIGBUS |
| refactor branch (v0.1.11) | 0.1.10 | RC=0 ✓ |
| refactor branch (v0.1.11) | 0.1.11 | RC=0 ✓ |

### Explanation

v0.1.3: `TVMFFIErrorCell` = 56 bytes (kind, message, backtrace, update_backtrace)
v0.1.8+: `TVMFFIErrorCell` = 72 bytes (adds cause_chain, extra_context)

When tilelang is compiled against v0.1.3 headers and linked against v0.1.8+ at runtime,
internal struct layouts (TVMFFIAny, TVMFFIObject, TVMFFIErrorCell) can differ.
The ErrorObj destructor in the runtime binary accesses `cause_chain` at the v0.1.8+
offset, but the memory at that offset may contain leftover data ("Internal") from a
different allocation, not the nullptr set by the constructor.

### What this means for this PR

This POC fix (zeroing cause_chain/extra_context in SafeCallContext) is a **workaround**
that prevents the SIGBUS by defensively clearing the fields — but it doesn't address
the underlying ABI mismatch. The proper fix is for downstream projects to ensure their
vendored tvm-ffi headers match the runtime version.

That said, zeroing these fields is still a reasonable defensive measure in case any
code path ever creates ErrorObj without proper initialization.

@tqchen @junrushao thoughts on whether we should still land this as a safety net?

PR #601, opened 2026-05-29
[FEAT] Add structural visitor and typed structural walk APIs

Kathryn-cat:
## Summary

This PR adds structural traversal APIs for TVM FFI values in C++ and Python.

The main user-facing addition is `StructuralWalk` / `tvm_ffi.structural_walk`, a pytree-style walk that can visit both object-backed nodes and POD leaves (`int`, `float`, `bool`, `str`, `bytes`). It supports typed callbacks, pre/post-order traversal, child skipping, early interruption, and def-region-aware callbacks.

## Key Changes

- Added C++ `StructuralVisitor`, `VisitInterrupt`, `WalkResult`, and `StructuralWalk`.
- Added Python `tvm_ffi.structural_walk`.
- Traversal operates on `AnyView`, allowing callbacks to match objects, containers, and POD values.
- C++ callbacks dispatch on their first argument and may optionally accept a second `TVMFFIDefRegionKind`.
- Python callbacks are passed as `(type, callback)` entries, with support for grouped types like `((int, float), callback)`, a single callback catch-all, or a sequence of entries.
- Python def-region-aware callbacks are passed via `with_def_region_kind=` and receive `(value, def_region_kind)`.
- Added runtime type-index matching for Python callbacks, including subclass matching, `Any` catch-all, `Object` matching, and string/bytes storage variants.
- Custom structural visit hooks now use `AnyView` for the visited value.
- Added visit error-context propagation for object-backed nodes when traversal fails.

## Examples

Python:

```python
visited = []

def on_int(value):
    visited.append(value)
    if value == 0:
        return tvm_ffi.VisitInterrupt(value)
    return tvm_ffi.WalkResult.ADVANCE

result = tvm_ffi.structural_walk(root, (int, on_int), order="pre")
```

Grouped callback types:

```python
tvm_ffi.structural_walk(
    root,
    [
        ((int, float), on_number),
        (MyNode, on_node),
        (object, on_any),
    ],
)
```

Def-region-aware callbacks:

```python
uses = []

tvm_ffi.structural_walk(
    func,
    with_def_region_kind=(
        Var,
        lambda var, kind: uses.append(var) if kind == tvm_ffi.DefRegionKind.NONE else None,
    ),
)
```

C++:

```cpp
auto result = StructuralWalk<WalkOrder::kPreOrder>(
    root,
    [&](const Add& add) -> Expected<WalkResult> {
      ++num_adds;
      return WalkResult::Advance();
    },
    [&](const Mul& mul) -> Expected<WalkResult> {
      return WalkResult::Skip();
    });
```

C++ def-region-aware walk:

```cpp
List<Var> uses;
auto result = StructuralWalk<WalkOrder::kPreOrder>(
    func,
    [&](const VarObj* var, TVMFFIDefRegionKind kind) -> Expected<WalkResult> {
      if (kind == kTVMFFIDefRegionKindNone) {
        uses.push_back(Var(GetObjectPtr<VarObj>(var)));
      }
      return WalkResult::Advance();
    });
```

## Testing

- Added C++ tests for structural visitor traversal, def-region hooks, callback dispatch, POD leaves, object pointers, skip/interrupt behavior, def-region-aware callbacks, and error propagation.
- Added Python tests for typed callbacks, grouped callback types, mixed callback forms, def-region-aware callbacks, nested containers, `Any`/`Object` behavior, post-order traversal, skip, and interrupts.
- Verified C++ structural visitor and visit error-context tests.


include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+
+/*!
+ * \brief C++ vtable for structural visitor dispatch.
+ *
+ * The vtable gives \ref StructuralVisitorObj an explicit dispatch table instead
+ * of relying on C++ virtual methods.
+ */
+struct StructuralVisitorVTable {
+  /*! \brief Visit callback. */
+  TVMFFIAny (*visit)(StructuralVisitorObj* visitor, const Object* value) noexcept = nullptr;
```
tqchen: document the argument and return value convention e.g. Error is passed through TVMFFIAny return result is moved

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+   */
+  TVM_FFI_INLINE Expected<Optional<VisitInterrupt>> VisitExpected(const ObjectRef& value) noexcept {
+    return details::ExpectedUnsafe::MoveFromTVMFFIAny<Optional<VisitInterrupt>>(
+        (*vtable_->visit)(this, value.get()));
+  }
+
+  /*!
+   * \brief Default visit behavior: type-attr lookup, then reflection fallback.
+   *
+   * The default path first checks the structural-visit type attribute for a
```
tqchen: Impelmentation note should not be part of the documentation, just say the DefaultVisit based on registered type attr of each type

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+namespace details {
+
+/*!
+ * \brief Return true when \p result already carries a traversal-stopping state.
+ * \tparam T The Expected success type.
+ * \param result The Expected value to inspect.
+ * \return Whether \p result stores an Error or VisitInterrupt.
+ */
+template <typename T>
+TVM_FFI_INLINE bool NeedEarlyReturn(const Expected<T>& result) noexcept {
```
tqchen: naming in public header needs more clarity, StructuralVisitNeedEarlyReturn

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+ * \brief Move raw Expected storage into another Expected type.
+ * \tparam T The destination Expected success type.
+ * \tparam U The source Expected success type.
+ * \param result The source Expected value to move from.
+ * \return Expected<T> backed by the moved raw storage.
+ *
+ * \note This assumes the stored value is ABI-compatible with T, or is Error.
+ */
+template <typename T, typename U>
+TVM_FFI_INLINE Expected<T> UnsafeMoveExpected(Expected<U>&& result) {
```
tqchen: this needs to be inlined, since there is already unsafe, or move this into ExpectedUnsafe

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+    return details::ExpectedUnsafe::MoveToTVMFFIAny(
+        static_cast<StructuralWalkCallbackVisitorObj*>(self)->VisitImpl(value_ref));
+  }
+
+  /*!
+   * \brief Visit one object according to the configured walk order.
+   * \param value The object to visit.
+   * \return Interrupt state, or an error if traversal failed.
+   */
+  Expected<Optional<VisitInterrupt>> VisitImpl(const ObjectRef& value) noexcept {
```
tqchen: VisitImpl shuold be private and take const Object*

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+
+  /*!
+   * \brief Dispatch from the erased visitor pointer to the concrete walk visitor.
+   * \param self The erased structural visitor object.
+   * \param value The object to visit.
+   * \return Interrupt state, or an error if traversal failed.
+   */
+  static TVMFFIAny DispatchVisit(StructuralVisitorObj* self, const Object* value) noexcept {
+    ObjectRef value_ref(
+        details::ObjectUnsafe::ObjectPtrFromUnowned<Object>(const_cast<Object*>(value)));
```
tqchen: You can do GetRef<ObjectRef>(value) for such promotio

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+
+      if (auto ref = value.template as<RefType>()) {
+        return std::get<I>(callbacks_)(*ref);
+      }
+      return InvokeMatchingCallback<I + 1>(value);
+    }
+  }
+
+  /*! \brief Concrete callback objects. */
+  std::tuple<Callbacks...> callbacks_;
```
tqchen: consider use a single callback, and use the chain trick(better than tail recursion) to generate such a callback

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,535 @@
...
+ */
+template <WalkOrder order, typename... Callbacks>
+Expected<Optional<VisitInterrupt>> StructuralWalkExpected(const ObjectRef& root,
+                                                          Callbacks&&... callbacks) noexcept {
+  try {
+    static_assert(sizeof...(Callbacks) != 0, "StructuralWalk requires at least one callback");
+
+    using Visitor = details::StructuralWalkCallbackVisitorObj<order, std::decay_t<Callbacks>...>;
+    StructuralVisitor visitor(make_object<Visitor>(std::forward<Callbacks>(callbacks)...));
+    return visitor->VisitExpected(root);
```
tqchen: if we are VisitExpected, it shiould be noexcept so no need to try catch around VisitExpected

include/tvm/ffi/expected.h
```diff
@@ -186,9 +195,106 @@ class Expected {
...
+  }
+
+  /*!
+   * \brief Return mutable underlying Any storage.
+   * \tparam T The Expected success type.
+   * \param result The Expected value to inspect.
+   * \return Mutable reference to the raw Any storage.
+   */
+  template <typename T>
+  TVM_FFI_INLINE static Any& GetData(Expected<T>& result) noexcept {
```
tqchen: only expose what we need we only need GetData fthat returns const ref

include/tvm/ffi/expected.h
```diff
@@ -186,9 +195,106 @@ class Expected {
...
+struct ExpectedUnsafe {
+  /*!
+   * \brief Copy a raw TVMFFIAny into Expected storage.
+   * \tparam T The Expected success type.
+   * \param raw The raw FFI value to copy.
+   * \return Expected backed by a copied Any storage.
+   */
+  template <typename T>
+  TVM_FFI_INLINE static Expected<T> CopyFromTVMFFIAny(TVMFFIAny raw) {
+    Expected<T> result;
```
tqchen: This is not used, don't expose

include/tvm/ffi/expected.h
```diff
@@ -186,9 +195,106 @@ class Expected {
...
+  }
+
+  /*!
+   * \brief Copy Expected storage to a raw TVMFFIAny.
+   * \tparam T The Expected success type.
+   * \param result The Expected value to copy from.
+   * \return Raw FFI value containing a copy of the underlying Any storage.
+   */
+  template <typename T>
+  TVM_FFI_INLINE static TVMFFIAny CopyToTVMFFIAny(const Expected<T>& result) {
```
tqchen: this is not used, don't expose

tqchen (changes requested):

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * \param visitor The active visitor.
+ * \param value The object being visited.
+ * \param attr The registered type attribute for \p value.
+ * \param handled Set to true if \p attr represents a structural visit hook.
+ * \return Expected interrupt state. An error means traversal failed.
+ */
+Expected<Optional<VisitInterrupt>> TryVisitWithTypeAttr(StructuralVisitorObj* visitor,
+                                                        const ObjectRef& value, AnyView attr,
+                                                        bool* handled) {
```
tqchen: this can be inlined into the caller

tqchen (changes requested):

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+
+  /// \cond Doxygen_Suppress
+  TVM_FFI_DEFINE_OBJECT_REF_METHODS_NOTNULLABLE(VisitInterrupt, ObjectRef, VisitInterruptObj);
+  /// \endcond
+};
+
+class StructuralVisitorObj;
+
+/*!
+ * \brief C++ vtable for structural visitor dispatch.
```
tqchen: VTable ABI for StructuralVisitor Dispatch. this function provides a stable ABI for visit method

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ *
+ * The vtable gives \ref StructuralVisitorObj an explicit dispatch table instead
```
tqchen: remove since this is more on implementation note

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ * of relying on C++ virtual methods.
+ *
+ * The ``visitor`` and ``value`` arguments are borrowed for the duration of the
+ * call. The callback returns a moved ``TVMFFIAny`` containing an
```
tqchen: move this note to detailed section of visit document

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+   * \return Expected interrupt state. An error means traversal failed.
+   */
+  TVM_FFI_DLL Expected<Optional<VisitInterrupt>> DefaultVisitExpected(
+      const ObjectRef& value) noexcept;
+
+  /*!
+   * \brief Return the current def-region context.
+   * \return The active def-region kind.
+   */
+  TVM_FFI_INLINE TVMFFIDefRegionKind GetDefRegionKind() const { return def_region_mode_; }
```
tqchen: def_region_kind() this is one exception of Google C style, direct variable getter can have variable nam without underscore

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ * \brief Read an Expected success value as a compatible raw storage type.
+ * \tparam T The type to read from the underlying Any storage.
+ * \tparam U The Expected success type.
+ * \param result The Expected value to read from.
+ * \return The stored value decoded as T.
+ *
+ * \note This assumes \p result stores T-compatible Any storage, or Error.
+ */
+template <typename T, typename U>
+TVM_FFI_INLINE T UnsafeExpectedValueAs(const Expected<U>& result) {
```
tqchen: this should be under ExpectedUnsafe::ValueAs

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+   * \tparam Callbacks Callable types whose first parameter selects the dispatched
+   *                   value type.
+   * \param callbacks Callbacks to be tested in order.
+   * \return A closure of type
+   *         ``Expected<WalkResult>(const ObjectRef&)``.
+   */
+  template <typename... Callbacks>
+  static auto FromChain(Callbacks... callbacks) {
+    return [=](const ObjectRef& x) mutable -> Expected<WalkResult> {
+      if (TVM_FFI_PREDICT_FALSE(!x.defined())) {
```
tqchen: callback do not ened to handle case where x is undefined, should be handled by the implementer of s_visit, protocol is not to clal nullptr with the walk callback.

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ private:
+  /*!
+   * \brief Invoke ``callback`` when ``x`` matches its first parameter type.
+   * \tparam Callback Callable whose first parameter is a subclass reference of ObjectRef.
+   * \param callback The callback under test.
+   * \param x The value to dispatch on.
+   * \return The callback result if it matched, ``std::nullopt`` otherwise.
+   */
+  template <typename Callback>
+  static std::optional<Expected<WalkResult>> TryCallLink(Callback& callback, const ObjectRef& x) {
```
tqchen: Supprt two variant of callback, 

const ObjectRef&
const ObjectSubClass*

Can use type trait(std::is_base_of to distinguish the two and if constexpr to switch here

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+
+/*!
+ * \brief Visitor used by callback-dispatched ``StructuralWalk``.
+ *
+ * \tparam order Callback placement relative to child traversal.
+ * \tparam Dispatch Callable returning ``Expected<WalkResult>`` when invoked with
+ *                  ``const ObjectRef&``.
+ */
+template <WalkOrder order, typename Dispatch>
+class StructuralWalkCallbackVisitorObj : public StructuralVisitorObj {
```
tqchen, line 533: This is not user facing and should be part of details namespace

tqchen: locality, WalkOrder should be placed right before here andnot details

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+  /*!
+   * \brief Visit one object according to the configured walk order.
+   * \param value The object to visit.
+   * \return Interrupt state, or an error if traversal failed.
+   */
+  TVMFFIAny VisitImpl(const Object* value) noexcept {
+    try {
+      ObjectRef value_ref = GetRef<ObjectRef>(value);
+      if constexpr (order == WalkOrder::kPreOrder) {
+        auto result = dispatch_(value_ref);
```
tqchen: To keep the impl simple, the dispatch callback itself should take const Object*

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+    return static_cast<StructuralWalkCallbackVisitorObj*>(self)->VisitImpl(value);
+  }
+
+  /*!
+   * \brief Visit one object according to the configured walk order.
+   * \param value The object to visit.
+   * \return Interrupt state, or an error if traversal failed.
+   */
+  TVMFFIAny VisitImpl(const Object* value) noexcept {
+    try {
```
tqchen: move try catch to dispatch_ StructuralWalkCallbackChain::FromChain's lambda , since this code is mostly supposed to be exception free, so not wrapping try catch is cleaner.

On catch, let us consider add a macro

TVM_FFI_VISIT_END_RETURN_EXPECTED

to https://github.com/apache/tvm-ffi/blob/main/include/tvm/ffi/extra/visit_error_context.h#L156

which will do the same catch, update context and return expected value. This will enable us to do structural error chain

Note on error cain, 

auto interrupt = DefaultVisitExpected(value_ref); needs some special handling, in the case of error (predict false), we need to update the error chain to attach Object to the error chain and return, interrup directly return

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+
+/*!
+ * \brief Walk an object graph structurally and invoke typed callbacks on selected node types.
+ *
+ * The callbacks are invoked only for nodes matching the first argument type of
+ * one of the callbacks.  Callback arguments must be ``const ObjectRef&`` or
+ * ``const`` references to subclasses of ``ObjectRef``.  Callbacks are tested in
+ * order, and the first match is used.
+ *
+ * Each callback should return ``Expected<WalkResult>``:
```
tqchen: add \sa WalkResult

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ * \tparam Callbacks Callback types.
+ * \param root The root object to visit.
+ * \param callbacks Callbacks invoked for matching nodes.
+ * \return ``std::nullopt`` if traversal completed, or the interrupt returned by
+ *         a callback.
+ *
+ * \note Return type of each callback should be ``Expected<WalkResult>``.
+ */
+template <WalkOrder order, typename... Callbacks>
+Expected<Optional<VisitInterrupt>> StructuralWalkExpected(const ObjectRef& root,
```
tqchen: To keep things considtent, simply return Expected<WalkResult> and enable 

WalkResult.InterrupValue()

WalkResult.IsInterrupt()

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ * \tparam Dispatch Callable returning ``Expected<WalkResult>`` when invoked with
+ *                  ``const ObjectRef&``.
+ * \param root The root object to visit.
+ * \param dispatch The dispatcher invoked on each visited node.
+ * \return ``std::nullopt`` if traversal completed, or the interrupt returned by
+ *         the dispatcher.
+ */
+template <WalkOrder order, typename Dispatch>
+TVM_FFI_INLINE Expected<Optional<VisitInterrupt>> StructuralWalkCore(const ObjectRef& root,
+                                                                     Dispatch dispatch) noexcept {
```
tqchen: should be inlined to StructuralWalkExpected

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+namespace details {
+
+/*!
+ * \brief Return true when \p result already carries a traversal-stopping state.
+ * \tparam T The Expected success type.
+ * \param result The Expected value to inspect.
+ * \return Whether \p result stores an Error or VisitInterrupt.
+ */
+template <typename T>
+TVM_FFI_INLINE bool StructuralVisitNeedEarlyReturn(const Expected<T>& result) noexcept {
```
tqchen: is it narrowed to Expected<VisitResult>, if so it is better to narrow that

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,231 @@
...
+
+        TVMFFIDefRegionKind kind = kTVMFFIDefRegionKindNone;
+        if (field_info->flags & kTVMFFIFieldFlagBitMaskSEqHashDefNonRecursive) {
+          kind = kTVMFFIDefRegionKindNonRecursive;
+        } else if (field_info->flags & kTVMFFIFieldFlagBitMaskSEqHashDefRecursive) {
+          kind = kTVMFFIDefRegionKindRecursive;
+        }
+
+        if (auto child = field_value.as<ObjectRef>()) {
+          if (TVM_FFI_PREDICT_FALSE(kind != kTVMFFIDefRegionKindNone)) {
```
tqchen: tihs is a case where prediction should not be here, predict can only be used to rule out error handling part, both are common.

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,231 @@
...
+      return Optional<VisitInterrupt>(std::nullopt);
+    }
+
+    static reflection::TypeAttrColumn column(reflection::type_attr::kStructuralVisit);
+    AnyView attr = column[value.type_index()];
+
+    // case 1: Type-specific override registered as an opaque ABI visit function pointer.
+    if (attr.type_index() == TypeIndex::kTVMFFIOpaquePtr) {
+      auto* visit_fn = reinterpret_cast<details::FStructuralVisit>(attr.cast<void*>());
+      if (TVM_FFI_PREDICT_FALSE(visit_fn == nullptr)) {
```
tqchen: skip check here, nullptr will be et to kTVMFFINone

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,231 @@
...
+      }
+      return details::ExpectedUnsafe::MoveFromTVMFFIAny<Optional<VisitInterrupt>>(
+          (*visit_fn)(this, value.get()));
+    }
+
+    // case 2: Type-specific override registered as an ffi::Function.
+    if (attr.type_index() == TypeIndex::kTVMFFIFunction) {
+      StructuralVisitor visitor_ref(
+          details::ObjectUnsafe::ObjectPtrFromUnowned<StructuralVisitorObj>(this));
+      return attr.cast<Function>().CallExpected<Optional<VisitInterrupt>>(visitor_ref, value);
```
tqchen: I think it is possible to past const Object* directly as argument here

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,231 @@
...
+
+/*!
+ * \brief Default structural visit implementation.
+ *
+ * The default path first checks for a type-specific structural visit hook, then
+ * falls back to reflected structural fields.
+ */
+Expected<Optional<VisitInterrupt>> StructuralVisitorObj::DefaultVisitExpected(
+    const ObjectRef& value) noexcept {
+  try {
```
tqchen: this is exception free code, so do not try catch

tests/cpp/extra/test_structural_visit.cc
```diff
@@ -0,0 +1,564 @@
...
+namespace refl = tvm::ffi::reflection;
+
+using FStructuralVisit = TVMFFIAny (*)(StructuralVisitorObj* visitor, const Object* value) noexcept;
+
+// ---------------------------------------------------------------------------
+// Minimal test fixtures.
+// ---------------------------------------------------------------------------
+
+class TLeafObj : public Object {
+ public:
```
tqchen: consider consolidate sme of the objects to testing_object.h

tests/cpp/extra/test_structural_visit.cc
```diff
@@ -0,0 +1,564 @@
...
+};
+
+class ProbeVisitorObj : public StructuralVisitorObj {
+ public:
+  ProbeVisitorObj() : StructuralVisitorObj(VTable()) {}
+
+  std::vector<ObjectRef> visited;
+  std::vector<TVMFFIDefRegionKind> modes;
+  ObjectRef interrupt_on;
+  String interrupt_payload = "stop";
```
tqchen: This is likely not need atm, focus ontesting StructuralWalk

tests/cpp/extra/test_structural_visit.cc
```diff
@@ -0,0 +1,564 @@
...
+  TPair(ObjectRef lhs, ObjectRef rhs) {
+    data_ = make_object<TPairObj>(std::move(lhs), std::move(rhs));
+  }
+  TVM_FFI_DEFINE_OBJECT_REF_METHODS_NOTNULLABLE(TPair, ObjectRef, TPairObj);
+};
+
+class TIgnoredObj : public Object {
+ public:
+  ObjectRef visible;
+  ObjectRef ignored;
```
tqchen: this can be skiped for now, to test ignored, have something lke binding with a comment field that is ignored, no need to have dedicated class test

tests/cpp/extra/test_structural_visit.cc
```diff
@@ -0,0 +1,564 @@
...
+};
+
+class TDefRegionObj : public Object {
+ public:
+  ObjectRef recursive;
+  ObjectRef plain_after_recursive;
+  ObjectRef non_recursive;
+  ObjectRef plain_after_non_recursive;
+
+  TDefRegionObj(ObjectRef recursive, ObjectRef plain_after_recursive, ObjectRef non_recursive,
```
tqchen: this is not needed since the node is quite artifical, good to have some reasonably common nodes, TPair, with function def(params, body) plus comment likely suffice to cover cases of interest, cross check testing_object.h

tests/cpp/extra/test_structural_visit.cc
```diff
@@ -0,0 +1,564 @@
...
+  Optional<VisitInterrupt> result = visitor->DefaultVisit(root);
+
+  EXPECT_FALSE(result.has_value());
+  ASSERT_EQ(AsProbe(visitor)->visited.size(), 2U);
+  EXPECT_TRUE(AsProbe(visitor)->visited[0].same_as(lhs));
+  EXPECT_TRUE(AsProbe(visitor)->visited[1].same_as(rhs));
+}
+
+TEST(StructuralVisitor, DefaultVisitTraversesList) {
+  ObjectRef lhs = TLeaf("lhs");
```
tqchen: always use StructuralWalk to test, test a few cases of composed containers, no need to have one for each

tqchen (changes requested):

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+ * - \ref Advance: continue traversal, including this node's children.
+ * - \ref Skip: continue traversal but skip this node's children.
+ * - \ref Interrupt: halt the entire walk, optionally carrying a payload.
+ */
+class WalkResult : public Variant<VisitInterrupt, int32_t> {
+ public:
+  /*! \brief Internal tag value carried by a \ref Skip result. */
+  static constexpr int32_t kSkipTag = 0;
+  /*! \brief Internal tag value carried by an \ref Advance result. */
+  static constexpr int32_t kAdvanceTag = 1;
```
tqchen: advance should be 0, since it is the most common case and default motion(go first)

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,628 @@
...
+struct TypeTraits<WalkResult> : public TypeTraits<WalkResult::Storage> {
+  using Base = TypeTraits<WalkResult::Storage>;
+
+  // Decode from borrowed Any storage after a strict type check.
+  TVM_FFI_INLINE static WalkResult CopyFromAnyViewAfterCheck(const TVMFFIAny* src) {
+    return WalkResult(Base::CopyFromAnyViewAfterCheck(src));
+  }
+  // Decode by moving from owned Any storage after a strict type check.
+  TVM_FFI_INLINE static WalkResult MoveFromAnyAfterCheck(TVMFFIAny* src) {
+    return WalkResult(Base::MoveFromAnyAfterCheck(src));
```
tqchen, line 464: // Auto-convert None -> Advance
    if (src->type_index == TypeIndex::kTVMFFINone) {
      return WalkResult::Advance();
    }

implement auto convert None to Advance, so ffi function can return None (or no return which translate to none) and implies advance

do it for the few places that involves conversion

tqchen (changes requested):

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,231 @@
...
+
+        if (auto child = field_value.as<ObjectRef>()) {
+          if (TVM_FFI_PREDICT_FALSE(kind != kTVMFFIDefRegionKindNone)) {
+            result =
+                visitor->WithDefRegionKind(kind, [&]() { return visitor->VisitExpected(*child); });
+          } else {
+            result = visitor->VisitExpected(*child);
+          }
+        }
+        return TVM_FFI_PREDICT_FALSE(StructuralVisitNeedEarlyReturn(result));
```
tqchen: this does not involve branch and do not need predict

tqchen (changes requested):

tests/cpp/extra/test_structural_visit.cc
Kathryn-cat: I've significantly reduced the `test_structural_visit.cc` test cases, and moved objects to `testing_object.h`. I'd like to keep `TestVisitorObj` in the test file because it is useful for helping to check visited nodes, def region kinds, and interrupt behavior. This is a useful fixture I think.

include/tvm/ffi/extra/structural_visit.h
Kathryn-cat: `Expected<T>` is templated because it can be `Expected<WalkResult>` or `Expected<Optional<VisitInterrupt>>`, used in both cases

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,225 @@
...
+
+/*!
+ * \brief Default structural visit implementation.
+ *
+ * The default path first checks for a type-specific structural visit hook, then
+ * falls back to reflected structural fields.
+ */
+Expected<Optional<VisitInterrupt>> StructuralVisitorObj::DefaultVisitExpected(
+    AnyView value) noexcept {
+  if (TVM_FFI_PREDICT_FALSE(value.type_index() < TypeIndex::kTVMFFIStaticObjectBegin)) {
```
tqchen: don't predict here since POD handling path can be common depending on branch predictator we can choose to do the job

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,225 @@
...
+    return Optional<VisitInterrupt>(std::nullopt);
+  }
+  ObjectRef value_ref = value.cast<ObjectRef>();
```
tqchen: likely we can forward value to function without obtaining value ref(only need to cast to const Object* likely in  VisitReflectedFields

tqchen: you can take benefit fo value.type_index() directly

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,655 @@
...
+// Return from the current ABI visit function if Result stops traversal.
+// Result must be an existing Expected variable whose raw storage can be moved to TVMFFIAny.
+#define TVM_FFI_S_VISIT_MAYBE_EARLY_RETURN(Result)                                          \
+  if (TVM_FFI_PREDICT_FALSE(::tvm::ffi::details::StructuralVisitNeedEarlyReturn(Result))) { \
+    return ::tvm::ffi::details::ExpectedUnsafe::MoveToTVMFFIAny(std::move(Result));         \
+  }
+
+// Return from the current ABI visit function if Result stops traversal.
+// If Result is an Error, append Node to the visit error context before returning.
+#define TVM_FFI_S_VISIT_MAYBE_EARLY_RETURN_WITH_CONTEXT(Result, Node)                          \
```
tqchen: WITH_ERROR_CONTEXT

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,655 @@
...
+      }
+    };
+  }
+
+  /*!
+   * \brief Build a dispatcher closure over runtime typed callbacks.
+   * \param callbacks Callback entries of ``(type_index, ffi::Function)``.
+   * \return A closure of type ``Expected<WalkResult>(AnyView)``.
+   */
+  static auto FromChain(const Array<Tuple<int32_t, Function>>& callbacks) {
```
tqchen: this can go into impl of ffi.StructuralWalk

include/tvm/ffi/object.h
```diff
@@ -1085,6 +1087,43 @@ TVM_FFI_INLINE bool IsObjectInstance(int32_t object_type_index) {
   }
 }
 
+// Runtime type-index matching
+TVM_FFI_INLINE bool IsTypeIndexMatch(int32_t actual_type_index, int32_t target_type_index) {
```
tqchen: RuntimeTypeIndexMatch, add documents.

include/tvm/ffi/extra/structural_visit.h
tqchen:
```c++
Expected<Optional<VisitInterrupt>> StructuralWalkExpected(AnyView root,
                                                          const Array<Tuple<int32_t, Function>>& callbacks, 
                                                          int order) noexcept {
  static_assert(sizeof...(Callbacks) != 0, "StructuralWalk requires at least one callback");
  auto dispatch = [callbacks](AnyView x) -> Expected<WalkResult> {
      for (const auto& entry : callbacks) {
        int32_t type_index = entry.template get<0>();
        if (!::tvm::ffi::details::IsTypeIndexMatch(x.type_index(), type_index)) {
          continue;
        }
        Function fn = entry.template get<1>();
        return fn.CallExpected<WalkResult>(x);
      }
      return WalkResult::Advance();
    };
    
  if (order == 0) {
    using Visitor = details::StructuralWalkCallbackVisitorObj<0, decltype(dispatch)>;
    StructuralVisitor visitor(make_object<Visitor>(std::move(dispatch)));
      return visitor->VisitExpected(root);

  } else {
    using Visitor = details::StructuralWalkCallbackVisitorObj<1, decltype(dispatch)>;
    StructuralVisitor visitor(make_object<Visitor>(std::move(dispatch)));
    return visitor->VisitExpected(root);
  }
}

```

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  if (TVM_FFI_PREDICT_FALSE(attr.type_index() != TypeIndex::kTVMFFINone)) {
+    return Unexpected(Error("TypeError",
+                            std::string(reflection::type_attr::kStructuralVisit) +
+                                " must be an opaque function pointer or ffi.Function",
+                            ""));
+  }
+
+  return details::VisitReflectedFields(this, value.cast<const Object*>(), type_index);
```
tqchen: static Expected<Optional<VisitInterrupt>> VisitReflectedFieldsExpected(StructuralVisitorObj*, const Object* obj) noexcept;

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,234 @@
...
+ * falls back to reflected structural fields.
+ */
+Expected<Optional<VisitInterrupt>> StructuralVisitorObj::DefaultVisitExpected(
+    AnyView value) noexcept {
+  int32_t type_index = value.type_index();
+  if (type_index < TypeIndex::kTVMFFIStaticObjectBegin) {
+    return Optional<VisitInterrupt>(std::nullopt);
+  }
+
+  static reflection::TypeAttrColumn column(reflection::type_attr::kStructuralVisit);
```
tqchen: move to inline

src/ffi/extra/structural_visit.cc
```diff
@@ -0,0 +1,234 @@
...
+// StructuralVisitor Object implementation.
+// ---------------------------------------------------------------------------
+
+/*!
+ * \brief Default structural visit implementation.
+ *
+ * The default path first checks for a type-specific structural visit hook, then
+ * falls back to reflected structural fields.
+ */
+Expected<Optional<VisitInterrupt>> StructuralVisitorObj::DefaultVisitExpected(
```
tqchen: TVM_FFI_INLINE

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,655 @@
...
+    using FirstArg = std::tuple_element_t<0, typename FuncInfo::ArgType>;
+    using TSub = std::remove_cv_t<std::remove_reference_t<FirstArg>>;
+    if constexpr (std::is_same_v<TSub, AnyView>) {
+      // callback on AnyView
+      return callback(x);
+    } else if constexpr (std::is_same_v<TSub, Any>) {
+      // callback on Any
+      return callback(Any(x));
+    } else if constexpr (std::is_pointer_v<TSub> &&
+                         std::is_base_of_v<Object, std::remove_cv_t<std::remove_pointer_t<TSub>>>) {
```
tqchen: else {
}

include/tvm/ffi/reflection/accessor.h
```diff
@@ -466,6 +466,27 @@ inline constexpr const char* kSHash = "__s_hash__";
...
  */
 inline constexpr const char* kSEqual = "__s_equal__";
+/*!
+ * \brief Custom structural visitor hook (used by ``StructuralVisitor``).
+ *
+ * The hook receives the active visitor and the current object value. Opaque
+ * pointer hooks return a ``TVMFFIAny`` that stores an
+ * ``Expected<Optional<VisitInterrupt>>`` result.
+ *
+ * Value type: either an opaque function pointer to a C++ structural visit hook
```
tqchen, line 476: update docs, we are now AnyView

spectrometerHBH (approved):

include/tvm/ffi/reflection/accessor.h
Kathryn-cat: This is the `__s_visit__` hook type signature, not the visitor vtable func ptr signature, so it should be `TVMFFIAny (*)(StructuralVisitorObj* visitor, const Object* value) noexcept`. `Object*` here is reasonable because only Object have type attributes, not AnyView, and `Object*` is good for ABI-compatibility as we discussed.

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,763 @@
...
+
+namespace details {
+
+/// \cond Doxygen_Suppress
+// Return from the current ABI visit function if Result stops traversal.
+// Result must be an existing Expected variable whose raw storage can be moved to TVMFFIAny.
+#define TVM_FFI_S_VISIT_MAYBE_EARLY_RETURN(Result)                                          \
+  if (TVM_FFI_PREDICT_FALSE(::tvm::ffi::details::StructuralVisitNeedEarlyReturn(Result))) { \
+    return ::tvm::ffi::details::ExpectedUnsafe::MoveToTVMFFIAny(std::move(Result));         \
+  }
```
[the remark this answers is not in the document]
Kathryn-cat: resolved

include/tvm/ffi/extra/structural_visit.h
```diff
@@ -0,0 +1,763 @@
...
+  if (TVM_FFI_PREDICT_FALSE(::tvm::ffi::details::StructuralVisitNeedEarlyReturn(Result))) {    \
+    if (TVM_FFI_PREDICT_FALSE((Result).type_index() == ::tvm::ffi::TypeIndex::kTVMFFIError)) { \
+      if ((Node).type_index() >= ::tvm::ffi::TypeIndex::kTVMFFIStaticObjectBegin) {            \
+        ::tvm::ffi::Error _tvm_ffi_visit_err_ = (Result).error();                              \
+        ::tvm::ffi::details::UpdateVisitErrorContext(_tvm_ffi_visit_err_,                      \
+                                                     (Node).cast<::tvm::ffi::ObjectRef>());    \
+      }                                                                                        \
+    }                                                                                          \
+    return ::tvm::ffi::details::ExpectedUnsafe::MoveToTVMFFIAny(std::move(Result));            \
+  }
```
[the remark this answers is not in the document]
Kathryn-cat: resolved

python/tvm_ffi/structural.py
```diff
@@ -231,3 +290,260 @@ def __hash__(self) -> int:
...
+        for slot in callback_slots:
+            if callable(slot):
+                callback_entries.append((Any, slot))
+            elif isinstance(slot, tuple) and len(slot) == 2 and callable(slot[1]):
+                add_callback_entry(slot)
+            elif isinstance(slot, Sequence):
+                for callback in slot:
+                    assert isinstance(callback, tuple)
+                    assert len(callback) == 2 and callable(callback[1])
+                    add_callback_entry(callback)
```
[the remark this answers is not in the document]
Kathryn-cat: resolved

tqchen (approved):

cyx-6 (approved):

tqchen:
Thanks @Kathryn-cat , this PR is merged

PR #607, opened 2026-06-05
[FIX] Drop test_empty_tensor_attributes (numpy dlpack stride change)

tqchen:
[FIX] Drop test_empty_tensor_attributes (numpy dlpack stride change for empty arrays)

NumPy silently changed how it exports strides for zero-element arrays via
`__dlpack__`. For a shape `(4, 0, 4)` array NumPy previously reported
strides `(0, 4, 1)`, but now reports `(0, 0, 0)`. Either choice is
arguably valid — any stride values are consistent when the array has zero
elements — but `tvm_ffi.from_dlpack` faithfully forwards whatever the
upstream exporter provides. The test was asserting a NumPy-side
implementation detail that has since changed, so it is removed rather than
patched. Deciding what the correct empty-array stride convention should be
is a separate question not in scope for this fix.

---

Also: remove `ndim` / `numel` / `size` / `is_contiguous` from the top-level
autosummary in `docs/reference/python/index.rst`. They are methods on
`tvm_ffi.Tensor` (added in #604), not top-level attributes, and they
are already documented via the `Tensor` autoclass — so a previous
attempt to qualify them as `Tensor.ndim` etc. produced duplicate-
object-description warnings. Removing them entirely fixes both the
original "failed to import" warnings and the subsequent duplicate
warnings.

---

Also: `Tensor.is_contiguous()` now returns True early when
`numel() == 0`, in both the C++ implementation
(`include/tvm/ffi/container/tensor.h`) and the Python (Cython)
side. This aligns with numpy/pytorch — an empty tensor is trivially
contiguous regardless of what strides the producer (e.g. numpy
2.3+ via `__dlpack__`) reports. See review comment
https://github.com/apache/tvm-ffi/pull/607#issuecomment-4634436950
for the underlying motivation.

### DIFF
--- a/docs/reference/python/index.rst
+++ b/docs/reference/python/index.rst
@@ -48,10 +48,6 @@ Tensor
   Device
   DLDeviceType
   device
-  ndim
-  numel
-  size
-  is_contiguous
 
 Function
 ~~~~~~~~
--- a/include/tvm/ffi/container/tensor.h
+++ b/include/tvm/ffi/container/tensor.h
@@ -56,6 +56,11 @@ inline bool IsDirectAddressDevice(const DLDevice& device) {
  */
 inline bool IsContiguous(const DLTensor& arr) {
   if (arr.strides == nullptr) return true;
+  // An empty tensor (numel == 0) is trivially contiguous regardless of strides,
+  // matching NumPy/PyTorch semantics.
+  for (int32_t i = 0; i < arr.ndim; ++i) {
+    if (arr.shape[i] == 0) return true;
+  }
   int64_t expected_stride = 1;
   for (int32_t i = arr.ndim; i != 0; --i) {
     int32_t k = i - 1;
--- a/python/tvm_ffi/cython/tensor.pxi
+++ b/python/tvm_ffi/cython/tensor.pxi
@@ -304,9 +304,14 @@ cdef class Tensor(CObject):
         """True if the Tensor is C-contiguous (row-major), False otherwise."""
         if self.cdltensor.strides == NULL:
             return True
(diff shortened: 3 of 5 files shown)


kainzhong:
Ah sorry I didn't notice I was using numpy with a outdated version when I run my tests locally. I just updated it and now I see it returns strides with all zeros.

Though I wonder for shape (4, 0, 4) if numpy reports strides (0, 0, 0) then my current implementation will say `is_contiguous()=False` since the expected stride (0, 4, 1) doesn't match (0, 0, 0) from numpy, which differs from numpy and pytorch since they would see an empty tensor as contiguous, but this implementation matches the C++ implementation in `include/tvm/ffi/container/tensor.h`. Do we need to also fix both of them by return True early when `numel()=0`?

tqchen:
@kainzhong seems we should match the pytorch behavior for empty strides, i will update in this patch

spectrometerHBH (approved):

docs/reference/python/index.rst
```diff
@@ -48,10 +48,6 @@ Tensor
   Device
   DLDeviceType
   device
-  ndim
-  numel
-  size
-  is_contiguous
```
junrushao, line 54: Why do we want to drop documentation of those 4 fields?

tqchen: previously there was a dup warning in sphinx, let me cross check

tqchen: This is doc page for the root index, and we only need Tensor, the functions are part of Tensor methods and will appear in Tensor page

PR #610, opened 2026-06-06
[FIX] is_contiguous returns True for empty tensors (numel() == 0)

tqchen:
[FIX] is_contiguous returns True for empty tensors (numel() == 0)

An empty tensor — one where any shape dimension is 0 — has no elements
and its stride values carry no meaning.  The previous implementation
iterated over strides and returned False when it encountered a stride
that did not match the expected row-major stride (e.g. strides all-zero
for a shape like (4, 0, 4)), which is incorrect.

This commit adds an early return when numel == 0 to both:
- C++: `IsContiguous(const DLTensor&)` in
  `include/tvm/ffi/container/tensor.h` (checks any shape dim == 0
  before the stride loop)
- Python/Cython: `is_contiguous()` in
  `python/tvm_ffi/cython/tensor.pxi` (same shape-zero scan before the
  stride loop)

Semantics now match NumPy and PyTorch: an empty tensor is trivially
contiguous.

Tests added:
- C++: `TEST(Tensor, EmptyTensorIsContiguous)` in
  `tests/cpp/test_tensor.cc` — constructs a strided empty tensor with
  strides (0,0,0) and asserts `IsContiguous() == true`.
- Python: `test_empty_tensor_attributes` in
  `tests/python/test_tensor.py` — existing test updated to use the
  correct numpy-reported strides (0,0,0) for shape (4,0,4) and now
  asserts `is_contiguous() == True`.

Closes https://github.com/apache/tvm-ffi/pull/607#issuecomment-4634436950

### DIFF
--- a/include/tvm/ffi/container/tensor.h
+++ b/include/tvm/ffi/container/tensor.h
@@ -56,6 +56,11 @@ inline bool IsDirectAddressDevice(const DLDevice& device) {
  */
 inline bool IsContiguous(const DLTensor& arr) {
   if (arr.strides == nullptr) return true;
+  // An empty tensor (numel == 0) is trivially contiguous regardless of strides,
+  // matching NumPy/PyTorch semantics.
+  for (int32_t i = 0; i < arr.ndim; ++i) {
+    if (arr.shape[i] == 0) return true;
+  }
   int64_t expected_stride = 1;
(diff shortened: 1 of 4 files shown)


tqchen:
Folding this into #607. Closing in favor of that PR.

PR #621, opened 2026-06-14
[PY] Distinguish FFI input and output annotations

Ubospica:
## Summary
- Add separate input/output rendering APIs for `TypeSchema` so stubgen can widen callable and constructor inputs without weakening returned or field types.
- Add a `__ffi_convert_type_schema__` TypeAttr hook for object conversions and wire `py_class`/`c_class` dataclass transform converter metadata.
- Generate `type_schema` metadata for Python-defined `@method` TypeMethods so reflection-driven stubgen can render their signatures.
- Cover container asymmetry, object convert schema recursion, stubgen signatures, partial type maps, `@method` metadata, and dataclass metadata with regression tests.
- Update the PR CI workflow to avoid Apache-disallowed third-party setup actions and run pre-commit without syncing the project during lint.

## Validation
- pre-commit hooks during commits, including ruff, ty, clang-format, cython-lint, CMake lint/format, and workflow YAML checks
- `python -m pytest tests/python/test_stubgen.py::test_py_class_method_metadata_renders_stub_signature tests/python/test_typed_method.py -q`
- `python -m pytest tests/python/test_stubgen.py -q`
- `CUDA_VISIBLE_DEVICES="" python -m pytest tests/python -q` (`2322 passed, 18 skipped, 2 xfailed`)
- Manual output dumped under `tmp/2026-06-14-update-ffi/manual_check.out`
- GitHub CI is green for lint, docs, Ubuntu, Ubuntu ARM, macOS, and Windows


Ubospica:
CI is green now, and the two Gemini review comments have been addressed and resolved:

- quoted `frozenset[int]` in the Cython annotation for Python 3.8 compatibility
- fixed input type-map defaults for partially populated `ty_map` and added a regression test

cc @tqchen @junrushao for review when you have time.

.github/workflows/ci_test.yml
```diff
@@ -71,11 +71,16 @@ jobs:
...
           fetch-depth: 0
           fetch-tags: true
-      - name: Set up uv
-        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098  # v7.3.1
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+      - name: Install uv
+        run: python -m pip install uv
```
junrushao: Curious why this change? I supposed apache does allow us to use `astral/setup-uv`?

Ubospica: I tried restoring `astral-sh/setup-uv` to simplify the workflow, but the PR workflow immediately hit `startup_failure` before any jobs were created. So it looks like this action is still blocked for this PR context

junrushao: That's weird. Does those hashes appear in Apache's allowlist: https://github.com/apache/infrastructure-actions/blob/main/actions.yml#L72

tqchen:
FYI astra issue should be fixed by https://github.com/apache/tvm-ffi/pull/624

tqchen:
@junrushao please take another look if have time, i think we can bring it in if the need aligns with what we want

Ubospica:
Yes, I think it is ready now. cc @junrushao feel free to check it when you get time.

tqchen (approved):

PR #623, opened 2026-06-17
[FEAT][RUST]Add tvm_ffi::Map in Rust

Seven-Streams:
This PR adds `tvm_ffi::Map` in Rust.

`tvm_ffi::MapObj` in Rust and the `ffi::MapObj` in C++ share the same memory layer at run time(when `TVM_FFI_DEBUG_WITH_ABI_CHANGE` is `False`). and `tvm_ffi::MapObj` will call the corresponding method of `ffi::Obj` in C++ through global function call directly in most cases.

### DIFF
--- a/rust/tvm-ffi/src/collections/map.rs
+++ b/rust/tvm-ffi/src/collections/map.rs
@@ -1,0 +1,469 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+//! Immutable [`Map`] container backed by the C++ `ffi.Map` object.
+//!
+//! The `MapObj` header mirrors C++ `MapBaseObj`, so [`Map::len`] reads `size`
+//! directly (no FFI). The hash-table storage itself is an implementation detail,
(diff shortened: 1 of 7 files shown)


tqchen:
cc @Ubospica please help review

tqchen (approved):

rust/tvm-ffi/src/collections/map.rs
```diff
@@ -0,0 +1,474 @@
...
+unsafe impl<K, V> AnyCompatible for Map<K, V>
+where
+    K: AnyCompatible,
+    V: AnyCompatible,
+{
+    fn type_str() -> String {
+        format!("Map<{}, {}>", K::type_str(), V::type_str())
+    }
+
+    unsafe fn check_any_strict(data: &TVMFFIAny) -> bool {
```
Ubospica, line 438: https://github.com/apache/tvm-ffi/blob/main/include/tvm/ffi/container/map_base.h#L1592

CheckAnyStrict loops over all elements in the map. Shall we align them?

rust/tvm-ffi/src/collections/map.rs
```diff
@@ -0,0 +1,474 @@
...
+/// and never allocates a `MapObj` itself (the C++ runtime allocates the larger
+/// `Small`/`DenseMapBaseObj` subclass), so the trailing field is never touched.
+/// The layout also assumes `TVM_FFI_DEBUG_WITH_ABI_CHANGE` is off (the default):
+/// with that debug flag set, `MapBaseObj` gains a *leading* `state_marker` field
+/// that would shift every offset below.
+#[repr(C)]
+#[derive(Object)]
+#[type_key = "ffi.Map"]
+#[type_index(TypeIndex::kTVMFFIMap)]
+pub struct MapObj {
```
Ubospica, line 60: https://github.com/apache/tvm-ffi/blob/230ae6c86ababb170baac6c485214e0a5af30702/include/tvm/ffi/container/map_base.h#L194-L197

TVM_FFI_DEBUG_WITH_ABI_CHANGE will modify the memory layout. Maybe we should either raise a warning when it's on, or support that in rust

tqchen: this one is  fine, since we do not have it on, maybe we should even consider emove it

rust/tvm-ffi/src/collections/map.rs
Seven-Streams: Aligned now. The original reason is that the logic of `Map` hasn't been fully rewritten in Rust yet, and it will introduce some overhead. But we can update it later.

Ubospica (approved):
Now it looks great to me!

PR #630, opened 2026-06-18
[FEAT][RUST]Add tvm_ffi::optional in Rust

Seven-Streams:
This PR supports `ffi::optionalPod`, `ffi::optionalStr` in `Rust`. It keeps the same memory layout as C++'s ffi::optional.

### DIFF
--- a/include/tvm/ffi/optional.h
+++ b/include/tvm/ffi/optional.h
@@ -29,7 +29,6 @@
 #include <tvm/ffi/object.h>
 #include <tvm/ffi/string.h>
 
-#include <cstdint>
 #include <optional>
 #include <string>
 #include <utility>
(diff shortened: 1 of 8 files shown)


rust/tvm-ffi/src/option.rs
```diff
@@ -0,0 +1,328 @@
...
+/// representation is byte-identical to the C++ field type (`i32` ↔ `int32_t`,
+/// `f64` ↔ `double`, …), so the mirror can overlay the C++ `std::optional<T>`.
+pub unsafe trait OptionalPod: Copy {}
+
+/// Layout-mirror of `std::optional<T>`: `{ T value @0; bool engaged @sizeof(T) }`.
+#[repr(C)]
+struct OptionalCell<T: OptionalPod> {
+    value: MaybeUninit<T>,
+    engaged: bool,
+}
```
Ubospica, line 84: We are currently assuming the memory layout from C++ side.

Shall we add a static assert in C++ side, like

```
static_assert(sizeof(ffi::Optional<int32_t>) == 8)
```

To avoid potential compilation errors?

rust/tvm-ffi/src/collections/optional.rs
```diff
@@ -0,0 +1,328 @@
...
+        let cell = unsafe { &mut *self.cell.get() };
+        match value {
+            Some(v) => {
+                cell.value = MaybeUninit::new(v);
+                cell.engaged = true;
+            }
+            None => cell.engaged = false,
+        }
+    }
+}
```
Ubospica: pub fn set(&mut self, value: Option<T>) {
        match value {
            Some(v) => {
                self.cell.value = MaybeUninit::new(v);
                self.cell.engaged = true;
            }
            None => self.cell.engaged = false,
        }
    }


Shall we just use a mutable set?

rust/tvm-ffi/src/option.rs
Seven-Streams: yeah added

include/tvm/ffi/optional.h
```diff
@@ -476,6 +477,22 @@ struct TypeTraits<Optional<T>> : public TypeTraitsBase {
     return R"({"type":"Optional","args":[)" + details::TypeSchema<T>::v() + "]}";
   }
 };
+
+/// \cond Doxygen_Suppress
+// The Rust binding (rust/tvm-ffi/src/option.rs) mirrors Optional<T> in place as
+// `{ T value; bool engaged; }`; fail early if a toolchain's std::optional deviates from that.
+namespace details {
```
tqchen: move static assert into test

tqchen (approved):

Seven-Streams:
#

PR #642, opened 2026-06-23
[PY] Support pickling ffi.String without cached FFI object

Ubospica:
## Summary
- Add a custom pickle reduction for `tvm_ffi.core.String` so only the Python string payload is serialized.
- Drop the hidden `_tvm_ffi_cached_object` on unpickle, since `String` is effectively `str` with optional FFI backing.
- Add a regression test for long FFI strings that carry a cached backing object.

### DIFF
--- a/python/tvm_ffi/cython/string.pxi
+++ b/python/tvm_ffi/cython/string.pxi
@@ -70,6 +70,9 @@ class Bytes(bytes, PyNativeObject):
         val._tvm_ffi_cached_object = None
         return val
 
+    def __reduce_ex__(self, protocol):
+        return (type(self), (bytes(self),))
+
     # pylint: disable=no-self-argument
     def __from_tvm_ffi_object__(cls, obj):
         """Construct from a given tvm object."""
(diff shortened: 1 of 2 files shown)


Ubospica:
Addressed the Gemini comments in db208dc:

- Added `Bytes.__reduce_ex__` so pickling `tvm_ffi.core.Bytes` only persists the Python `bytes` payload and drops `_tvm_ffi_cached_object` on unpickle.
- Added a regression test for cached FFI `Bytes` returned by `testing.echo`.
- Kept `Bytes.__slots__` unchanged because Python raises `TypeError: nonempty __slots__ not supported for subtype of 'bytes'`; the reduction hook is sufficient to avoid serializing the cached FFI object.

Local validation:
- `uv pip install --reinstall --verbose . && .venv/bin/pytest -vvs tests/python/test_string.py`
- commit hooks passed after fixing the local pre-commit cache dependency versions for mypy (`numpy<2.0`, `pytest<8`) to match the repo's Python 3.9 mypy target.

Ubospica:
Superseded by #643, which is based on the latest  and includes the Gemini Bytes review fixes without the dirty merge state.

Ubospica:
Closing in favor of #643.

PR #643, opened 2026-06-23
[PY] Support pickling FFI String and Bytes

Ubospica:
## Summary
- Add pickle reductions for `tvm_ffi.core.String` and `tvm_ffi.core.Bytes` so only the Python payload is serialized.
- Ensure unpickled values drop the hidden `_tvm_ffi_cached_object` backing object.
- Add regression coverage for cached FFI String/Bytes returned from `testing.echo`.

## Notes
- This is a clean branch based on the latest `apache/main` and supersedes #642, which became dirty after main advanced.
- `Bytes.__slots__` is intentionally unchanged because Python does not support non-empty `__slots__` on `bytes` subclasses.

## Test plan
- `uv pip install --reinstall --verbose . && .venv/bin/pytest -vvs tests/python/test_string.py`
- `PYTHONPATH=/raid/user_data/yixind/miniforge3/lib/python3.12/site-packages pre-commit run --files python/tvm_ffi/cython/string.pxi tests/python/test_string.py`

### DIFF
--- a/python/tvm_ffi/cython/string.pxi
+++ b/python/tvm_ffi/cython/string.pxi
@@ -58,9 +58,7 @@ class String(str, PyNativeObject):
         return val
 
     def __reduce_ex__(self, protocol):
-        state = self.__dict__.copy() if hasattr(self, "__dict__") else {}
-        state.pop("_tvm_ffi_cached_object", None)
-        return (type(self), (str(self),), state or None)
+        return (str, (str(self),))
 
     # pylint: disable=no-self-argument
     def __from_tvm_ffi_object__(cls, obj: Any) -> "String":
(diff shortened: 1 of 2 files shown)


Ubospica:
Addressed the latest Gemini comments in cb9272a:

- `String.__reduce_ex__` and `Bytes.__reduce_ex__` now return state as the third reduction element.
- The state keeps ordinary instance attributes and drops only `_tvm_ffi_cached_object`.
- Added tests for `String` subclass state and `Bytes` instance state preservation.

Ubospica:
Updated the pickle semantics in a030c1f per requested behavior:

- `pickle.loads(pickle.dumps(tvm_ffi.core.String(...)))` now returns raw Python `str`.
- `pickle.loads(pickle.dumps(tvm_ffi.core.Bytes(...)))` now returns raw Python `bytes`.
- Tests now assert exact native types for both directly constructed values and cached FFI-returned values.

This intentionally drops FFI wrapper type/subclass state across pickle, matching the desired raw Python payload behavior.

tqchen (approved):

PR #657, opened 2026-07-06
[FFI] Back Optional and Variant by TVMFFIAny for a stable ABI layout

tqchen:
This PR makes `tvm::ffi::Optional<T>` and `tvm::ffi::Variant<...>` layout stable across the contained type `T` by backing them with `Any`. Note that `Optional<T>` still falls back to `std::optional<T>` when `T` does not support Any storage.

This helps stabilize the future object-layout ABI of `Optional` and `Variant`.


tqchen:
NOTE: this PR do not break the C ABI so it. But it does impact the extra json Stringfy ABI, will followup on downstream tvm update. This is one step forward as previously there is no guarantee on the layout but now we have one

junrushao (approved):

PR #658, opened 2026-07-07
[ORCJIT] Add shared session and high-level load_module

cyx-6:
Adds a process-wide shared execution session and a unified `load_module` on the `tvm_ffi_orcjit` addon, mirroring core `tvm_ffi.load_module`.

## Changes

- **`default_session()`** — leaked, process-wide shared `ExecutionSession`; settable ORC-runtime-path registry resolved on first use.
- **`ExecutionSession.load_module(objects, name)`** — loads one or more objects (path `str`/`Path` or in-memory bytes; single item or list) into one fresh JITDylib, injects context symbols eagerly, expands any embedded library binary into an import tree, and returns a plain `tvm_ffi.Module`.
- **Recursive session lock** serializing all compound JITDylib ops (create / add / lookup+init / set-link-order / teardown) for safe shared-session use across threads; the resolved-call hot path stays lock-free.
- **Namespace rename** — registered globals and the object key move to `ffi_orcjit`; `set_link_order` kept but dropped from the advertised surface.
- **`cpp.build`** — copy a single object through instead of `ld -r`, which some system linkers use to stamp a nonzero `sh_addr` on zero-size sections (e.g. `.note.GNU-stack`), producing a JITLink block that overlaps in `EHFrameEdgeFixer` and breaks C++ (eh_frame) loads.

Also folds in earlier addon cleanups (hide static-linked archive symbols; numpydoc docstrings).

## Testing

New `tests/test_session_load_module.py` covers input shapes, eager context, drop, embedded-binary expansion, and concurrent shared-session use. Full addon suite green (154 passed, 3 skipped) on Linux aarch64.


addons/tvm_ffi_orcjit/src/ffi/orcjit_dylib.cc
```diff
@@ -97,18 +202,75 @@ ORCJITDynamicLibraryObj::~ORCJITDynamicLibraryObj() {
...
+
+void ORCJITDynamicLibraryObj::AddObjectBuffer(std::unique_ptr<llvm::MemoryBuffer> buffer) {
+  // Compound topology op: adding an object mutates shared linker state and must
+  // not race concurrent create / add / lookup / teardown on this session.
+  std::lock_guard<std::recursive_mutex> lock(session_->mutex());
+  TVM_FFI_ORCJIT_LLVM_CALL(jit_->addObjectFile(*dylib_, std::move(buffer)));
   context_symbol_refreshed_.store(false, std::memory_order_release);
 }
 
+Module ORCJITExecutionSessionObj::LoadModule(const Array<Any>& objects, const String& name) {
```
tqchen: ffi::Variant<String, Bytes>

addons/tvm_ffi_orcjit/src/ffi/orcjit_session.cc
```diff
@@ -203,9 +204,53 @@ ORCJITExecutionSession::ORCJITExecutionSession(const std::string& orc_rt_path,
...
+
+namespace {
+// Small settable registry for the process-wide default ORC runtime path, using
+// the leaked-singleton idiom (never torn down during interpreter finalization).
+struct DefaultOrcRuntimePath {
+  std::mutex mutex;
+  String path;
+
+  static DefaultOrcRuntimePath* Global() {
+    static auto* inst = new DefaultOrcRuntimePath();
```
tqchen: this is not needed, maybe we can simply have python register the tvm_ffi_orcjit.DefaultOrcRuntimePath function

addons/tvm_ffi_orcjit/src/ffi/orcjit_session.h
```diff
@@ -58,13 +80,60 @@ class ORCJITExecutionSessionObj : public Object {
...
+   *
+   * A leaked, never-destroyed singleton so multiple callers in one process
+   * share one LLVM ExecutionSession — hence process symbols, the slab arena,
+   * and cross-library linking. Never torn down (interpreter finalization could
+   * otherwise call back into the host language during teardown). Created on
+   * first call using \ref GetDefaultOrcRuntimePath.
+   *
+   * \return The shared execution session.
+   */
+  static ORCJITExecutionSession Default();
```
tqchen: GlobalDefault

addons/tvm_ffi_orcjit/src/ffi/orcjit_session.cc
```diff
@@ -203,9 +204,53 @@ ORCJITExecutionSession::ORCJITExecutionSession(const std::string& orc_rt_path,
...
+      new ORCJITExecutionSession(GetDefaultOrcRuntimePath().operator std::string(), 0);
+  return *inst;
+}
+
 ORCJITDynamicLibrary ORCJITExecutionSessionObj::CreateDynamicLibrary(const String& name) {
   TVM_FFI_CHECK(jit_ != nullptr, InternalError) << "ExecutionSession not initialized";
 
+  // Compound topology op — serialize against concurrent create / add / lookup /
+  // teardown on this shared session (lock order: session lock first).
+  std::lock_guard<std::recursive_mutex> lock(session_mutex_);
```
tqchen: do we need recursive muted here? seems normal one is fine

addons/tvm_ffi_orcjit/src/ffi/orcjit_session.h
```diff
@@ -58,13 +80,60 @@ class ORCJITExecutionSessionObj : public Object {
...
+   * Compound topology operations (create / add object / set link order /
+   * symbol lookup+init / teardown) are individually locked by LLVM but not
+   * atomic with respect to one another; a shared session driven by multiple
+   * threads (the FFI call path releases the GIL) must serialize them. Recursive
+   * because they nest on one thread (create → JIT'd constructor → function
+   * lookup). The already-resolved-call hot path never takes it.
+   *
+   * \return Reference to the recursive session mutex.
+   */
+  std::recursive_mutex& mutex() { return session_mutex_; }
```
tqchen: likely no need to expose

addons/tvm_ffi_orcjit/src/ffi/orcjit_dylib.cc
```diff
@@ -225,15 +398,22 @@ static void RegisterOrcJITFunctions() {
   refl::ObjectDef<ORCJITExecutionSessionObj>();
 
   refl::GlobalDef()
-      .def("orcjit.ExecutionSession",
+      .def("ffi_orcjit.ExecutionSession",
            [](const std::string& orc_rt_path, int64_t slab_size_bytes) {
              return ORCJITExecutionSession(orc_rt_path, slab_size_bytes);
            })
-      .def("orcjit.ExecutionSessionCreateDynamicLibrary",
+      .def("ffi_orcjit.DefaultSession", []() { return ORCJITExecutionSessionObj::Default(); })
```
tqchen: consider write tvm_ffi_orcjit as full path name

addons/tvm_ffi_orcjit/src/ffi/orcjit_dylib.cc
```diff
@@ -198,6 +366,11 @@ Optional<Function> ORCJITDynamicLibraryObj::GetFunction(const String& name) {
   // TVM-FFI exports have __tvm_ffi_ prefix
   std::string symbol_name = symbol::tvm_ffi_symbol_prefix + std::string(name);
 
+  // Compound lookup+init: serialize against concurrent session operations
+  // (lock order: session lock → refresh mutex → LLVM lookup). The returned
+  // Function, once resolved, is invoked lock-free on the hot path.
```
tqchen, line 339: i think we can further simplify if we assume ORCJITDynamicLibrary is unit operation, we explicitly do refresh in LoadModule, then GetFunction do not need to do that, avoid exposing functions like SetLinkOrder AddObjectBytes, leave those as internal

tqchen (changes requested):

tqchen (approved):

PR #659, opened 2026-07-10
[ORCJIT] Add test cases for test_string_return

cchung100m:
Hi Committers,

This PR adds test cases for `test_string_return`. Any suggestions would be appreciated if you are available.


tqchen:
I am not sure the intention about the PR, since our goal is to have reasonable but minimal coverage is here. i think we should close for now

cchung100m:
Hi @tqchen,
Thank you for the quick response and the valuable feedback. I'll keep an eye on the codebase and continue looking for other opportunities to contribute. Thanks again!

PR #663, opened 2026-07-13
refactor(dataclasses)!: separate field resolution and native lifetimes

junrushao:
Architecture:
- Resolve annotations from each declaring class so forward references, recursive classes, and mixins no longer depend on merged inherited annotations.
- Centralize typed reflection conversion in TypeAttr and ObjectDef while native object-valued fields receive explicit placement construction and destruction.
- Carry nested field paths through conversion failures and compute init-only properties after FFI-backed construction.

Public Interfaces:
- Add init_property for values derived once after dataclass initialization.
- Add reflection conversion hooks for type attributes and object definitions.
- Make py_class use tree structural equality by default; callers can still opt out explicitly with structural_eq=None.

UI/UX:
- None.

Behavioral Changes:
- Dataclass field resolution now respects the class that declares each field.
- Native field conversion reports the complete path to a failing nested value.
- Device strings and reflected ancestor types use the shared conversion path.

Docs:
- Update the structural equality guide for the new py_class default.
- Keep lower-level reflection and init_property behavior documented through API comments and focused tests; a broader narrative guide can follow separately.

Tests:
- Clean editable C++/Cython build completed successfully.
- 1103 focused dataclass, structural, converter, typed-method, and function tests passed; 2 expected failures remained expected.
- Scoped pre-commit hooks passed for all files in this boundary.

Untested Edge Cases:
- Rust bindings, Windows/MSVC, and the full cross-platform test matrix were not run for this intermediate boundary.
- Stub regeneration exposes pre-existing drift in mutable-container annotations and an omitted test helper, so those unrelated generated edits are deferred.

BREAKING CHANGE: py_class now defaults to tree structural equality and field resolution follows each declaring class. Extensions that depend on the previous default or annotation-merging behavior must opt out or update their declarations and rebuild C++/Cython artifacts.


python/tvm_ffi/_dunder.py
```diff
@@ -104,6 +114,22 @@ def __init__(self: Any, *args: Any, **kwargs: Any) -> None:
...
+    precedence over parent definitions of the same name (first-seen wins).
+    """
+    funcs: list[tuple[str, Callable[..., Any]]] = []
+    seen: set[str] = set()
+    for cls in type_cls.__mro__:
+        for name, func in cls.__dict__.get("__ffi_init_property_funcs__", {}).items():
+            if name not in seen:
+                funcs.append((name, func))
+                seen.add(name)
+    return funcs
```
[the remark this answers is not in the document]
junrushao: Fixed in 36ee686. _collect_init_property_funcs now walks the reversed MRO into an insertion-ordered dict, so parent properties are initialized before child properties while child definitions still replace inherited ones. I added regressions for a child property depending on a parent property and for child override selection. The complete test_dataclass_py_class.py file passes (386 passed, 1 version-gated skip on Python 3.13), and the Python 3.14 TestInitProperty group passes 9/9. Thanks!

python/tvm_ffi/utils/descriptors.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
+
+        ann = self._return_annotation if self._return_annotation is not None else Any
+        # Inject into the owner's own __annotations__ so on_fields_resolved
+        # processes this name as a typed field.
+        if "__annotations__" not in owner.__dict__:
+            owner.__annotations__ = {}
+        owner.__annotations__[name] = ann
```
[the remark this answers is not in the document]
junrushao: Fixed in 36ee686. On Python 3.14+, both the decorated function return annotation and the owner annotations now use annotationlib.get_annotations with Format.FORWARDREF, then the computed field is merged into the complete mapping. I used this instead of directly mutating owner.__annotations__ because direct access can eagerly evaluate unresolved self or forward references during __set_name__. I added a no-future PEP 749 regression preserving a normal field, an unresolved self-referential field, and the init_property return annotation. The Python 3.14 TestInitProperty group passes 9/9. Thanks!

include/tvm/ffi/reflection/registry.h
tqchen: call RegisterTypeAttrValue

include/tvm/ffi/reflection/registry.h
```diff
@@ -1004,16 +1104,32 @@ class TypeAttrDef : public ReflectionDefBase {
...
+   *
+   * \return The TypeAttrDef object.
+   */
+  template <typename TSelf>
+  TVM_FFI_INLINE TypeAttrDef& def_convert() {
+    if constexpr (details::HasContainerType<TSelf>::value) {
+      static_assert(std::is_same_v<typename TSelf::ContainerType, Class>,
+                    "TSelf::ContainerType must match TypeAttrDef's object type");
+    }
+    return def(type_attr::kConvert, &details::FFIConvertFromAnyViewToObjectRef<TSelf>);
```
tqchen, line 1130: this seems to be specialized for convert, ideally we dpn't bloat the api here per ffi key, 

def(type_attr::kConvert, &details::FFIConvertFromAnyViewToObjectRef<TSelf>);

is a fast enough oneliner that i think we should remove this one also removes need of ContainerType

include/tvm/ffi/type_traits.h
```diff
@@ -536,6 +537,88 @@ struct TypeTraits<void*> : public TypeTraitsBase {
...
+    str = str.substr(0, space_pos);
+  }
+  size_t colon_pos = str.find(':');
+  std::string_view name = colon_pos == std::string_view::npos ? str : str.substr(0, colon_pos);
+  if (name.empty()) return std::nullopt;
+  if (str.find(':', colon_pos == std::string_view::npos ? str.size() : colon_pos + 1) !=
+      std::string_view::npos) {
+    return std::nullopt;
+  }
+  std::optional<DLDeviceType> device_type = TryParseDLDeviceType(name);
```
tqchen: now that we do string handling perhaps worthadding ffi/device.h like dtype.h

include/tvm/ffi/reflection/registry.h
junrushao: Thanks for the suggestion. I would like to keep `def_convert`: this is expected to be a common registration for asymmetric conversions, where the reflected object type and the returned public reference type differ. The helper gives that pattern a named API, avoids repeating the `kConvert`/`FFIConvertFromAnyViewToObjectRef` wiring at every call site, and retains the compile-time `ContainerType` consistency check when available. I am therefore keeping the helper rather than expanding each registration into the generic one-liner.

include/tvm/ffi/reflection/registry.h
junrushao: Done in 2979977. `TypeAttrDef::attr` now delegates to `RegisterTypeAttrValue(type_index_, name, std::move(value))`, so type-attribute values share the centralized registration path. Thanks!

include/tvm/ffi/type_traits.h
junrushao: Done in 2979977. I moved the device string parsing and `TypeTraits<DLDevice>` into the new public `tvm/ffi/device.h`, included it from the umbrella/tensor headers and direct consumers, and added focused C++ coverage for native, valid-string, and invalid-string conversions. Thanks!

include/tvm/ffi/device.h
```diff
@@ -0,0 +1,155 @@
...
+    if (ch < '0' || ch > '9') return std::nullopt;
+    value = value * 10 + (ch - '0');
+    if (value > std::numeric_limits<int32_t>::max()) return std::nullopt;
+  }
+  value *= sign;
+  if (value < std::numeric_limits<int32_t>::min() || value > std::numeric_limits<int32_t>::max()) {
+    return std::nullopt;
+  }
+  return static_cast<int32_t>(value);
+}
```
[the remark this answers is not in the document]
junrushao: it's too defensive

python/tvm_ffi/dataclasses/_resolve_fields.py
```diff
@@ -0,0 +1,462 @@
...
+        changed = False
+        remaining: list[_PendingClass] = []
+        for entry in _PENDING_CLASSES:
+            resolved = resolve_type_hints_by_owner(entry.cls, entry.globalns)
+            if resolved is None:
+                remaining.append(entry)
+            else:
+                on_fields_resolved(entry.type_info, resolved)
+                changed = True
+        _PENDING_CLASSES[:] = remaining
```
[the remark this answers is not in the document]
junrushao: it's too defensive

python/tvm_ffi/dataclasses/py_class.py
```diff
@@ -521,24 +402,25 @@ def _register_fields_into_type(
     # ``tvm_ffi.dataclasses.fields()`` compat layer can recover defaults
     # and default_factory values.  _register_fields preserves order, so
     # own_fields and type_info.fields line up 1:1.
+    assert len(own_fields) == len(type_info.fields)
     for py_field, type_field in zip(own_fields, type_info.fields):
         type_field.dataclass_field = py_field
```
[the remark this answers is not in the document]
junrushao: it makes no difference

python/tvm_ffi/dataclasses/_resolve_fields.py
```diff
@@ -0,0 +1,462 @@
...
+    owners = _field_owner_classes(cls)
+    localns.update({owner.__name__: owner for owner in owners})
+    unresolved: list[str] = []
+    for owner in owners:
+        for name, ann_str in own_annotations(owner).items():
+            if isinstance(ann_str, str):
+                try:
+                    eval(ann_str, globalns, localns)  # pylint: disable=eval-used
+                except (NameError, AttributeError) as err:
+                    unresolved.append(f"{name}: {ann_str} ({err})")
```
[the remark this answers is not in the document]
junrushao: Good catch. I updated the diagnostic evaluation to catch `Exception`, so failures from arbitrary annotation expressions are included in the final unresolved-forward-reference `TypeError` while `BaseException` subclasses such as `KeyboardInterrupt` and `SystemExit` still propagate. I also added an end-to-end regression with a missing type followed by `len(1)`; before this change it leaked the inner `TypeError`, and now the final diagnostic includes both fields. The complete `test_dataclass_py_class.py` file passes (387 passed, 1 version-gated skip), and the scoped pre-commit checks pass. Thanks!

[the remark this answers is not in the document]
junrushao: This is a bit too defensive, but updated

python/tvm_ffi/cython/function.pxi
```diff
@@ -398,8 +398,10 @@ cdef int TVMFFIPyArgSetterDType_(
 ) except -1:
     """Setter for dtype"""
     cdef object arg = <object>py_arg
-    # dtype is a subclass of str, so this check occur before str
-    arg = arg._tvm_ffi_dtype
+    # tvm_ffi.dtype is a subclass of str, so this check occurs before str.
+    # The internal DataType helper can also be produced by TypeSchema converters.
+    if not isinstance(arg, DataType):
```
tqchen: more efficient way hre is to allow a separate setter for CDataType, so not paying isinstance cost here

python/tvm_ffi/utils/descriptors.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _get_forward_annotations(obj: Any) -> dict[str, Any]:
+    """Return Python 3.14+ annotations without forcing unresolved names."""
+    annotationlib: Any = importlib.import_module("annotationlib")
+    return annotationlib.get_annotations(obj, format=annotationlib.Format.FORWARDREF)
+
+
+class init_property(Generic[_T]):
+    """Auto-registered C++ field with eager computation at ``__init__`` time.
```
tqchen: this seems should belong to dataclass category as utils is generic thing that do not depend on tvm ffi constructs like filelock

tests/cpp/test_device.cc
```diff
@@ -0,0 +1,60 @@
...
+  DLDevice cpu = cpu_view.cast<DLDevice>();
+  EXPECT_EQ(cpu.device_type, kDLCPU);
+  EXPECT_EQ(cpu.device_id, 0);
+
+  Any cuda_any = String("cuda:3");
+  DLDevice cuda = cuda_any.cast<DLDevice>();
+  EXPECT_EQ(cuda.device_type, kDLCUDA);
+  EXPECT_EQ(cuda.device_id, 3);
+
+  AnyView opencl_view = "opencl:-2";
```
tqchen: is negative value allowed? i think we should reject negative value? also simplifies parser

include/tvm/ffi/device.h
```diff
@@ -0,0 +1,155 @@
...
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace tvm {
+namespace ffi {
+namespace details {
+
+TVM_FFI_INLINE static std::optional<DLDeviceType> TryParseDLDeviceType(std::string_view name) {
+  if (name == "llvm" || name == "cpu" || name == "c" || name == "test") return kDLCPU;
```
tqchen: i know that we had some broad translation on python side which tries to be compact with some compiler target, as part of core i think it does not makes sense likely, let us only keep cpu, cuda, opencl here.

i will followup with downstream changes so they do not directly rely on target name like llvm maps to cpu

python/tvm_ffi/cython/function.pxi
junrushao: Done. I split the cached paths: internal `core.DataType` now uses a dedicated `TVMFFIPyArgSetterCDataType_` that directly copies the `DLDataType` struct, while public `tvm_ffi.dtype` uses the unwrapping setter. The one-time factory selects between them, so the cached hot path no longer pays an `isinstance` cost. I added a direct `core.DataType` FFI round-trip regression; the generated Cython code contains a direct field access with no type check.

tests/cpp/test_device.cc
junrushao: Agreed. Textual device indices are now digits-only, so negative or explicitly signed spellings are rejected and the sign/minimum-range branches are gone. The tests cover negative rejection, acceptance of `INT32_MAX`, and rejection of `INT32_MAX + 1`.

include/tvm/ffi/device.h
junrushao: Agreed. The core String/Any-to-DLDevice fallback now recognizes only the canonical `cpu`, `cuda`, and `opencl` names. I left the existing broader Python `Device` name handling unchanged, since this request is specifically about the core conversion boundary. The C++ test now rejects every removed target/backend alias while keeping native `DLDevice` passthrough unchanged.

python/tvm_ffi/utils/descriptors.py
junrushao: Done. I moved `init_property` into `dataclasses/field.py`, export it as `tvm_ffi.dataclasses.init_property`, and updated `py_class` plus the tests to use the dataclasses API. The generic `utils.descriptors` module and its utils re-export are removed; no compatibility shim is retained because this API is new in this unmerged PR.

tqchen (approved):

PR #670, opened 2026-07-15
feat(abi): generate layout-compatible C++ object declarations

junrushao:
## Summary

- Add `tvm_ffi.dataclasses.gen_abi_cpp` to generate self-contained, lookup-only C++17 object declarations from registered reflection metadata.
- Preserve reflected inheritance and layout across native, Python-defined, recursive, and mutually recursive object types, including `ObjectPtr<TObject>` container fields.
- Expose `tvm::ffi::is_object_subclass_v` as the shared complete-type query and explicit specialization point for generated incomplete object declarations.
- Extend reflection metadata and C++ lookup helpers enough to validate carriers and emit compile-time size, alignment, and field-offset assertions.
- Add frozen presentation support for C++-reflected dataclass fields and keep generator regression fixtures collectable on Python 3.9.

## Architecture

- Expand exact or globbed type-key selectors through their dynamic object ancestry, lower reflected fields to ABI-compatible C++ carriers, and emit deterministically grouped namespaces and declarations.
- Generate lookup-only views that read existing registered objects without registering types, allocating objects, or changing the C ABI.
- Add `TVM_FFI_DECLARE_OBJECT_INFO_LOOKUP` for cached runtime type-index lookup.
- Move `is_object_subclass_v` to public `tvm::ffi`, retain it as the gate for `TypeTraits<ObjectPtr<T>>`, and have generated headers emit direct inline explicit specializations after forward declarations.
- Specialize only genuinely incomplete recursive types in handwritten compile tests; complete object types continue through the primary `std::is_base_of_v` definition.
- Expose field alignment, static type indices, total size, native-metadata availability, and schema subtype relationships through Python reflection so generation can prove layout and inheritance compatibility.
- Match derived-field placement to Microsoft and Itanium-family tail-padding rules while preserving upstream `ObjectPtr`, `WeakObjectPtr`, container, hashing, and subsumption behavior.

## Public Interfaces

- Export `tvm_ffi.dataclasses.gen_abi_cpp`, accepting an exact type key, shell-style pattern, or sequence of selectors and returning C++17 header source.
- Add `TVM_FFI_DECLARE_OBJECT_INFO_LOOKUP` to the public C++ headers.
- Expose `tvm::ffi::is_object_subclass_v` as the public object-subclass query and customization point for intentionally declared incomplete `Object` types.
- Expose `TypeSchema.is_subtype_of`, `TypeField.alignment`, `TypeField.field_static_type_index`, and `TypeInfo.total_size` in Python reflection.
- Add the optional `frozen` parameter to `c_class` for read-only presentation of C++-reflected fields.
- Preserve established `ObjectPtr<TBase>` construction and assignment from `ObjectPtr<TSub>`.

## UI/UX

- None.

## Behavioral Changes

- Generate inheritance-preserving declarations with escaped non-C++ identifiers, opaque dependencies, deterministic namespace ordering, and typed `ObjectPtr` carriers where reflection proves the object type.
- Emit direct inline `is_object_subclass_v` specializations for generated forward declarations instead of a dedicated macro.
- Use erased `ObjectPtr<Object>` carriers for schema-less custom object fields.
- Reject unsafe native aliases, ambiguous `Optional` or `Union` carriers, unsupported static objects, unmatched selectors, and native types without fixed size metadata.
- Place Python-defined subclass fields at ABI-correct tail-padding offsets on Windows and Itanium-family platforms.
- Support reflection-based subtype checks and frozen reflected C++ fields with an explicit descriptor-level mutation escape hatch.
- Preserve backing-object identity across tested object-pointer container covariance and erased-value round trips.
- Spell runtime-resolved generator test unions with `typing.Optional` and `typing.Union` so Python 3.9 can collect the complete generator test file.

## Breaking Changes

- None. The feature is additive and does not alter the C ABI, registered object layout, or existing object-pointer conversion behavior.

## Docs

- Document `c_class(..., frozen=True)`, read-only reflected fields, inheritance behavior, and the controlled descriptor `set` escape hatch in the dataclass reflection guide.
- Document `gen_abi_cpp` selector, output, and non-mutating semantics in its exported function docstring; no separate generator walkthrough is included.
- Document the public `is_object_subclass_v` query and lookup-only metadata macro inline in the C++ headers.

## Tests

- Compiler probe without incomplete-type specialization — failed on `std::is_base_of` as expected.
- Compiler probe with explicit `is_object_subclass_v` specialization — passed.
- C++ build and full CTest suite — build passed; all 451 enabled tests passed and 2 remained disabled.
- Full generator test file under the normal development Python — 14 passed.
- Full generator test file under isolated Python 3.9 — 14 passed.
- Full Python unit suite before the review-driven specialization update — 2,393 passed, 18 skipped, 3 xfailed.
- Pre-commit on all files — passed all hooks.
- Pinned clang-tidy 21.1.1 on `tests/cpp/test_abi_object.cc`, `tests/cpp/test_object_ptr.cc`, and `tests/cpp/test_any.cc` — passed.
- `git diff --check` — passed.
- Rust tests were not run because Cargo and Rustup are unavailable locally.

## Untested Edge Cases

- The complete Python suite was not rerun after the review-driven C++ specialization/output update; both development and Python 3.9 generator suites cover the changed generated text and compilation path.
- Rust workspace tests remain unexecuted locally; no Rust source or binding is modified.
- Linux and Windows compiler matrices remain CI-only. Local generated-header compilation and platform-conditional layout tests cover the paths available on macOS; CI supplies cross-platform ABI confirmation.


include/tvm/ffi/object.h
```diff
@@ -531,12 +528,19 @@ class ObjectPtr {
 
 namespace details {
 
+template <typename T, typename = void>
+struct IsObjectSubclass : std::false_type {};
+
+template <typename T>
+struct IsObjectSubclass<T, std::void_t<decltype(sizeof(T))>>
```
tqchen: seems not needed anymore now that we generate object with subclassing relationship

junrushao: Agreed—the new `IsObjectSubclass` class trait has been removed in `17fe6b2c`, and the existing `is_object_subclass_v = std::is_base_of_v<Object, T>` is restored. The generated-header macro now only specializes that existing variable for forward-declared types. That early specialization is still needed for recursive and mutually recursive `List<ObjectPtr<T>>` fields, because container traits instantiate before the pointee definition is complete; the recursive generator/C++ tests cover this case.

python/tvm_ffi/cython/type_info.pxi
```diff
@@ -928,6 +960,26 @@ _ORIGIN_NATIVE_LAYOUT = {
     "Union": (16, 8, -1),
 }
 
+
+cdef int64_t _get_subclass_offset(object type_info):
+    """Compute where C++ places a byte-aligned field in a direct subclass."""
+    cdef int64_t end
+    if sys.platform == "win32":
+        # The Microsoft C++ ABI does not reuse a base class's tail padding.
+        return type_info.total_size
```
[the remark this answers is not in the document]
junrushao: Fixed in `17fe6b2c`: `_get_subclass_offset(None)` now returns `sizeof(TVMFFIObject)` before either ABI-specific path can dereference `type_info`. The editable extension was rebuilt, and the full Python suite passes.

python/tvm_ffi/dataclasses/gen_abi_cpp.py
```diff
@@ -0,0 +1,666 @@
...
+                5: _Carrier("DLDataType", 4, 2),
+                6: _Carrier("DLDevice", 8, 4),
+                _OBJECT_TYPE_INDEX: _Carrier("::tvm::ffi::ObjectPtr<::tvm::ffi::Object>", 8, 8),
+            }
+            carrier = static_carriers.get(field.field_static_type_index)
+            if carrier is None:
+                raise ValueError(
+                    f"Cannot determine carrier for {owner.type_key}.{field.name}: "
+                    f"missing type schema and static type index {field.field_static_type_index}"
+                )
```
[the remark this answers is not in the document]
junrushao: Fixed in `17fe6b2c`: schema-less static object indices now fall back to `ObjectPtr<Object>` through `_static_carrier`. I also added a focused regression test for a custom object index without a schema.

python/tvm_ffi/dataclasses/gen_abi_cpp.py
```diff
@@ -0,0 +1,666 @@
...
+            "#pragma once",
+            "",
+            "#include <tvm/ffi/tvm_ffi.h>",
+            "",
+        ]
+
+        dependencies = sorted(
+            self.dependencies.values(),
+            key=lambda info: (_cpp_name(info.type_key).qualified, info.type_key),
+        )
```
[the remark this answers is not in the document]
junrushao: Fixed in `17fe6b2c`: dependency ordering now uses `(namespaces, object_name, type_key)` via `_cpp_name_sort_key`. A regression test covers the case where lexicographic qualified-name ordering would split one namespace around a nested namespace.

include/tvm/ffi/container/container_details.h
```diff
@@ -158,9 +158,23 @@ class ReverseIterAdapter {
...
+// A declared ABI object is the declaration-time trust boundary for its recursive
+// ObjectPtr fields. The native path continues to use TypeTraits unchanged.
+template <typename TObject>
+constexpr bool IsObjectPtrStorageEnabled() {
+  if constexpr (is_declared_abi_object_v<std::remove_cv_t<TObject>>) {
+    return std::is_same_v<TObject, std::remove_cv_t<TObject>>;
+  } else {
+    return TypeTraits<ObjectPtr<TObject>>::storage_enabled;
+  }
+}
```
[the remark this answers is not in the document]
junrushao: Thanks. This thread was superseded by the consolidation: `IsObjectPtrStorageEnabled`, `is_declared_abi_object_v`, and the `ObjectPtr`-specific storage dispatch have been removed. Generated incomplete types now specialize the existing `is_object_subclass_v` directly, so this PR no longer introduces a const-specific storage path. The upstream unqualified `TypeTraits<ObjectPtr<T>>` contract remains unchanged.

include/tvm/ffi/object.h
```diff
@@ -546,6 +546,19 @@ inline constexpr bool is_qualified_object_v =
...
+ *
+ * Invoke this macro at global namespace scope after forward declaring Type.
+ * Generated ABI headers use it so recursive container fields can instantiate
+ * ObjectPtr traits before the pointee type's definition is complete.
+ *
+ * \param Type The fully qualified Object subclass.
+ */
+#define TVM_FFI_DECLARE_OBJECT_SUBCLASS(Type) \
+  template <>                                 \
+  inline constexpr bool tvm::ffi::details::is_object_subclass_v<Type> = true
```
tqchen: minor, given this is a general mechnanism like type_subsume_v we can keep it under tvm::ffi::

would be good to gave a testcase to confirm it helps recursive decl of ObjectPtr<T>.

This macro is one liner and maybe worthwhile to just do the two liner as it is in places that need it

tqchen: minor, leave this in `tvm::ffi::is_object_subclass_v<Type>`

junrushao: Addressed in `02d4f312`: the trait is now `tvm::ffi::is_object_subclass_v<Type>` rather than an implementation detail, and the `ObjectPtr` traits plus generated explicit specializations all use that public name.

junrushao: Addressed in `02d4f312`. `is_object_subclass_v` now lives in `tvm::ffi`, and `TypeTraits<ObjectPtr<TObject>>` uses that public trait directly. I removed `TVM_FFI_DECLARE_OBJECT_SUBCLASS`; generated headers now emit the two-line inline specialization immediately after each forward declaration.

I also made the recursive requirement explicit in `test_abi_object.cc`: only the still-incomplete `RecursiveObj` is specialized, its `ObjectPtr` storage trait is asserted before the class definition, and the class then declares `List<ObjectPtr<RecursiveObj>>`. The mutual-recursion coverage remains in `test_object_ptr.cc`. A compiler probe confirms that omitting the specialization fails in `std::is_base_of_v<Object, NodeObj>` while the direct specialization compiles. Full CTest (451 enabled), generator tests on Python 3.9 and the dev interpreter, pre-commit, and targeted clang-tidy all pass.

junrushao: A minimal example of the ordering this enables is:

```cpp
namespace example {
struct NodeObj;
}

template <>
inline constexpr bool
    tvm::ffi::is_object_subclass_v<::example::NodeObj> = true;

namespace example {
struct NodeObj : public tvm::ffi::Object {
  tvm::ffi::List<tvm::ffi::ObjectPtr<NodeObj>> children;
};
}
```

When `children` is declared, `NodeObj` remains incomplete until the closing brace. `List` instantiates `TypeTraits<ObjectPtr<NodeObj>>`, so without the specialization the primary `std::is_base_of_v<Object, NodeObj>` is ill-formed. Generated headers now follow exactly this forward-declare → specialize → define ordering.

tqchen (approved):

PR #678, opened 2026-07-20
Improve typing of overloaded methods in stubgen

mbrookhart:


junrushao (approved):
Thanks for improving the generator!

PR #684, opened 2026-07-23
[FEAT][RUST] Support runtime-checked try_cast between ObjectRef types

Seven-Streams:
This PR adds direct, runtime-checked casting between Rust object handles, mirroring C++ `ObjectRef::as<T>` / `Downcast<T>`.

**ObjectRefCast** is blanket-implemented for types implementing both `ObjectRefCore` and `AnyCompatible`. `try_cast::<B>(self)` consumes the source and, on success, rewraps the same object as `B` without cloning or copying it. A failed cast returns a `TypeError` describing the source and target types.

Target compatibility is delegated to `B::check_any_strict`. ObjectRef hierarchies accept `B` and its runtime subtypes, while parameterized containers use their complete type semantics.

```rust
let base: TestBase = derived.try_cast()?; // upcast
let obj: ObjectRef = base.try_cast()?;    // upcast to root
let d: TestDerived = obj.try_cast()?;     // checked downcast
```

The direct `A -> B` path reuses the existing `AnyCompatible` conversion traits through a raw `TVMFFIAny`, so callers do not need to construct an `Any` or `AnyView`. It first creates a non-owning view for the target check and transfers ownership only after that check succeeds, keeping failure and panic paths leak-free.

Derived ObjectRef conversions preserve the runtime type index stored in the object header. A derived object upcast to a base handle can therefore still round-trip through `Any` or `AnyView` and downcast correctly.

A hidden `is_instance_of` helper provides hierarchy checks for derive-generated object containers. The cast path uses `assert_unchecked` after establishing the object type-index invariant; release codegen confirms that the intermediate `TVMFFIAny` storage and redundant object-range check are eliminated.

The PR also includes the derive fixes required by downstream crates and focused tests covering hierarchy casts, failure ownership, dynamic-type preservation through `Any`/`AnyView`, direct `try_cast_from_any_view`, and parameterized-container mismatch.

### DIFF
--- a/rust/tvm-ffi-macros/src/object_macros.rs
+++ b/rust/tvm-ffi-macros/src/object_macros.rs
@@ -58,7 +58,7 @@ pub fn derive_object(input: proc_macro::TokenStream) -> TokenStream {
                                 &type_key_arg, &mut tindex
                             );
                             if ret != 0 {
-                                proc_macro_error::abort!("Failed to get type index for type key: {}", #type_key);
+                                panic!("Failed to get type index for type key: {}", #type_key);
                             }
                             tindex
                         }
@@ -143,7 +143,7 @@ pub fn derive_object_ref(input: proc_macro::TokenStream) -> TokenStream {
 
         // implement AnyCompatible for #struct_name
         unsafe impl #tvm_ffi_crate::type_traits::AnyCompatible for #struct_name {
-            fn type_str() -> String {
+            fn type_str() -> std::string::String {
                 type ContainerType = <#struct_name as #tvm_ffi_crate::object::ObjectRefCore>
                     ::ContainerType;
                 <ContainerType as #tvm_ffi_crate::object::ObjectCore>::TYPE_KEY.into()
@@ -171,7 +171,7 @@ pub fn derive_object_ref(input: proc_macro::TokenStream) -> TokenStream {
                     ::ContainerType;
                 let type_index =
                     <ContainerType as #tvm_ffi_crate::object::ObjectCore>::type_index();
-                data.type_index == type_index as i32
+                #tvm_ffi_crate::object::is_instance_of(data.type_index, type_index)
             }
 
             unsafe fn copy_from_any_view_after_check(
@@ -227,7 +227,7 @@ pub fn derive_object_ref(input: proc_macro::TokenStream) -> TokenStream {
                     ::ContainerType;
                 let type_index =
                     <ContainerType as #tvm_ffi_crate::object::ObjectCore>::type_index();
-                if data.type_index == type_index as i32 {
+                if #tvm_ffi_crate::object::is_instance_of(data.type_index, type_index) {
                     Ok(Self::copy_from_any_view_after_check(data))
                 } else {
                     Err(())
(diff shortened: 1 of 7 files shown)


rust/tvm-ffi-macros/src/object_macros.rs
```diff
@@ -192,23 +195,24 @@ pub fn derive_object_ref(input: proc_macro::TokenStream) -> TokenStream {
...
-                    <ContainerType as #tvm_ffi_crate::object::ObjectCore>::type_index();
-                data.type_index = type_index as i32;
-                data.small_str_len = 0;
                 let data_ptr = #tvm_ffi_crate::object::ObjectArc::into_raw(
                     src.data
                 );
-                data.data_union.v_obj =
-                    data_ptr as *mut ContainerType as *mut  #tvm_ffi_crate::tvm_ffi_sys::TVMFFIObject;
+                let object_ptr =
+                    data_ptr as *mut ContainerType as *mut #tvm_ffi_crate::tvm_ffi_sys::TVMFFIObject;
```
Seven-Streams, line 210: Why the order is changed

rust/tvm-ffi/src/object.rs
```diff
@@ -95,13 +99,120 @@ pub unsafe trait ObjectCoreWithExtraItems: ObjectCore {
...
+/// table.
+#[doc(hidden)]
+#[inline(always)]
+pub fn is_instance_of(object_type_index: i32, target_type_index: i32) -> bool {
+    if object_type_index == target_type_index {
+        return true;
+    }
+    let object_begin = TypeIndex::kTVMFFIStaticObjectBegin as i32;
+    // Every registered object is an instance of the root Object type.
+    if target_type_index == object_begin {
```
Seven-Streams: not sure if it's necessary

rust/tvm-ffi-macros/src/object_macros.rs
tlopex: We need the object pointer first because the type index is now read from the runtime object header instead of ContainerType::type_index(), which preserves the dynamic subtype after an upcast.

rust/tvm-ffi/src/object.rs
tlopex: I think it is. this is the root object fast path. Every object-range type must cast to objectref, and the C++ uses the same range check while avoiding type-table lookups.

tqchen: given this is runtime check, likely not having the specialization is ok, depending on how freq conversion to ObjectRef is.

rust/tvm-ffi/src/object.rs
```diff
@@ -95,13 +99,120 @@ pub unsafe trait ObjectCoreWithExtraItems: ObjectCore {
...
+    if object_type_index < target_type_index {
+        return false;
+    }
+    unsafe {
+        let object_info = TVMFFIGetTypeInfo(object_type_index);
+        let target_info = TVMFFIGetTypeInfo(target_type_index);
+        if object_info.is_null() || target_info.is_null() {
+            return false;
+        }
+        let target_depth = (*target_info).type_depth;
```
tqchen: might worthcheck other template trick for target_type_index. Note in the case of C++ TargetType can be a type T, where things like type_depth can be directly queried as a compile time constant rather than runtime

rust/tvm-ffi/src/object.rs
tlopex: agreed

tlopex (approved):

PR #685, opened 2026-07-24
[Feat][Rust] Add leaf lookup dispatch for match_any

tlopex:
This PR adds a leaf-only lookup fast path to Rust match_any!. For at least two unguarded arms with simple bindings, when every pattern corresponds to one stable runtime TypeIndex, it lazily maps each TypeIndex to a source-order ArmId and uses a native Rust match to select only the relevant conversion. The table stores no closures or arm bodies, duplicate types preserve first-arm semantics, and misses use the _ fallback. Single-arm, guarded, non-leaf/category, or content-dependent matches retain the Phase 1 ordered path. The PR also adds opt-in final-type metadata and focused integration tests. The two-arm threshold is based on release benchmarks of standard ObjectRef conversions.


rust/tvm-ffi/src/extra/module.rs
Seven-Streams, line 1: Why we need to add modules here instead of in tests

rust/tvm-ffi-macros/src/object_macros.rs
```diff
@@ -31,6 +31,11 @@ pub fn derive_object(input: proc_macro::TokenStream) -> TokenStream {
     let type_key = get_attr(&derive_input, "type_key")
         .map(attr_to_str)
         .expect("Expect #[type_key = \"<my_type_key>\"] attribute");
+    let type_final = match get_attr(&derive_input, "type_final") {
```
Seven-Streams, line 34: I think `type_final` can be `false` in cpp side?

rust/tvm-ffi/src/match_any_internal.rs
Seven-Streams, line 1: Tests should not be here. why we have another file for `match_any`

rust/tvm-ffi/src/object.rs
```diff
@@ -50,6 +50,11 @@ unsafe impl<T: Send + Sync + ObjectCore> Sync for ObjectArc<T> {}
 pub unsafe trait ObjectCore: Sized + 'static {
     /// the type key of the object
     const TYPE_KEY: &'static str;
+    /// Whether every instance of this type has exactly `Self::type_index()`.
+    ///
+    /// A final type must never have a registered subtype.
+    #[doc(hidden)]
```
Seven-Streams, line 66: I'm not sure if it's necessary. I think `ObjectCore` is not designed to be exposed to users.

rust/tvm-ffi/src/extra/module.rs
tlopex: okay, would move it

rust/tvm-ffi/src/match_any_internal.rs
tlopex: moved

rust/tvm-ffi-macros/src/object_macros.rs
tlopex: Yes. false is the default, so a type whose C++ definition has _type_final = false simply omits the attribute.

rust/tvm-ffi/src/object.rs
tlopex: `ObjectCore` is already public because downstream `#[derive(Object)]`expansions need it; TYPE_FINAL is default-false and #[doc(hidden)], so it remains internal derive metadata rather than a user-facing API.

rust/tvm-ffi/src/match_any_internal.rs
Seven-Streams: maybe rename to `match_any_internal`?

rust/tvm-ffi-macros/src/match_any.rs
```diff
@@ -24,6 +24,11 @@ use syn::{braced, parenthesized, Expr, Pat, Path, Result, Token};
 
 use crate::utils::get_tvm_ffi_crate;
 
+// Keep single-arm matches ordered. Release benchmarks of the standard ObjectRef
+// conversion path show that lookup pays off from two arms when a later arm or
+// the fallback is reached.
+const MIN_LOOKUP_TABLE_ARMS: usize = 2;
```
Seven-Streams, line 30: only 2?

rust/tvm-ffi-macros/src/match_any.rs
```diff
@@ -105,41 +110,225 @@ pub fn expand(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+                #rejected => {
+                    ::core::mem::drop(#rejected);
+                    #next
+                }
+            }
+        }
+    })
+}
+
+fn expand_leaf_lookup_match(
```
Seven-Streams: breaking it into several parts may be better for understanding.

rust/tvm-ffi-macros/src/match_any.rs
```diff
@@ -105,41 +110,225 @@ pub fn expand(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+
+    let lookup_arm_id = quote! {
+        {
+            use #tvm_ffi::match_any::LeafPatternMetadata as _;
+
+            let #probe =
+                #tvm_ffi::match_any::LeafPatternProbe::<#pattern_list>::new();
+            match (&#probe).leaf_pattern_list_id() {
+                ::core::option::Option::Some(#pattern_list_id) => {
+                    static #static_table: ::std::sync::OnceLock<
```
Seven-Streams, line 240: one problem is that when the arm is something like `T()`, where `T` is related to the external function, then the table may never be used

rust/tvm-ffi-macros/src/match_any.rs
```diff
@@ -105,41 +110,225 @@ pub fn expand(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+                    #table.lookup(#pattern_list_id, #view.type_index())
+                }
+                ::core::option::Option::None => {
+                    ::core::result::Result::Err(())
+                }
+            }
+        }
+    };
+
+    let ordered_selection = arms.iter().enumerate().rev().fold(
```
Seven-Streams: can we reuse the previous logic

rust/tvm-ffi-macros/src/match_any.rs
tlopex: I have a test that shows 2 is best.

rust/tvm-ffi/src/match_any_internal.rs
tlopex: agreed

rust/tvm-ffi-macros/src/match_any.rs
tlopex: yes, would change it

rust/tvm-ffi-macros/src/match_any.rs
tlopex: updated

rust/tvm-ffi-macros/src/match_any.rs
tlopex: the case is not frequent so we just keep the current implenment

Seven-Streams (approved):

tqchen (approved):

PR #696, opened 2026-08-01
[CORE] Staging Object Tying Feature

tqchen:
## Summary

Object tying remains compiled but is deliberately staged dormant: canonical-wrapper detection returns false, while C++, Python-defined dataclasses, and Rust objects continue to use matching legacy allocation and deletion paths.

This staging has three goals:

- let downstream projects upgrade without making existing generated caches depend on the new allocator contract;
- keep the compatibility change compact by changing only the activation/allocation boundary instead of removing the object-tying implementation;
- preserve the custom allocator API and a clean path for a later coordinated object-tying activation.

Legacy wrapper identity and reference-count expectations are restored, while activation-only tying tests remain as skipped coverage for the later rollout.

## Testing

- C++: 465 enabled tests passed (2 disabled).
- Python: 2,381 passed, 78 skipped, 2 expected failures.
- Rust: 138 unit/integration tests and 3 doctests passed; formatting is clean.
- Pre-commit: all changed-file checks passed, including ruff, ty, and clang-format.
- Order-balanced microbenchmarks place the dormant path about 1.3-7.9% behind active tying and 4.7-12.5% ahead of a full revert.

### DIFF
--- a/include/tvm/ffi/memory.h
+++ b/include/tvm/ffi/memory.h
@@ -148,12 +148,7 @@ class ObjAllocatorBase {
 // Simple allocator that uses new/delete.
 class SimpleObjAllocator : public ObjAllocatorBase<SimpleObjAllocator> {
  private:
-  /*! \brief Guard a custom-allocator allocation until ownership is transferred to an object.
-   *         If placement construction throws, free the raw block through its allocator's
-   *         ``delete_space`` (the same path the live deleter uses). ``data`` is a body pointer
-   *         offset past the ``TVMFFIObjectAllocHeader``, so a plain ``AlignedFree`` would be wrong;
-   *         the header (with ``delete_space`` wired) is set up by ``allocate`` before construction,
-   *         so this is valid even though the body was never constructed. */
+  /*! \brief Guard a simple-allocator allocation until ownership is transferred to an object. */
   class AllocGuard {
    public:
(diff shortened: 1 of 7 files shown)


tqchen:
Post-publication FlashInfer confirmation against published head `89031a1c`:

- Rebuilt the real generated `silu_and_mul` CUDA source against the published headers for H200 (`sm_90a`). The resulting DSO has SHA-256 `8c9e044ce5d239d946274918bd589e9fbdc6dbdb3b433b448facf2c329db0b52`.
- The DSO has no undefined `TVMFFIGetCustomAllocator` or `TVMFFISetCustomAllocator` reference.
- With `LD_BIND_NOW=1`, isolated released tvm-ffi 0.1.11 and 0.1.12 runtimes both load the DSO and produce the expected argument-count `TypeError`; this exercises cache-side error-object construction without a symbol-resolution failure.
- Both released runtimes also complete a valid FP16 `silu_and_mul` launch on H200, synchronize successfully, and match the Torch reference at `rtol=1e-3`, `atol=1e-3`.
- In the reverse direction, the pre-allocator-change owned-Shape cache still loads, returns `[2, 3]`, and destroys cleanly with the dormant runtime.

The detailed compatibility investigation and benchmark matrix remain on the task issue.

cyx-6 (approved):

PR #703, opened 2026-08-06
fix: create wheel for GIL build of Python 3.14

tmct:
Hi, please could we get the non-free-threading Python 3.14 built for this library as well?


tqchen (approved):

tqchen:
cc @cyx-6

tmct:
Thank you

PR #704, opened 2026-08-07
[Feat][Rust] Add native structural_map and structural_mutate

tlopex:
This PR adds Rust-native `structural_map` and `structural_mutate` support as the Rust counterpart to the C++ structural mutation APIs introduced in #649. It builds on the Rust `structural_visit` and `structural_walk` infrastructure from #693.

## Design

Rust owns:

- Callback dispatch and walk order.
- Definition-region propagation.
- DAG and FreeVar identity remapping.
- Error context and panic propagation.
- The user-facing structural-map and structural-mutator APIs.

Type-specific traversal remains with the implementation that registered the runtime type.

For default recursion, Rust looks up the existing structural mutation type attributes:

- `__s_mutate__`
- `__s_maybe_inplace_mutate__`

Rust constructs an ABI-compatible `ffi.StructuralMutator` object. Registered hooks receive this active mutator and recursively call its language-independent vtable to return child values to Rust.

This allows the existing C++ hooks for Array, List, Map, and Dict to manage their private storage while Rust continues to control callback dispatch and recursive transformation.

Hooks may be registered as either opaque function pointers or `ffi.Function` objects.

This PR does not add a Map/Dict-specific C API, does not add `TVMFFIMapMutateValues`, and does not mirror the private C++ Map/Dict storage layout.

## Rust APIs

This PR provides:

- `structural_map` with pre-order and post-order callbacks.
- A low-level `StructuralMutator` trait for custom recursion.
- `InplaceValue` for engine-authorized in-place mutation.
- `StructuralVarRemap` for owning DAG and FreeVar substitutions.
- Typed closures and ordered callback tuples of up to eight handlers.
- `#[dispatch(map)]` generated dispatch.
- Optional `DefRegionKind` callback arguments.
- Fallible callbacks returning `Result<Any>`.
- Shared callback value handling between structural visit and map.
- Structural error context for objects and reflected fields.

Callback tuples use ordered first-match dispatch. A `&MapValue` handler acts as the catch-all handler and should therefore appear last.

## Mutation semantics

The root is consumed so Rust ownership and the runtime reference count determine whether in-place mutation is allowed.

- Uniquely owned Array, List, Map, and Dict objects may reuse their storage.
- Passing a shared root, such as `root.clone()`, selects copy-on-write behavior.
- Map and Dict keys are structural anchors and are not mapped.
- If no value changes, the original object may be returned.
- Reflected objects use `__ffi_shallow_copy__` together with registered field getters and setters.
- FreeVar and DAG-node identities are mapped once and reuse the completed mapped result.
- `None` and other inline values are passed through normal callback dispatch.
- Changes completed through an in-place path are not rolled back if a later callback fails.

Rust panics are caught before crossing the C ABI. The panic is resumed only after the registered runtime hook returns, allowing C++ RAII cleanup to finish first.

A retained `ffi.StructuralMutator` reference also fails safely after the active structural-mutation call ends instead of keeping a dangling Rust callback context.

## Shared structural callback support

The structural visit and map implementations share:

- Borrowed callback-value handling.
- Typed object-node casts.
- Callback tuple arity generation.

The procedural macro implementation is generalized from visit-only dispatch to support both `#[dispatch(visit)]` and `#[dispatch(map)]`.


rust/tvm-ffi-macros/src/dispatch.rs
Seven-Streams, line 1: maybe we need to rename it to `dispatch.rs` now

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1609 @@
...
+const SHALLOW_COPY_ATTR: &str = "__ffi_shallow_copy__";
+const FLAG_SEQ_HASH_IGNORE: i64 = kTVMFFIFieldFlagBitMaskSEqHashIgnore as i64;
+const FLAG_SETTER_IS_FUNCTION: i64 = kTVMFFIFieldFlagBitSetterIsFunctionObj as i64;
+
+/// Borrowed value passed to structural-map callbacks.
+///
+/// This is the same representation used by structural-visit callbacks; the
+/// alias keeps typed casts and borrowed node checks on one audited unsafe
+/// implementation.
+pub type MapValue = VisitValue;
```
Seven-Streams: it looks weird. maybe we need to design a common.rs to store them

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1609 @@
...
+                MapChain {
+                    links: self,
+                    markers: PhantomData,
+                }
+            }
+        }
+    };
+}
+
+impl_map_chain!((F0, M0, 0));
```
Seven-Streams: what's this

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1609 @@
...
+        match raw.type_index {
+            x if x == TVMFFITypeIndex::kTVMFFIArray as i32
+                || x == TVMFFITypeIndex::kTVMFFIList as i32 =>
+            {
+                return self.map_sequence(raw, def_region_kind, permit);
+            }
+            x if x == TVMFFITypeIndex::kTVMFFIMap as i32
+                || x == TVMFFITypeIndex::kTVMFFIDict as i32 =>
+            {
+                return self.map_values(raw, def_region_kind, permit);
```
Seven-Streams: can we have a better name for `map_values`

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1609 @@
...
+    }
+}
+
+#[derive(Clone, Copy)]
+struct CheckedSequence {
+    data: *mut TVMFFIAny,
+    size: usize,
+}
+
+fn checked_sequence(raw: TVMFFIAny) -> Result<CheckedSequence> {
```
Seven-Streams: I'm wondering if we need to validate it.

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1609 @@
...
+
+fn shallow_copy_column() -> Option<TypeAttrColumn> {
+    CachedColumn {
+        cache: &SHALLOW_COPY_COLUMN,
+        name: SHALLOW_COPY_ATTR,
+    }
+    .get()
+}
+
+#[cfg(test)]
```
Seven-Streams: maybe we should not put tests here.

rust/tvm-ffi/src/extra/structural_mutate.rs
tlopex: Rust requires separate implementations for tuples containing 1 to 8 callbacks. Already added comments

rust/tvm-ffi/src/extra/structural_mutate.rs
tlopex: would simplify them

rust/tvm-ffi/src/extra/structural_mutate.rs
tlopex: okay

rust/tvm-ffi-macros/src/dispatch.rs
tlopex: addressed

rust/tvm-ffi/src/extra/structural_mutate.rs
Seven-Streams, line 1: move the tests to test files.

rust/tvm-ffi/src/extra/structural_visit.rs
Seven-Streams, line 1: ditto

rust/tvm-ffi/src/extra/structural_mutate.rs
tlopex: addressed

rust/tvm-ffi/src/extra/structural_visit.rs
tlopex: addressed

Seven-Streams (approved):

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1811 @@
...
+};
+use crate::tvm_ffi_sys::{TVMFFIObjectHandle, TVMFFISEqHashKind};
+
+use super::structural_common::impl_callback_chain_tuple_arities;
+use super::structural_visit::{
+    field_def_region, for_each_field, free_var_child_region, type_attr_column, type_key_of,
+    DefRegionKind, TypeAttrColumn, WalkOrder,
+};
+
+const STRUCTURAL_MUTATE_ATTR: &str = "__s_mutate__";
```
Seven-Streams, line 69: maybe we need to algin `__s_visit__` with it,

Seven-Streams, line 69: maybe we need to add an api to add hooks from rust side.

rust/tvm-ffi/src/extra/structural_mutate.rs
```diff
@@ -0,0 +1,1811 @@
...
+
+fn with_value_context(error: Error, raw: TVMFFIAny) -> Error {
+    if raw.type_index < TVMFFITypeIndex::kTVMFFIStaticObjectBegin as i32 {
+        error
+    } else {
+        with_error_context(error, &format!("object `{}`", type_key_of(raw.type_index)))
+    }
+}
+
+fn with_error_context(error: Error, frame: &str) -> Error {
```
Seven-Streams, line 1804: maybe we can unify the error handling.

Seven-Streams (approved):

tqchen (approved):

PR #706, opened 2026-08-11
[FIX] Support free-threaded Python cross-builds

### LINKED ISSUE #705: Cross-compiling against free-threaded Python 3.14 (3.14t) fails to find Development.Module
## Description
I'm one of the maintainers of the conda-forge package for apache-tvm-ffi and ran into this. Cross-compiling `apache-tvm-ffi` 0.1.13.post3 against free-threaded CPython 3.14 fails during CMake configuration, even though the target development headers are installed.

This is reproducible in [conda-forge/apache-tvm-ffi-feedstock#20](https://github.com/conda-forge/apache-tvm-ffi-feedstock/pull/20) for:

- `linux-64` → `linux-ppc64le`
- `osx-64` → `osx-arm64`

The corresponding cross-compiled GIL-enabled Python 3.14 builds pass, as do native Python 3.14t builds.

## Error

Both failing jobs report:

```text
-- Found Python: ... (found version "3.14.6") found components: Interpreter
-- Free-threaded Python detected.
CMake Error ...
  Could NOT find Python (missing: Development.Module)
  (found suitable exact version "3.14.6")
...
  CMakeLists.txt:262 (find_package)
```

CI logs:

- [Linux PPC64LE Python 3.14t](https://github.com/conda-forge/apache-tvm-ffi-feedstock/actions/runs/31447767536/job/93645550751)
- [macOS ARM64 Python 3.14t](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_build/results?buildId=1565871&view=logs&jobId=0bffcc46-b591-524d-b470-fe7cc095b4e2)

The target packages contain `include/python3.14t/Python.h`; their `pyconfig.h` defines `Py_GIL_DISABLED 1`, and their target sysconfig data has `ABIFLAGS='t'`.

## Analysis

The current CMake logic first finds the host interpreter and detects that it is free-threaded, then performs a second lookup requesting both `Interpreter` and `Development.Module`:

https://github.com/apache/tvm-ffi/blob/f59eae05349dd77d5de07a1883fe88fabe6bab65/CMakeLists.txt#L241-L266

Two things are needed for the cross-build:

1. `Python_FIND_ABI` must explicitly select the free-threaded ABI. On POSIX, CMake defaults it to `ANY;ANY;ANY;OFF`, which excludes the `t` ABI.
2. The second cross-compilation lookup should request only the target `Development.Module`, rather than combining it with the host `Interpreter`.

Using CMake 4.4.2 with the exact PPC64LE and macOS ARM64 target packages:

- The current combined lookup fails.
- Setting `Python_FIND_ABI` while retaining the combined lookup still fails.
- Setting `Python_FIND_ABI` and making the second lookup `Development.Module`-only succeeds.

This also follows CMake’s documented model of separate host-interpreter and target-development searches:

https://cmake.org/cmake/help/v4.4/module/FindPython.html

## Proposed fix

For free-threaded Python, set:

```cmake
set(Python_FIND_ABI "OFF;ANY;ANY;ON")
```

When `CMAKE_CROSSCOMPILING` is true, retain the first Interpreter-only lookup for version detection and later Cython use, but make the second lookup:

```cmake
find_package(
  Python ${_tvm_ffi_python_version} EXACT
  COMPONENTS Development.Module
  REQUIRED
)
```

Native builds can retain the existing combined `Interpreter Development.Module` lookup.

This seems easy enough to fix so throwing up a PR and linking it to this issue.

iamthebot:
## Summary

The conda-forge Python 3.14 free-threading migration is blocked for TVM FFI on platforms that rely on cross-compilation. Linux PPC64LE and macOS ARM64 builds fail during CMake's Python development lookup even though the target 3.14t headers are installed, while native 3.14t builds and GIL-enabled Python 3.14 cross-builds succeed.

This PR makes Python discovery distinguish the host interpreter used to run Cython from the target development artifacts used to build the extension. That lets downstream packagers build Python 3.14t wheels for those architectures without changing native builds or GIL-enabled cross-build behavior.

## Background

Cross-compiling against Python 3.14t currently finds the host interpreter and detects
`Py_GIL_DISABLED`, but the second `FindPython` call combines that host interpreter with target
`Development.Module` artifacts. CMake 4.4 also enforces the `gil_disabled` ABI default of `OFF`,
so it rejects the installed `python3.14t` target headers and reports
`missing: Development.Module`.

The failing configurations are visible in
[conda-forge/apache-tvm-ffi-feedstock#20](https://github.com/conda-forge/apache-tvm-ffi-feedstock/pull/20):
Linux x86_64 to PPC64LE and macOS x86_64 to ARM64. Native Python 3.14t and cross-built GIL-enabled
Python 3.14 already pass.

This change keeps the first interpreter for free-thread detection and Cython, performs a
development-only target lookup when cross-compiling, and uses `SKBUILD_SOABI` for the target
extension suffix when available. Native discovery remains unchanged.

Fixes #705.

## Testing

- configured with CMake 4.4.2 against the exact PPC64LE Python 3.14t target headers; generated
  `core.cpython-314t-powerpc64le-linux-gnu.so`
- configured with CMake 4.4.2 against the exact macOS ARM64 Python 3.14t target headers; generated
  `core.cpython-314t-darwin.so`
- verified the cross path on CMake 3.26.4 and 3.30.5 with developer warnings treated as errors
- built and imported an editable native Python 3.14t package; verified the extension retained its
  `cpython-314t` suffix
- `pytest -n0 -vvs tests/python/test_free_threaded_python_helpers.py tests/python/test_build.py::test_build_cpp`
- `cmake-format --check CMakeLists.txt`
- `cmake-lint CMakeLists.txt`
- ASF header, file-type, and whitespace checks

## AI Disclosure

Was AI used in the creation of this PR?

- [ ] No. This was fully written by a human.
- [ ] Limited (uses other than code generation, such as ideation or debugging, or very limited code generation)
- [ ] Moderate (code generated by an LLM with continuous human review and intervention)
- [x] High (key decisions made by a human with active steering and feedback; final code reviewed by the human)
- [ ] Vibe Coded (generated largely autonomously from an initial prompt with limited human steering or review)

The human identified the failing downstream migration, set the upstream issue and PR scope, and
made the key workflow decisions. OpenAI Codex (GPT-5), including delegated Codex agents,
investigated the CMake behavior, implemented the change, ran the native and cross-build validation
matrix, reviewed the final diff, and prepared this PR. The human reviewed and approved the final
diff; agent review did not replace that review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,12 +243,10 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
     COMPONENTS Interpreter
     REQUIRED
   )
-  set(_tvm_ffi_python_executable "${Python_EXECUTABLE}")
-  set(_tvm_ffi_python_version ${Python_VERSION})
 
   # Run a Python script to check for free-threaded build
   execute_process(
-    COMMAND "${_tvm_ffi_python_executable}" -c
+    COMMAND "${Python_EXECUTABLE}" -c
             "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') == 1)"
     OUTPUT_VARIABLE PYTHON_IS_FREE_THREADED
     OUTPUT_STRIP_TRAILING_WHITESPACE
@@ -274,7 +272,7 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
     set(_tvm_ffi_python_components Interpreter Development.Module)
   endif ()
   find_package(
-    Python ${_tvm_ffi_python_version} EXACT
+    Python ${Python_VERSION} EXACT
     COMPONENTS ${_tvm_ffi_python_components}
     REQUIRED
   )
@@ -296,8 +294,8 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
   )
   add_custom_command(
     OUTPUT ${_core_cpp}
-    COMMAND "${_tvm_ffi_python_executable}" -m cython --cplus ${_core_pyx} -o ${_core_cpp}
-            --module-name "tvm_ffi.core"
+    COMMAND "${Python_EXECUTABLE}" -m cython --cplus ${_core_pyx} -o ${_core_cpp} --module-name
+            "tvm_ffi.core"
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     COMMENT "Transpiling ${_core_pyx} to ${_core_cpp}"
     DEPENDS ${_cython_sources}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,25 +243,39 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
     COMPONENTS Interpreter
     REQUIRED
   )
+  set(_tvm_ffi_python_executable "${Python_EXECUTABLE}")
+  set(_tvm_ffi_python_version ${Python_VERSION})
 
   # Run a Python script to check for free-threaded build
   execute_process(
-    COMMAND ${Python_EXECUTABLE} -c
+    COMMAND "${_tvm_ffi_python_executable}" -c
             "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') == 1)"
     OUTPUT_VARIABLE PYTHON_IS_FREE_THREADED
     OUTPUT_STRIP_TRAILING_WHITESPACE
   )
   if (PYTHON_IS_FREE_THREADED)
     message(STATUS "Free-threaded Python detected.")
+    if (CMAKE_CROSSCOMPILING
+        AND CMAKE_VERSION VERSION_GREATER_EQUAL "4.4"
+        AND NOT DEFINED Python_FIND_ABI
+    )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -243,25 +243,39 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
     COMPONENTS Interpreter
     REQUIRED
   )
+  set(_tvm_ffi_python_executable "${Python_EXECUTABLE}")
```
tqchen: seems this is python executable itself, if so avoid setting new item

CMakeLists.txt
```diff
@@ -243,25 +243,39 @@ if (TVM_FFI_BUILD_PYTHON_MODULE)
...
+  set(_tvm_ffi_python_version ${Python_VERSION})
```
tqchen: seems this is just python version, if so avoid aliasing

tqchen (commented):
thanks for contributing, some quick remarks

tqchen:
cc @cyx-6

tqchen (approved):

PR #709, opened 2026-08-21
[FIX] Make ffi.Error picklable

cyx-6:
## Motivation

`Error.py_error()` attaches the FFI error object to the Python exception it raises:

```python
py_error.__tvm_ffi_error__ = self
```

`BaseException.__reduce__` includes the instance `__dict__`, so pickling an FFI-originated exception pickles that `Error` through `CObject.__getstate__` → `ToJSONGraph`. `ffi.Error` has no native creator and no `__ffi_new__` type attr, so this fails:

```
TypeError: Type `ffi.Error` does not support ToJSONGraph
           (no native creator or __ffi_new__ type attr)
```

Harnesses that ship exceptions across a process boundary lose the original error and report this instead. It was reported downstream against a CUDA IMA (error 700), where the replacement message made it look like tvm-ffi had suppressed a coredump — it had not; the coredump happens before the error is raised, and the only real defect is that the exception could not be pickled.

`copy.deepcopy` hits the same path.

## Changes

`ffi.Error` now pickles by value. An error is fully described by its `(kind, message, backtrace)` strings, so `Error.__reduce__` reconstructs it through `__init__` rather than the JSON graph. This mirrors `Device` and `DataType`, which already carry hand-written `__reduce__` implementations for the same reason.

`extra_context` is intentionally dropped — it may hold arbitrary native payloads, and preserving it only when it happens to be serializable would make pickling an error depend on where the error came from.

The attribute stays a live `Error` rather than degrading to `None`, because `set_last_ffi_error` calls `update_backtrace` on `__tvm_ffi_error__` when a restored exception propagates back through the FFI; a `None` there would fail on an unrelated `hasattr` check that is already in the code.

## Not in scope

An earlier revision of this PR also rewrote the message raised when pickling types that genuinely cannot round-trip (`ffi.Function`, `ffi.Module`, `ffi.Tensor`, `ffi.OpaquePyObject`). Per review, that is dropped here and can be revisited separately — this PR is now only the `Error` fix. Those types are unchanged and still surface the serializer's own message.

## Testing

New tests in `tests/python/test_error.py`: exception and bare-`Error` pickle round trips, `deepcopy`, NULL handle, the `extra_context` drop, and re-propagating a restored exception through an FFI call (which exercises `set_last_ffi_error`'s `update_backtrace` on the restored object).

Full Python suite: 2421 passed, 52 skipped, 2 xfailed. `pre-commit` clean on all changed files.


python/tvm_ffi/cython/object.pxi
```diff
@@ -47,6 +48,54 @@ def _new_object(cls):
     return cls.__new__(cls)
 
 
+# Types that hold a live process resource (a code pointer, a device allocation,
+# a dlopen handle, a PyObject*) and therefore have no value representation that
+# survives a pickle round trip. Maps type key -> explanation of the remedy.
+_UNPICKLABLE_TYPE_REASONS: dict[str, str] = {
```
tqchen: we can skip this for now and focus on the Error reduce

tqchen (approved):

PR #712, opened 2026-08-26
[FEAT][RUST] Add AnyCompatible::to_any.

Seven-Streams:
## Summary

Add `AnyCompatible::to_any(&self) -> Any`, the by-reference counterpart of `Any::from(value)`.

`From<AnyView> for Any` now dispatches inline instead of always calling `TVMFFIAnyViewToOwnedAny`: an object increfs, a self-contained cell is a bitwise copy, and only the three borrowed forms Rust never produces (`kTVMFFIRawStr`, `kTVMFFIByteArrayPtr`, `kTVMFFIObjectRValueRef`) still go to the runtime, behind `#[cold] #[inline(never)]`. This mirrors C++ `details::InplaceConvertAnyViewToAny`. `is_plain_inline` moves from `extra::structural_common` into `any.rs`.

## Motivation

`Any::from` takes ownership. A value reachable only through a shared reference — a field behind a `Deref` into object storage, e.g. `node.a` on a `&AddObj` — cannot be moved out, so every such site has to write `Any::from(node.a.clone())` today.

`impl From<&T> for Any` cannot be added instead: `&` is fundamental, so a downstream crate may implement `AnyCompatible` for its own `&T` and the two impls overlap (E0119). A provided trait method has no such conflict, and every `AnyCompatible` type — including `Option<T>` and the containers — picks it up for free.

## Usage

```rust
fn first_operand(node: &AddObj) -> Any {
    node.a.to_any() // was: Any::from(node.a.clone())
}
```

`to_any()` leaves the borrowed value usable and retains object-backed values by increfing them.

## Efficiency and Test

`to_any` costs whatever `From<AnyView> for Any` costs, which no longer crosses the C ABI on every conversion. Where the type index is a compile-time constant (`i64`, `Array`, `Map`) the dispatch folds away and `to_any` matches `Any::from`. For `String`/`Bytes` and derived object refs the index is read at runtime, so the normalizing branch survives and the caller keeps a stack frame. That is deliberate: a type whose view is a borrowed representation — the Rust counterpart of C++ `RValueRef<T>` — needs that branch to stay.

The result is on par with `Any::from`: neither emits a call or an unwind path, and on the object path both do the same single `lock incq`. Verified on the release asm (x86-64, `-C codegen-units=1`).

Tests in `tests/test_any.rs`: `to_any` matches `Any::from` across the scalar, small-string/bytes, and object representations and increfs exactly once (given back on drop); and it works on fields reached through `Deref` — `String`, `Option<String>`, `i64` — where `Any::from` does not compile.

Docs: new "Converting Borrowed Values into `Any`" section in `docs/guides/rust_lang_guide.md`.


tlopex (approved):

tlopex:
LGTM @tqchen

PR #715, opened 2026-08-28
[FEAT][RUST] Support Any elements in FFI containers

tlopex:
This PR adds support for heterogeneous Rust FFI containers such as:

- `Array<Any>`
- `Map<String, Any>`
- `Map<Any, Any>`

`Any` cannot implement `AnyCompatible` because doing so would conflict with Rust's existing identity conversion from `Any` to `Any`. This PR therefore introduces a sealed internal `ContainerElement` trait.

All existing `AnyCompatible` types implement `ContainerElement` automatically, while `Any` receives a dedicated implementation. Regular typed containers continue to use their existing conversion behavior.

This allows Rust bindings and generated IR wrappers to represent heterogeneous TVM fields such as attributes and annotations using the standard `Array` and `Map` types, without introducing a separate `AnyMap` abstraction.

The PR covers:

- Container construction and iteration
- Strict type checking and element-wise conversion
- `Any` and `AnyView` round trips
- Object reference-count preservation
- Passing heterogeneous containers through typed packed calls

The complete Rust workspace test suite passes.

### DIFF
--- a/rust/tvm-ffi/src/type_traits.rs
+++ b/rust/tvm-ffi/src/type_traits.rs
@@ -83,126 +83,106 @@ pub unsafe trait AnyCompatible: Sized {
     }
 }
 
-/// A value that can be stored in an FFI container.
+/// Marker for a value that can be stored in an FFI container.
 ///
-/// Every [`AnyCompatible`] type implements this trait automatically. [`Any`]
-/// is supported as the open-value case, allowing heterogeneous containers such
-/// as `Array<Any>` and `Map<String, Any>`. It cannot implement
-/// [`AnyCompatible`] directly because that would overlap with Rust's identity
-/// `From<Any> for Any` implementation.
+/// This is an implementation detail of [`crate::Array`] and [`crate::Map`].
+/// Users should implement [`AnyCompatible`]; the blanket implementation below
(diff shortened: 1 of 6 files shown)


rust/tvm-ffi/src/type_traits.rs
```diff
@@ -83,6 +83,129 @@ pub unsafe trait AnyCompatible: Sized {
...
+/// is supported as the open-value case, allowing heterogeneous containers such
+/// as `Array<Any>` and `Map<String, Any>`. It cannot implement
+/// [`AnyCompatible`] directly because that would overlap with Rust's identity
+/// `From<Any> for Any` implementation.
+///
+/// This trait is sealed because its methods carry the same ownership and type
+/// safety contracts as [`AnyCompatible`]. Downstream types should implement
+/// [`AnyCompatible`] instead.
+#[doc(hidden)]
+pub unsafe trait ContainerElement: container_element_sealed::Sealed + Sized {
```
Seven-Streams: I'm wondering whether there will be a conflict between any compatible and container element

Seven-Streams (approved):

tqchen (approved):

PR #719, opened 2026-08-30
[CORE] Propagate static object type bounds in Any conversions

tqchen:
Add narrow static-object type-index assumptions to the non-null `ObjectPtr`, `ObjectRef`, and raw-object conversion paths into `AnyView` and `Any`.

Runtime validation for unconstrained incoming `Any` and `AnyView` values remains unchanged.

Validation: all 471 enabled CTest entries passed. The GCC 11 debug build required demoting only the pre-existing `-Werror=redundant-move` warning in `include/tvm/ffi/container/array.h:867`.

### DIFF
--- a/include/tvm/ffi/object.h
+++ b/include/tvm/ffi/object.h
@@ -1436,6 +1436,7 @@ struct TypeTraits<ObjectPtr<TObject>,
     }
     TVMFFIObject* obj_ptr = details::ObjectUnsafe::TVMFFIObjectPtrFromObjectPtr(src);
     result->type_index = obj_ptr->type_index;
+    TVM_FFI_UNSAFE_ASSUME(result->type_index >= TypeIndex::kTVMFFIStaticObjectBegin);
     result->zero_padding = 0;
     TVM_FFI_CLEAR_PTR_PADDING_IN_FFI_ANY(result);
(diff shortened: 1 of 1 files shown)


tqchen:
Assembly and benchmark follow-up:

Using GCC 11.4 with `-O3 -DNDEBUG`, 50,000,000 iterations per sample, median-of-five sampling, and three runs pinned to CPU 0:

- `const ObjectType* -> AnyView -> const ObjectType*`: 2.50381 ns/op before, 2.35846 ns/op after (5.8% lower). Identity baselines were 1.39015 and 1.39103 ns/op.
- `ObjectRef -> Any -> move cast<ObjectRef>`: 2.78202 ns/op before, 1.66852 ns/op after (40.0% lower). The patched path matches its 1.66948 ns/op identity baseline.

`objdump -d -Mintel` confirms that the AnyView round trip no longer emits the redundant `cmp type_index, 0x3f; jle cold` lower-bound branch, while retaining object-type compatibility validation. The Any move round trip no longer loads/tests `type_index` and has no error or destructor call path; it reduces to ownership pointer moves plus the nullable-object branch.

spectrometerHBH (approved):

PR #741, opened 2026-09-04
[REFACTOR][EXTRA] Optimize StructuralMap engine

tqchen:
This PR improves the `StructuralMap` engine by bringing lambda callback style to direct calling style. We also internally used the raw `TVMFFIAny` ABI path so engine overhead is minimized in the default path.

Main changes:

- **Direct calling style.** The callback chain was continuation-passing, a lambda nested per link. The mutator now owns its callbacks and tests them with a `(... || ...)` fold; selection, identity remap, descent, and invocation are one straight-line function.
- **Raw `TVMFFIAny` on the ABI boundary.** Hooks are C-ABI function pointers returning `TVMFFIAny`, a 16-byte POD that passes in registers, while `Expected<Any>` is classified MEMORY and forces the result to the stack. Descent through an unmatched node is that boundary, so it now stays raw end to end. Hook bodies are unchanged: `TVM_FFI_S_MUTATE_ASSIGN_OR_RETURN` still takes a typed `Expected<Any>` and moves out only on return.
- **Split static and dynamic selection.** `StructuralMapMutatorObj` (header) and `StructuralMapDynMutatorObj` (`.cc`) now share only the identity remap. This also removes a hazard: the dynamic form kept its selected `Function` in mutable state, where a post-order walk could descend into a matching child and overwrite the parent's selection.
- **Error context is attached by the engine, not by hooks.** The engine names a node where it dispatches into that node, giving one frame per node instead of two on some paths. Both `MAYBE_EARLY_RETURN` macros lose their node argument, and the visit engine no longer routes its own walk through the hook-facing macro.
- **API.** Removes `TVM_FFI_S_MUTATE_ASSIGN_FROM`, `SMutateResultUnchanged`, `AssignOrReturnHelper`, and `MutateWithIdentityRemapExpected`. Adds `TVM_FFI_UNSAFE_S_MUTATE_ASSIGN_OR_RETURN_SKIP_CHECK` for hooks that have measured the type check to matter; it is UB on a wrong-typed result and has no call sites here. Also fixes `AnyUnsafe::CheckAnyStrict`, which was missing the `Any` special case both of its siblings had.
- **Behavior change.** When descent hands back a node whose type no longer matches the link that selected it, the engine reports a `TypeError` instead of passing the node through. Default mutation must preserve a node's type, so this only fires when a hook has broken that invariant. Match-before-descent ordering, identity-remap entry and exit, and the pre-order in-place rules are unchanged.

### DIFF
--- a/include/tvm/ffi/any.h
+++ b/include/tvm/ffi/any.h
@@ -609,7 +609,12 @@ struct AnyUnsafe : public ObjectUnsafe {
 
   template <typename T>
   TVM_FFI_INLINE static bool CheckAnyStrict(const Any& ref) {
-    return TypeTraits<T>::CheckAnyStrict(&(ref.data_));
+    if constexpr (!std::is_same_v<T, Any>) {
+      return TypeTraits<T>::CheckAnyStrict(&(ref.data_));
+    } else {
+      // Any holds any value, so there is nothing to check against.
+      return true;
+    }
   }
 
   template <typename T>
--- a/include/tvm/ffi/expected.h
+++ b/include/tvm/ffi/expected.h
@@ -348,29 +348,6 @@ class Expected<void> {
 
 namespace details {
 
-// Helper for TVM_FFI_S_MUTATE_ASSIGN_OR_RETURN.
-class AssignOrReturnHelper {
- public:
-  TVM_FFI_INLINE explicit AssignOrReturnHelper(Any&& data) : data_(std::move(data)) {}
-
-  template <typename T>
-  TVM_FFI_INLINE Expected<T> TryMove() && {
-    if constexpr (!std::is_same_v<T, Any>) {
-      const TVMFFIAny* data = AnyUnsafe::TVMFFIAnyPtrFromAny(data_);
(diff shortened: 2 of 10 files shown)


tqchen:
## Speed

Measured on a nested tree fixture, `-O3`, ns/node, medians of 11 pinned process runs. Both sides are self-consistent git states built identically.

| | before | after | change |
|---|---:|---:|---:|
| hooked `StructuralMap` (replace) | 39.35 | **28.58** | **−27.4%** |

Cross-confirmed on a standalone harness with the same node shapes — a non-final expression base with 64 child slots and a type field, a `FreeVar`-kind variable taking an identity remap on every match, a view type whose field check costs a type test plus a dereference plus a second type test, and binary nodes carrying a result-type guard. Its engine cost over its own floor tracks the full fixture's to within 1% (+15.51 vs +15.33 ns/node). The harness is not part of this PR; it lives on `bench-structural-mutate-harness`.

spectrometerHBH (approved):

tlopex (approved):

PR #744, opened 2026-09-05
[REFACTOR][EXTRA] Move structural-mutate error construction out of the hot path

tqchen:
`TryLink` inlines into the per-node dispatch function, so two error constructions that only the matched path can reach put their string and `Error` construction into a stack frame every node pays for.

This follows `BadStructuralMutateHookError`, which is already `TVM_FFI_COLD_CODE static` for exactly this reason — its doc comment says *"Inlined, its three string literals and Error construction land in the traversal's hot path for no reason."* These were the two sites it missed.

No semantic change: both checks still fire identically, only the location of their error construction moves.

Per-node dispatch function: stack frame 216 → 120 bytes, 358 → 318 instructions. Both helpers land in `.text.unlikely.*`.

### DIFF
--- a/include/tvm/ffi/extra/structural_mutate.h
+++ b/include/tvm/ffi/extra/structural_mutate.h
@@ -612,17 +612,24 @@ namespace details {
 
 /// \endcond
 
+// Out of line so its strings and Error construction stay out of the hot path of whatever hook
+// body TVM_FFI_S_MUTATE_ASSIGN_OR_RETURN expands into. Same reason as
+// BadStructuralMutateHookError.
+TVM_FFI_COLD_CODE inline TVMFFIAny SMutateDeclaredTypeErrorRaw() noexcept {
+  return AnyUnsafe::MoveAnyToTVMFFIAny(
+      Any(Error("TypeError", "structural mutate result does not match the declared type", "")));
+}
+
 /// \cond Doxygen_Suppress
-#define TVM_FFI_S_MUTATE_ASSIGN_OR_RETURN_IMPL_(Result, Type, Name, ResultExpr)                  \
-  auto Result = (ResultExpr); /* NOLINT(bugprone-macro-parentheses) */                           \
-  TVM_FFI_S_MUTATE_MAYBE_EARLY_RETURN(Result);                                                   \
-  if (TVM_FFI_PREDICT_FALSE(!::tvm::ffi::details::AnyUnsafe::CheckAnyStrict<Type>(               \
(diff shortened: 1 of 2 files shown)


tqchen:
## Speed

Standalone harness mirroring the TIR node shapes, `-O3` for library and benchmark, pinned. Baseline and candidate libraries built identically; the two binaries run **interleaved over 11 pairs**, medians in ns/node.

| | before | after | |
|---|---:|---:|---:|
| no link matches | 14.12 | 13.80 | −2.3% |
| link matches, no rebuild | 21.37 | 20.79 | −2.7% |
| link matches, node replaced | 22.79 | 21.22 | **−6.9%** |

The first row is the one that generalizes: it is what a node runs when no link matches, which is most nodes on a real tree. This fixture matches 4 nodes of 17, a far higher rate than a realistic traversal.

tlopex (approved):

tqchen:
Correction to the measurements above, from re-running the A/B against the exact pushed head.

The **−6.9% on the replace arm does not reproduce.** On the final branch state it measures +0.6%, and its baseline also shifted from 22.79 to 21.23 ns/node between runs — same `origin/main` library, different benchmark binary. Do not rely on that row.

What holds, 13 interleaved pairs against identically built libraries:

| arm | base | new | delta |
|---|---:|---:|---:|
| minimal-vtable floor *(control)* | 8.84 | 8.98 | +0.14 (+1.6%) |
| no link matches | 14.56 | **13.88** | **−0.68 (−4.7%)** |
| link matches, no rebuild | 21.83 | **20.91** | **−0.92 (−4.2%)** |
| link matches, node replaced | 21.23 | 21.36 | +0.13 (+0.6%) |

The floor is a control that should not move, and its +1.6% drift sets the real noise level at about ±0.15 ns/node. The two improvements are 4–6× that and their medians are well separated; the replace row is inside it.

Two caveats on this harness, both found the hard way and both worth knowing before trusting any number from it:

- Absolute values shift with code layout. Registering unrelated hooks in the same binary moved a number by 17%.
- Arms timed sequentially in one process are not comparable either. In the current binary the replace arm measures *cheaper* than the no-rebuild arm, which is backwards, and the difference tracks position in `main()` rather than work done.

The codegen effect is not subject to either problem and is unambiguous: per-node dispatch frame 216 → 120 bytes and 358 → 318 instructions, maybe-in-place 232 → 120 and 421 → 352, with both error builders emitted into `.text.unlikely`.

Issue #727, opened 2026-09-02
TVMFFIBacktrace crashes under concurrency on Windows: DbgHelp used without serialization, per-call SymInitialize/SymCleanup, unchecked SymInitialize


pisarev:
**Component:** `tvm-ffi` — `src/ffi/backtrace_win.cc`
**Affected version:** `apache-tvm-ffi` 0.1.13.post3 (confirmed in the published
sdist and wheel), also present in older trees
**Platform:** Windows x64, MSVC
**Severity:** a C++ exception raised while another thread is also reporting one
kills the process, and the original error message is never printed

## Summary

The crash originates in unsynchronized access to shared DbgHelp symbol-handler
state. Repeated `SymInitialize`/`SymCleanup` around every backtrace and eager
symbol loading increase the amount of DbgHelp work performed on that path. In
addition, the return value of `SymInitialize` is ignored, so an initialization
failure can leave subsequent symbol-handler calls operating without a
successfully established session.

The practical effect is worse than a crash in the error path. The *original*
exception is never printed, because the process dies while the library is
formatting the diagnostic for it. Every failure inside a multi-threaded TVM pass
therefore looks like a silent process death with no message at all.

That is how it turned up here. MetaSchedule tuning on Windows died at task
initialization with exit code 127, no Python traceback and no log line. It took
a native stack under `cdb` to see that the visible crash was in the *reporter*,
not in the code being reported on:

```
tvm_compiler!tvm::s_tir::SampleComputeLocation+0x5a
tvm_compiler!tvm::s_tir::ComputeInline+0x35b
tvm_ffi!TVMFFIBacktrace+0x257
dbghelp!SymInitialize -> dbghelp!LoadModule   <- access violation
```

The C++ exception was raised on thread `5bec`; the access violation happened on
thread `6c34`.

## What is wrong

`src/ffi/backtrace_win.cc`, lines 53–58 and 141:

```cpp
HANDLE process = GetCurrentProcess();
HANDLE thread = GetCurrentThread();

SymSetOptions(SYMOPT_LOAD_LINES | SYMOPT_UNDNAME);
SymInitialize(process, NULL, TRUE);     // line 58
...
SymCleanup(process);                    // line 141
```

**One correctness bug:**

1. **No serialization.** The DbgHelp documentation states that its functions are
   single-threaded and that calling them from more than one thread leads to
   unexpected behaviour or memory corruption; the caller is required to
   serialize them. There is no lock here. This applies to the whole body —
[...]

PR #754, opened 2026-09-06
[WIN] Enhance threadsafety of DbgHelp

tqchen:
`src/ffi/backtrace_win.cc` called DbgHelp unguarded, though DbgHelp is
single-threaded and requires the caller to serialize it. The process died while
formatting a backtrace for another thread's exception, so the original error was
never printed.

- One mutex around every DbgHelp call made by tvm_ffi.
- A duplicated process handle as the session key. `GetCurrentProcess()` has
  the same value in every component, so another DbgHelp user in the process could
  close our symbols. A fresh key also makes `SymInitialize`'s failure
  unambiguous, so its result can be checked.
- Initialize once and keep the session, dropping the per-call
  `SymInitialize`/`SymCleanup` and with it module enumeration inside the lock.

There is deliberately no destructor: `SymCleanup` at static-destruction time
would reintroduce teardown at an unpredictable point relative to DLL unload.

A failed initialization now returns a backtrace with only the caller's own frame.
`StackWalk64` resolves through the session, so without one there is no unwinding
at all.

### DIFF
--- a/src/ffi/backtrace_win.cc
+++ b/src/ffi/backtrace_win.cc
@@ -32,9 +32,50 @@
 #include <tvm/ffi/error.h>
 
 #include <iostream>
+#include <mutex>
 #include <vector>
 
 #include "./backtrace_utils.h"
+
+namespace {
+
+/*! \brief Global singleton holding tvm_ffi's DbgHelp symbol session. */
+struct DbgHelpSession {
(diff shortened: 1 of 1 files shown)


tqchen:
fixes #727 , thanks to @pisarev

tlopex (approved):

PR #766, opened 2026-09-08
[Rust][Fix] Preserve object lifetime and thread-safety contracts

tlopex:
This PR addresses object construction and lifetime issues in Rust bindings without changing the C ABI.
- Add a no-alloc stubgen directive to preserve readable fields while suppressing generated allocators for registry-owned types and their descendants.
- Reuse the existing nullable and custom-new directives for fields that native code may move out, without introducing a separate nullable-storage mechanism.
- Make objects non-Send/non-Sync by default, including erased and borrowed views. Keep Function shareable by requiring thread-safe callback captures.
- Require unique ownership for mutable ObjectArc access, align reference-count ordering with C++, and preserve allocation metadata when weak references outlive objects with trailing storage.

These changes tighten Rust source-level contracts. Downstream code relying on cross-thread object sharing or non-thread-safe callback captures may require adaptation.

### DIFF
--- a/python/tvm_ffi/stub/rust_generator/generator.py
+++ b/python/tvm_ffi/stub/rust_generator/generator.py
@@ -50,7 +50,7 @@ class RustGenerator:
     syntax = C.RUST_SYNTAX
     source_exts = frozenset({".rs"})
     directive_kinds = C_RUST.RUST_DIRECTIVE_KINDS
-    shared_directive_kinds = frozenset({"no-alloc"})
+    shared_directive_kinds = frozenset({"no-alloc", "nullable", "opaque"})
 
     def default_ty_map(self) -> dict[str, str]:
         """Return the default FFI-origin -> Rust-type name map."""
--- a/python/tvm_ffi/stub/rust_generator/generator.py
+++ b/python/tvm_ffi/stub/rust_generator/generator.py
@@ -69,7 +69,6 @@ class RustGenerator:
         else:
             imports.directives.add(name, payload, lineno)
             if name == "no-alloc":
-                # A misspelled type must not silently leave its allocator enabled.
                 object_info_from_type_key(payload.strip())
 
     def canonical_type_name(self, type_key: str) -> str:
--- a/python/tvm_ffi/stub/rust_generator/generator.py
+++ b/python/tvm_ffi/stub/rust_generator/generator.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from .. import consts as C
+from ..lib_state import object_info_from_type_key
 from . import consts as C_RUST
 from .codegen import (
     finalize_rust_module_tree,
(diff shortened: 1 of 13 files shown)


python/tvm_ffi/stub/rust_generator/generator.py
```diff
@@ -66,6 +68,9 @@ def add_directive(self, imports: RustImports, name: str, payload: str, lineno: i
             imports.record(payload.split(";", 1)[0].strip())
         else:
             imports.directives.add(name, payload, lineno)
+            if name == "no-alloc":
+                # A misspelled type must not silently leave its allocator enabled.
```
Seven-Streams:
```suggestion
```

rust/tvm-ffi/src/function.rs
```diff
@@ -311,15 +317,26 @@ impl Function {
...
+    /// ```
     /// # Arguments
     /// * `func` - The packed function in signature of `Fn(&[AnyView]) -> Result<Any>`
     ///
     /// # Returns
     /// * `Function` - The function
     pub fn from_packed<F>(func: F) -> Self
     where
-        F: Fn(&[AnyView]) -> Result<Any> + 'static,
+        F: Fn(&[AnyView]) -> Result<Any> + Send + Sync + 'static,
```
Seven-Streams, line 339: will it break the current behavior?

rust/tvm-ffi/src/object.rs
```diff
@@ -25,14 +25,30 @@ pub use tvm_ffi_sys::TVMFFITypeIndex as TypeIndex;
...
     header: TVMFFIObject,
+    _thread_confined: std::marker::PhantomData<std::rc::Rc<()>>,
 }
 
-/// Arc-like wrapper for Object that allows shared ownership
+const _: () = {
+    assert!(std::mem::size_of::<Object>() == std::mem::size_of::<TVMFFIObject>());
+    assert!(std::mem::align_of::<Object>() == std::mem::align_of::<TVMFFIObject>());
+    assert!(std::mem::offset_of!(Object, header) == 0);
+};
```
Seven-Streams: Why add these tests?

python/tvm_ffi/stub/rust_generator/generator.py
tlopex: sure

rust/tvm-ffi/src/function.rs
tlopex: Yes, it can break existing code if the callback captures something like `Rc` or `Cell`. The problem is that `Function` could already be shared across threads, but we weren’t checking whether its callback was safe to share. Adding `Send + Sync` fixes that mismatch.

rust/tvm-ffi/src/object.rs
tlopex: would remove

Seven-Streams (approved):

tqchen (approved):

PR #769, opened 2026-09-08
[FFI][REFACTOR] Clarify DefRegionKind

tqchen:
Rename the two def-region kinds from Recursive / NonRecursive to Pattern /
Simple, with the matching field-flag bits and the Python, Rust and doc
mirrors: a pattern def region matches the value's type as a pattern, binding
its free variables; a simple def region defines the variable alone and walks
its type as uses. The comments now state the binding rule and the clamp that
custom hooks on FreeVar types must apply.

A pattern region propagates: entering any kind inside it has no effect, so a
simple def nested under a pattern (a pattern field, or map_free_vars at the
top level) still binds the free variables in its type. Applied in the equal,
hash, visit and mutate engines and in the Rust visitor and mutator, with a
test for the nested case.

Python keeps "def" as the only alias of "def-pattern".

No ABI changes involved: enum values and flag bit positions are unchanged.

### DIFF
--- a/docs/concepts/structural_eq_hash.rst
+++ b/docs/concepts/structural_eq_hash.rst
@@ -635,14 +635,14 @@ Use for:
   redundant to compare.
 - **Debug annotations** — names, comments, metadata for human consumption.
 
-``structural_eq="def-recursive"`` / ``"def-non-recursive"`` — Definition region
+``structural_eq="def-pattern"`` / ``"def-simple"`` — Definition region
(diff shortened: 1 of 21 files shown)


tqchen:
Validation on this head: C++ test suite 493/493, `cargo test --workspace` all suites ok, `cargo check` clean, clang-format clean.

cyx-6 (approved):

PR #770, opened 2026-09-08
[Rust] Correct object ownership and callback threading contracts

tlopex:
This PR updates Rust object and callback safety using the existing APIs, without introducing thread-local functions or new C ABI interfaces.

- Remove the blanket `Send + Sync` bounds from `Function::from_packed` and `Function::from_typed`. These constructors become `unsafe`: callers must preserve the captures’ threading requirements, including calls and final release through native copies.
- Remove unconditional `Send`/`Sync` implementations for `Function`. Make global registration `unsafe`, since registered functions become accessible across threads.
- Remove `ObjectArc::DerefMut`, which could overwrite a live reference-count header even under unique ownership. Keep the common object header free of thread-confinement markers, without inferring that arbitrary object handles are thread-safe.
- Make reflected field getters accept owning handles and update stubgen accordingly.
- Fix Rust-allocated arrays to release their live elements correctly, including after native mutation or partial initialization.
- Handle empty shapes without constructing slices from null pointers.

These are Rust source-compatibility changes. The C ABI and object layouts remain unchanged.


Seven-Streams (approved):

tqchen (approved):

rust/tvm-ffi/src/collections/array.rs
```diff
@@ -45,7 +46,29 @@ pub struct ArrayObj {
 unsafe impl ObjectCoreWithExtraItems for ArrayObj {
     type ExtraItem = TVMFFIAny;
     fn extra_items_count(this: &Self) -> usize {
-        this.size as usize
+        this.capacity as usize
+    }
+}
+
+impl ArrayObj {
```
Seven-Streams, line 53: why we need to add it?

rust/tvm-ffi/src/macros.rs
```diff
@@ -49,13 +49,19 @@ macro_rules! function_name {
 #[macro_export]
 macro_rules! cached_global_func {
     ($name:literal) => {{
-        static FUNC: std::sync::LazyLock<$crate::function::Function> =
+        static FUNC: std::sync::LazyLock<std::sync::atomic::AtomicPtr<$crate::function::Function>> =
```
Seven-Streams: do we have to?

tlopex: here we remove send+sync of func so we have to modify it

rust/tvm-ffi/src/collections/array.rs
tlopex: here fixes a bug of memory leak of array

PR #774, opened 2026-09-10
[PERF] Forward single StructuralMutate callbacks directly

tqchen:
Forward single StructuralMutate callback results directly and use bool-match/result-output dispatch for multiple callbacks. Preserve owning conversions, typed fallback, first-match behavior, and cold error-context attachment.

### DIFF
--- a/include/tvm/ffi/extra/structural_mutate.h
+++ b/include/tvm/ffi/extra/structural_mutate.h
@@ -585,8 +585,8 @@ class StructuralMutatorObj : public Object {
   }
 
   /*! \brief Keep address-taking for error decoration off the successful raw-result path. */
-  TVM_FFI_COLD_CODE TVM_FFI_NO_INLINE static TVMFFIAny AnnotateDefaultErrorRaw(
-      TVMFFIAny result, AnyView value) noexcept {
+  TVM_FFI_COLD_CODE static TVMFFIAny AnnotateDefaultErrorRaw(TVMFFIAny result,
+                                                             AnyView value) noexcept {
     details::UpdateVisitErrorContext(result, value);
     return result;
   }
@@ -1608,8 +1608,8 @@ class StructuralMutateEngine : public Parent {
 
   /*! \brief Mutate one value, handing a matched callback ownership of descent. */
   TVMFFIAny MutateImplRaw(AnyView value) noexcept {
-    if (std::optional<Expected<Any>> matched = DispatchCallbacks(value, false)) {
-      Expected<Any> result = *std::move(matched);
+    Expected<Any> result{Any()};
+    if (DispatchCallbacks(value, false, &result)) {
       if (TVM_FFI_PREDICT_FALSE(result.is_err())) {
         // Keep callback-boundary context in addition to the default-descent
         // context: a callback may return a rebuilt value, so the two nodes can differ.
@@ -1622,8 +1622,8 @@ class StructuralMutateEngine : public Parent {
 
   /*! \brief Maybe mutate one value in place, with callback-owned descent. */
   TVMFFIAny MaybeInplaceMutateImplRaw(AnyView value) noexcept {
-    if (std::optional<Expected<Any>> matched = DispatchCallbacks(value, true)) {
-      Expected<Any> result = *std::move(matched);
+    Expected<Any> result{Any()};
+    if (DispatchCallbacks(value, true, &result)) {
       if (TVM_FFI_PREDICT_FALSE(result.is_err())) {
         // Keep callback-boundary context in addition to the default-descent
         // context: a callback may return a rebuilt value, so the two nodes can differ.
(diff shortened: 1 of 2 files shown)


tqchen:
Validation at `0c9a893316fa73dcab59b4e189af6e335fac3073`: all 500 C++ tests pass with AddressSanitizer, including 28 focused structural mutation/map/result-protocol tests. The new single-versus-multi regressions cover owning match lifetimes, custom Parent behavior, returned/thrown errors, result carriers, first-match short-circuiting, typed fallback, and both mutation entries. Formatting and whitespace checks pass. The local GCC build uses `-Wno-error=redundant-move` for the existing warning in `array.h`; no unrelated source change is included.

Preliminary x86-64 benchmark, nanoseconds per visited occurrence:

| Control | Base `fb8e1024` | This change |
| --- | ---: | ---: |
| AnyView normal-hook passthrough | 14.773 | 5.806 |
| Owning Any normal-hook passthrough | 29.476 | 20.142 |
| Typed pointer normal-hook hit | 14.489 | 6.901 |
| Typed miss/default fallback | 7.621 | 6.135 |
| Direct default, error checking retained | 6.252 | 5.887 |
| Raw hook floor | 5.413 | 4.819 |

Same GCC 11.4 `-O3 -DNDEBUG -std=c++20` build flags, strong stack protection, callback source, fixture, and timing boundaries in both builds. Passthrough callbacks are explicitly inlined in both comparisons. The retained, unchanged expression fixture has 17 visited occurrences (15 unique nodes). Each cell uses ten interleaved independent processes; each process reports the median of nine 4096-traversal batches after warmup. Engine construction is timed; fixture construction and output destruction are not. The AnyView process range is 14.611–15.054 before and 5.685–6.632 after.

The AnyView vtable entry now contains the hook dispatch directly: no extra entry trampoline, callback-selection branch, matched-result stack spill, or canary frame. It retains one hot error-tag branch and forwards the raw result in registers. Both default-hook and Parent callback annotations remain on the error path. This codegen finding depends on the callback actually inlining; the specialization does not force arbitrary user callbacks to inline.

The callback is close to direct default here. Control medians also moved (raw floor −11.0%, direct default −5.8%), so the entire timing delta is not an isolated measurement of callback selection: code layout and the shared default-error helper also changed. The callback engine constructs additional map state compared with the two minimal controls. Owning Any retains its required reference management, and typed callbacks retain their type probe.

tqchen:
The test-only follow-up d0b642764eac6aefde9820d2683c3c2323b40414 addresses the clang-tidy findings: preserve the reference-count integer type and inspect read-only helper arguments by const reference. Targeted clang-tidy 21.1.1 and all 500 C++ tests with AddressSanitizer pass locally. The production header is unchanged, so the assembly and benchmark evidence above still applies.

tqchen:
The revised head `b6785df8` uses cold-only error helpers, the `AttachVisitErrorContextRaw` name, and bool-hit/result-output matching for multiple callbacks. The net test addition is one short single-callback delegation test; all pre-existing tests remain. All 497 C++ tests pass with AddressSanitizer, targeted clang-tidy 21.1.1 passes, and whitespace checks pass.

Matched GCC 11.4 `-O3 -DNDEBUG -std=c++20` measurements compare `d0b6427` with the cold-only/bool-output change `95ac35d` (before the final name-only commit):

| Control | Before ns/visit (process range) | After ns/visit (process range) |
| --- | ---: | ---: |
| Single AnyView passthrough | 5.829 (5.766–8.532) | 6.350 (6.290–6.464) |
| Multi: first AnyView hit | 14.378 (14.284–14.544) | 10.031 (9.967–10.190) |
| Multi: typed miss, then AnyView hit | 16.887 (16.788–17.029) | 10.004 (9.958–10.162) |
| Multi: all typed misses | 7.064 (5.909–7.385) | 7.002 (6.921–7.243) |
| Direct default, errors retained | 6.103 (6.086–6.166) | 6.037 (5.967–6.966) |
| Raw hook floor | 4.815 (4.794–6.274) | 4.812 (4.761–4.891) |

Ten interleaved processes per cell/build; identical explicitly inlined passthrough, 17-visit unchanged/retained fixture, and timing boundaries as above. Each process reports the median of nine 4096-traversal batches after warmup. Fresh engine construction is timed; fixture construction and result destruction are not. Controls have stable medians here, but ranges include outliers and this compares the combined revisions, not an isolated instruction cost.

At the final renamed head, single AnyView still forwards the result in registers with one hot Error check, no callback match/fallback branch, and no entry trampoline or result spill. Unlike the previous forced-noinline build, GCC inlines callback error decoration and restores a 32-byte frame plus canary. For multi-callback first-AnyView matching, the optional engaged byte and later selection/fallback disappear, but result-output stack storage and cleanup remain. Typed misses retain their required probes. Both default and callback error contexts are preserved; codegen depends on the callback actually inlining.

tlopex (approved):

PR #777, opened 2026-09-11
[EXTRA] Enable typed mutation result conversions and subsumption

tqchen:
Typed structural mutation helpers need to convert compatible `UnchangedOr` results and propagate child errors across different success types. The assignment macro also repeats checks for already compatible typed results.

Add implicit `UnchangedOr` conversions and storage-subsumption rules for `UnchangedOr` and `Expected`. Initialize subsumed conversions directly from their stored carrier, avoiding an intermediate assignment. Use subsumption to omit redundant assignment checks while retaining checks for erased or narrower sources, and propagate child errors through the existing error-only return helper. Correct the borrowed `AnyView` adapter for wrapper values.

### DIFF
--- a/include/tvm/ffi/expected.h
+++ b/include/tvm/ffi/expected.h
@@ -76,9 +76,11 @@ Unexpected(E) -> Unexpected<E>;
 template <typename T>
 class Expected;
 
+/// \cond Doxygen_Suppress
 /*! \brief Whether an Expected success value can reuse another success value's storage. */
 template <typename T, typename U>
 inline constexpr bool type_subsumes_v<Expected<T>, Expected<U>> = type_subsumes_v<T, U>;
+/// \endcond
 
 namespace details {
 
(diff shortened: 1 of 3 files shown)


tqchen:
Validation for the reduced test in 53ccd417: the affected C++ target was rebuilt, all 36 focused ASan tests passed, and clang-tidy 21.1.1 and changed-line clang-format passed. The subsequent header edits add only Doxygen suppression comments; the documentation build is being checked in CI.

The implementation was previously validated at 3abde92c with 498 enabled C++ ASan tests and 58 Python structural tests after an editable rebuild. No library behavior changed in this follow-up.

tqchen:
Validation at 5c7e059e: 498 enabled C++ ASan tests passed, followed by an editable rebuild and 58 Python structural tests. Clang-tidy 21.1.1 passed on both relevant test translation units, and changed-line formatting passed.

With GCC 11.4, both wrapper-conversion probes emit the same normalized instructions as their explicit raw-carrier counterparts. A controlled TVM expression-dispatch comparison using the same headers produced byte-identical compiler shared libraries for implicit and explicit raw conversions (SHA256 `c7e07d607d4b9b95142cf056e7351e9c81d4fd4d92ad6234b3fd7d16df650a13`).

spectrometerHBH (approved):

PR #779, opened 2026-09-12
[IR] Expose protected remap state and skip empty lookups

tqchen:
Expose `var_remap_` as protected state so derived structural map engines can inspect the remap environment. Skip the identity-map lookup when the environment is empty, after validating the key.

### DIFF
--- a/include/tvm/ffi/extra/structural_mutate.h
+++ b/include/tvm/ffi/extra/structural_mutate.h
@@ -1131,6 +1131,7 @@ class StructuralMapEngineBase : public StructuralMutatorObj {
   template <typename Parent>
   friend class details::StructuralMutateDynEngine;
 
+  /*! \brief Identity-substitution environment keyed by object identity. */
   // Raw-pointer key: IncRef once on first insert, DecRef all keys in the destructor.
   std::unordered_map<const Object*, Any> var_remap_;
 };
--- a/include/tvm/ffi/extra/structural_mutate.h
+++ b/include/tvm/ffi/extra/structural_mutate.h
@@ -1085,6 +1085,9 @@ class StructuralMapEngineBase : public StructuralMutatorObj {
     }
   }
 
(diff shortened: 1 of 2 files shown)


tqchen:
Validation for 7ac4a4f61f73cf7605591a4f0fac5a6963b96e97: all 26 `UnchangedOr`, `StructuralMap`, `StructuralMutate`, `StructuralMapDyn`, and `StructuralMutateDyn` C++ tests passed. Coverage includes empty/nonempty remap state, a valid empty lookup, and null/scalar key validation.

Rebuilt all 17 runtime/test objects depending on the changed header, verified the dependency contents of 47 reused objects, and linked the runtime, testing library, and test executable consistently. The test objects and inline header code used AddressSanitizer; the Debug runtime was uninstrumented and leak detection was disabled.

tqchen:
Updated in `1ec0fc40fd42e88b9dee352bd9b89dae5f36b1f6`: `StructuralMapEngineBase` uses one protected section for its remap state and implementation helpers. The accessor and its helper-only assertions are removed; no new tests were added. The existing lookup and insertion bodies are unchanged.

Validation: all 26 existing focused C++ tests passed after rebuilding the 17 header-dependent objects and verifying the dependencies of 47 reused objects, linked against the rebuilt runtime. Formatting passed. ASan covers test objects and inline headers only; the Debug runtime is uninstrumented and leak detection was disabled.

tqchen:
Added the missing Doxygen brief for protected `var_remap_` in `b08ff2c35bea7e9abcae61014abc97eb688cc0fa`. This is a one-line comment change.

Doxygen 1.9.8 header API generation passed with zero warnings using the repository configuration and Exhale defaults, with warnings treated as errors. The generated XML includes the protected member and its brief. Formatting passed; the full documentation job will run in CI.

cyx-6 (approved):

PR #780, opened 2026-09-12
[PERF] Reduce structural traversal result overhead

tqchen:
Keep compiled structural visit hooks on the raw TVMFFIAny path through default descent and callback selection to avoid repeated result materialization. Keep error-context decoration off the success path, inline WalkResult lifecycle operations, and retain typed Parent customization while preserving callback order, descent control, errors and interrupt ownership.

### DIFF
--- a/include/tvm/ffi/extra/structural_visit.h
+++ b/include/tvm/ffi/extra/structural_visit.h
@@ -221,32 +221,8 @@ class StructuralVisitorObj : public Object {
    * \return Expected interrupt state. An error means traversal failed.
    */
   TVM_FFI_INLINE Expected<Optional<VisitInterrupt>> DefaultVisitExpected(AnyView value) noexcept {
-    int32_t type_index = value.type_index();
-    static reflection::TypeAttrColumn column(reflection::type_attr::kStructuralVisit);
-    AnyView attr = column[type_index];
-
-    // case 1: Type-specific override registered as an opaque ABI visit function pointer.
-    if (attr.type_index() == TypeIndex::kTVMFFIOpaquePtr) {
-      auto* visit_fn = reinterpret_cast<FStructuralVisit>(attr.cast<void*>());
-      TVMFFIAny raw = (*visit_fn)(this, value);
-      return details::ExpectedUnsafe::MoveFromTVMFFIAny<Optional<VisitInterrupt>>(raw);
-    }
-
-    // case 2: Type-specific override registered as an ffi::Function.
-    if (attr.type_index() == TypeIndex::kTVMFFIFunction) {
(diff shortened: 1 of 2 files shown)


tqchen:
Validation for `0c685e4426787c99853ef4d5724d8a1983e76377` (source tree unchanged by the commit squash):

- All 498 existing C++ tests passed; two preexisting tests remain disabled. AddressSanitizer instruments the test executable; the shared runtime is a normal Debug build. The local GCC build demoted the existing redundant-move warning from Werror.
- The final diff preserves explicit `Optional<VisitInterrupt>` signatures and adds no tests. The only test-file change removes obsolete raw forwarding from an existing visitor layer.
- All six optimized compiled hooks and the fixture's 17 logical occurrences/four Var matches were checked outside timing. AnyView pure Advance callback machinery erases; the typed Var-count callback retains its type probe and increment, while Advance result cleanup and callback-control branches disappear. Required child-result propagation remains.

**Performance before versus after this PR — ns/node**

Split/fuse mini-TIR on an AMD EPYC 7413, GCC 11.4/O3, benchmark CPU 6 and timing driver CPU 46. Other local jobs and refresh commands were paused for the measurement window; CPU 6's sibling CPU 30 recorded no active scheduler ticks during the run. Each cell has 12 interleaved independent processes with all nine inner samples retained. Values are median [minimum, maximum] of process medians, normalized by 17 logical node occurrences. Walk includes managed visitor construction/destruction; the narrower compiled-hook floor uses a stack visitor.

| Case | Before PR | After PR |
| --- | ---: | ---: |
| Compiled-hook floor | 4.206 [4.185, 4.279] | 4.229 [4.211, 4.282] |
| Borrowed Var match, Advance | 15.595 [15.533, 15.740] | 6.260 [6.100, 6.344] |
| Borrowed Var match, count then Advance | 16.001 [15.935, 16.091] | 5.919 [5.884, 6.306] |

The before executable uses the same source tree as PR base `3b0e30b1`; the after executable was built from the final explicit-signature source tree, identical to the current PR head. Both preserved executables and their own runtimes were measured together, with source/artifact bindings and optimized hooks checked outside timing. All 72 process results and 648 inner samples passed an independent numerical and provenance audit.

The table uses the explicitly isolated run. A preceding run with weaker isolation produced higher, more variable timings for both versions; its full samples remain preserved. Repeating the identical binaries under the stronger isolation recovered the earlier performance range. This supports an environmental explanation, without identifying a particular competing job. Counting and pure Advance have different generated bodies, so their difference is not an isolated counting cost.

The AnyView hot paths save five registers without local result slots or cleanup calls; the typed fixture also retains a 32-byte local frame/canary. These measurements remain above the narrower compiled-hook floor.

cyx-6 (approved):

PR #771, opened 2026-09-09
[EXTRA] Simplify StructuralMap engine policy

tqchen:
[EXTRA] Simplify StructuralMap engine policy

StructuralMap needs identity state only to keep variables consistent when
descent rewrites a definition. Callback results are separate: callbacks run
at every occurrence in their selected pre- or post-descent position.

Make default C++ descent the sole owner of FreeVar and DAG remapping, with
definitions binding rewritten results and unchanged simple definitions left
absent. Keep the remap ABI stable and use an owned raw-pointer-keyed store for
efficient identity lookups.

Adopt whole raw TVMFFIAny results by value while leaving destructor-visible
slot moves pointer-based. Initialize owning Any storage member-wise so GCC
retains scalar state and avoids an overlapping narrow-store/wide-reload stall.

Update existing C++ and Python expectations in place and document
StructuralMap as post(D(pre)) alongside StructuralWalk's stateless tree
traversal. Rust remains unchanged for a follow-up.

PR #792, opened 2026-09-15
[Fix][Rust] Align structural mutation remapping policy

tlopex:
Rust structural map currently caches complete callback results for FreeVar and DAG identities, skipping callbacks at subsequent occurrences. Align it with the C++ policy from #771: run callbacks at every occurrence and handle identity substitutions only in default reflected descent or registered type hooks.

Share the default remapping path with structural mutate. Unbound FreeVar uses retain their identity without field descent; unchanged Simple definitions stay absent, while Pattern definitions and DAG nodes can cache Unchanged. Preserve owned remap keys and update the Rust guide.


docs/guides/rust_lang_guide.md
```diff
@@ -477,10 +477,13 @@ container's children are mapped forces the non-in-place path. Reflected
...
-is `FreeVar` or `DAGNode` are mapped once. The engine caches the complete final
-result (including callback replacement and child mapping), reuses that exact
-mapped identity at later occurrences, and does not invoke the callback again.
+Within one `structural_map` call, callbacks run at every occurrence, including
+repeated `FreeVar` and `DAGNode` identities. Default reflected recursion manages
+identity substitutions; callback replacements are not cached. A variable use
+without a recorded substitution retains the original without visiting fields.
+Unchanged simple definitions need no entry; pattern definitions and DAG nodes
+can cache `Unchanged` to skip repeated descent. Registered hooks manage their
+own remapping.
```
Seven-Streams: I think the doc can be simplified. We do not need to mention `FreeVar` and `DAGNode` again since we have aligned them.

Seven-Streams (approved):

spectrometerHBH (approved):

PR #794, opened 2026-09-15
[FEAT][Rust] Add composable policies for structural visit and walk

tlopex:
Add `VisitPolicy<State>` to customize default recursion independently of callbacks, corresponding to the C++ Parent extension mechanism. Attach policies through `VisitCallbacks::with_policy` or `WalkWithPolicy`; nested tuples compose policies sharing the callback state.

Add `VisitContext::default_visit_children(value, kind)` to continue the policy chain with an explicit value and definition region, bypassing the target's callback while its children re-enter full dispatch. Preserve Pattern precedence and restore the enclosing region on return, error, or interrupt. Keep `visit_children()` as the current-value/current-region shorthand.

Tests cover composition, shared state, generated walk dispatch, callback ordering, Skip, policy interrupts/errors, retargeted descent, and region propagation through reflected fields and Function-valued hooks.


rust/tvm-ffi/src/extra/structural_visit/layer.rs
```diff
@@ -0,0 +1,320 @@
...
+///         self.depths.push(self.depth);
+///         Some(Ok(WalkResult::Advance))
+///     }
+/// }
+/// let mut walker = WalkWithLayer::new(Stats::default(), ArrayScope);
+/// walker.walk(&root, WalkOrder::PreOrder)?;
+/// assert_eq!(walker.state().depths, vec![2, 1]);
+/// assert_eq!(walker.state().depth, 0);
+/// # Ok::<(), tvm_ffi::Error>(())
+/// ```
```
Seven-Streams: too many docs

rust/tvm-ffi/tests/test_structural_visit.rs
```diff
@@ -23,13 +23,133 @@ use tvm_ffi::{
...
-    VisitInterrupt, VisitValue, WalkOrder, WalkResult, RUNTIME_ERROR,
+    VisitInterrupt, VisitLayer, VisitValue, WalkDispatch, WalkOrder, WalkResult, WalkWithLayer,
+    RUNTIME_ERROR,
 };
 
 fn runtime_error(message: &str) -> Error {
     Error::new(RUNTIME_ERROR, message, "")
 }
 
+#[test]
```
Seven-Streams, line 33: add a dispatch test

Seven-Streams (approved):

spectrometerHBH (approved):

